PR #19, opened 2013-12-02
Support MOK blacklist

lcp:
This patch set adds the mokx support to allow the user to import the banned keys and hashes. It also includes the patches to improve the hash list to support multiple hashes in one signature list.

The related changes of mokutil is available in https://github.com/lcp/mokutil/tree/mokx-import

### DIFF
--- a/MokManager.c
+++ b/MokManager.c
@@ -513,8 +513,8 @@ static void show_efi_hash (EFI_GUID Type, void *Mok, UINTN MokSize)
 	UINTN hash_num;
 	UINT8 *hash;
 	CHAR16 **menu_strings;
-	int key_num = 0;
-	int i;
+	UINTN key_num = 0;
+	UINTN i;
 
 	sig_size = sha_size(Type) + sizeof(EFI_GUID);
 	if ((MokSize % sig_size) != 0) {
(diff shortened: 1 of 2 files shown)


vathpela:
I've pulled this into master, but I'm going to push a change to turn SHA-1 back off for our mok lists; we really shouldn't be encouraging its use at all.

TBH I'm a little unsure we should be keeping 3b8cc123ce (from some time ago), but since it's just for display, there's not a major concern.

lcp:
"openssl x509 -fingerprint" uses sha1 by default. For convenience, I switched the certificate digest from sha256 to sha1, so the user can verify the key with the digest. I would happily to use sha256 once openssl ditches sha1.

vathpela:
Yeah, that's why I didn't make 6d4803a1c0 also stop 3b8cc12 from working; it's one thing for display, but actually storing sha1 as the thing being trusted is a different level of safety. I'm having trouble seeing "openssl has a bad default" as a reason to let users do things they shouldn't.

vathpela:
FYI I've also got https://github.com/vathpela/mokutil/commit/8d5000f345e027b920929f32f36bd98cce5fcb48 in the RHEL mokutil package, so users don't do something that won't work.

PR #34, opened 2014-11-10
Fix the MokManager crash caused by the objcopy parameters

lcp:


vathpela:
Thanks, I've pushed dd5efad to master (it's 929b5b7 there.)

PR #37, opened 2015-04-08
Update Cryptlib and openssl

lcp:
Update Cryptlib to r16559 and openssl to 0.9.8zf

Signed-off-by: Gary Ching-Pang Lin <glin@suse.com>


vathpela:

PR #38, opened 2015-04-15
Explicitly request sysv-style ELF hash sections

mjg59:
We depend on there being a .hash section in the binary, and that's not
the case on distributions that default to building with gnu-style ELF
hashes. Explicitly request sysv-style hashes in order to avoid building
broken binaries.

Signed-off-by: Matthew Garrett <mjg59@coreos.com>


vathpela:
Pulled, thanks.

PR #41, opened 2015-07-15
Update openssl to 1.0.2d

lcp:
This is the update of the internal openssl and Cryptlib.
BTW, the new debug hook is really useful :-)


vathpela:

PR #43, opened 2015-10-14
Add HTTPBoot support

lcp:
This is a series of patches to add HTTPBoot support to shim.


lcp:
Updated the headers and wording.

lcp:
I've tested the patches in Plugfest 2016, and it worked on the systems that support HTTPBoot IPv4/IPv6.

vathpela:
Thanks, pushed to master.

PR #46, opened 2015-12-15
Update to openssl to 1.0.2e

lcp:
Also update Cryptlib to edk2 r19218
- Undefine NO_BUILTIN_VA_FUNCS in Cryptlib/OpenSSL/ for x86_64 to use
  the gcc builtins and remove all EFIAPI from the functions
- Move the most of defines into the headers instead of Makefile
- Remove the global variable 'timeval'
- Remove the unused code: crypto/pqueue/* and crypto/ts/*
- Include bn.h in MokManager.c due to the changes in openssl


lcp:
Close this one since 1.0.2g is coming.

Issue #47, opened 2015-12-15
please mirror MokSBState as MokSBStateRT


cyphermox:
There doesn't appear to be a way to verify whether validation in shim is enabled or disabled (the state of MokSBState / MokSB) past a reboot after initially disabling validation, for instance.

This would allow users to run mokutil only when it's really required to enable or disable validation, rather than trying to apply a change that is already in effect.

PR #48, opened 2016-01-12
shim: mirror MokSBState in runtime so the kernel can make use of it.

cyphermox:
Closes: #47


vathpela:
I've pushed this to master and it'll be in the next release.

PR #49, opened 2016-01-25
Fix the size of MokDBState

jiazhang0:
MokDBState is a 8-bit unsigned integer. Looks like a typo here.


vathpela:
Thanks, pushed to master.

PR #50, opened 2016-02-01
Measure state and second stage into TPM

mjg59:
Add support for measuring the MOK database and secure boot state into a
TPM, and do the same for the second stage loader. This avoids a hole in
TPM measurement between the firmware and the second stage loader.


vathpela:
I think I pushed this long long ago?  Seems identical to 22b58f245549a5d1f1d617a790c03966c4356d1c in master.

PR #59, opened 2016-07-18
Don't close file twice in should_use_fallback error path

bantin:
When fallback.efi is not present, the should_use_fallback error path
attempts to close a file that has already been closed, resulting in a
hang. This issue only affects certain systems.

This is a regression from version 0.8 and was introduced by commit
4794822.

Signed-off-by: Benjamin Antin <ben.antin@endlessm.com>


vathpela:
Thanks, pushed to master.

PR #60, opened 2016-07-28
MokManager: Remove the usage of APPEND_WRITE

lcp:
We got the bug report about the usage of APPEND_WRITE that may cause the
failure when writing a variable in Lenovo machines. Although
EFI_VARIABLE_APPEND_WRITE already exists in the UEFI spec for years,
unfortunately, some vendors just ignore it and never implement the
attribute. This commit removes the usage of EFI_VARIABLE_APPEND_WRITE to
make MokManager work on those machines.

https://github.com/rhinstaller/shim/issues/55

### DIFF
--- a/MokManager.c
+++ b/MokManager.c
@@ -915,6 +915,10 @@ out:
 		FreePool(old_data);
 	}
 
+	if (new_data != NULL) {
+		FreePool(new_data);
+	}
+
 	return status;
 }
 
--- a/MokManager.c
+++ b/MokManager.c
@@ -871,6 +871,16 @@ static EFI_STATUS write_db (CHAR16 *db_name, void *MokNew, UINTN MokNewSize)
 	UINTN old_size;
 	UINTN new_size;
 
+	status = uefi_call_wrapper(RT->SetVariable, 5, db_name,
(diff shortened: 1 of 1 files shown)


adempewolff:
Tested on Lenovo Thinkpad P50.  Fixes the problem--successfully was able to enroll a new key.

MokManager.c
```diff
@@ -863,6 +861,53 @@ static EFI_STATUS match_password (PASSWORD_CRYPT *pw_crypt,
...
+	CopyMem(new_data + old_size, MokNew, MokNewSize);
+
+	status = uefi_call_wrapper(RT->SetVariable, 5, db_name,
+				   &shim_lock_guid,
+				   EFI_VARIABLE_NON_VOLATILE
+				   | EFI_VARIABLE_BOOTSERVICE_ACCESS,
+				   new_size, new_data);
+
+out:
+	if (old_size > 0) {
```
lcp: Ah, I forgot to free new_data. Thanks for pointing it out. Appended a patch to fix it.

vathpela:
Thanks, pushed.

PR #63, opened 2016-08-22
Do some parsing of Extended Key Usage certificate extensions

cyphermox:


shim.c
```diff
@@ -388,6 +392,39 @@ static BOOLEAN verify_x509(UINT8 *Cert, UINTN CertSize)
...
+	x509 = d2i_X509 (NULL, &Temp, (long) CertSize);
+	if (x509 != NULL) {
+		eku = X509_get_ext_d2i(x509, NID_ext_key_usage, NULL, NULL);
+
+		if (eku) {
+			int i = 0;
+			for (i = 0; i < sk_ASN1_OBJECT_num(eku); i++) {
+				ASN1_OBJECT *key_usage = sk_ASN1_OBJECT_value(eku, i);
+
+				if (OBJ_cmp(module_signing, key_usage) == 0)
```
tyhicks: There should probably be some discussion around this. It implements a black list where only the 1.3.6.1.4.1.2312.16.1.2 key purpose OID (module signing key) is blocked. The way that I read the [Extended Key Usage section of RFC5280](https://tools.ietf.org/html/rfc5280#section-4.2.1.12) is that if a eku is specified, that the expected eku must be specified in the cert:

>  If the extension is present, then the certificate MUST only be used for one of the purposes indicated.

So if an eku is not specified, then any key will be accepted. However, if an eku is set, I think the RFC reads as if we should implement a whitelist of key purpose OIDs. Reading through the previously submitted [kernel patch set from David Howells](https://lkml.org/lkml/2015/7/17/567) I think that is probably the 1.3.6.1.4.1.2312.16.1.1 key purpose OID (firmware signing key) but I'm not completely sure about that.

vathpela: The hard part about that would be that we haven't required the 1.3.6.1.4.1.2312.16.1.1 EKU in the past, which means requiring it now would mean existing signing keys stop working - and also that we'd need to special purpose this check, since that EKU isn't in the normal keys in "db" either.  That's a pretty big reason not to move to requiring it.

tyhicks: I think you might have misunderstood my suggestion. I suggested that if an EKU is present, then it must be the correct EKU OID. If there is no EKU present, then the certificate is accepted (pending any other checks). That would allow all existing signing keys to continue working (assuming that they don't have an EKU set).

shim.c
```diff
@@ -388,6 +392,39 @@ static BOOLEAN verify_x509(UINT8 *Cert, UINTN CertSize)
...
 }
 
+static BOOLEAN verify_eku(UINT8 *Cert, UINTN CertSize)
+{
+	X509 *x509;
+	CONST UINT8 *Temp = Cert;
+	EXTENDED_KEY_USAGE *eku;
+	ASN1_OBJECT *module_signing;
+
+	module_signing = OBJ_nid2obj(OBJ_create("1.3.6.1.4.1.2312.16.1.2",
```
vathpela: One more suggestion - can you make the OID string a #define?  So that this code reads OBJ_create(OID_EKU_MODSIGN, NULL, NULL) or similar?

shim.c
vathpela: Okay, that makes sense, but I'm not quite sure I understand your suggestion.  Are you suggesting we /also/ check to see if any EKUs are present, and if so, only continue if the "firmware signing" EKU is among them?

cyphermox: For the time being, the idea is just to explicitly disallow "module signing" keys.In the future, to further enforce the types of keys to be used, we could check if EKUs are present, and if they are only continue if "firmware signing" is there (assuming that's what we want to enforce); and for compatibility with older keys that are currently in the wild, let through keys that don't specify EKU.

vathpela:
pushed, thanks.

PR #67, opened 2016-11-16
shim/tpm: fix trigger failure caused by NULL arguments 

jiazhang0:
Certain BIOS may make the strict check on the last 3 arguments passed
to get_event_log() and don't expect NULL pointers are passed. In order
to work around this failure (EFI_INVALID_PARAMETER), pass them even
though we really don't use it.


vathpela:
Make that string stay as one string, and include specific information on hardware+firmware information in the change log if you can.

With that, I'm fine with this.

PR #68, opened 2016-11-18
[V2] shim/tpm: fix trigger failure caused by NULL arguments 

jiazhang0:


vathpela:

PR #69, opened 2016-11-29
shim: fix the mirroring MokSBState fail

Ivanhu5866:
Some machines have already embedded MokSBStateRT varaible with
EFI_VARIABLE_NON_VOLATILE attribute, and some users might disable shim
vailidation manually by creating MokSBStateRT. It causes mirroring MokSBState
fail because the variable cannot be set with different attribute again, and gets
error massage every time when booting.

Fix it with checking the MokSBStateRT existence and deleting it before
mirroring it.


vathpela:

PR #73, opened 2017-02-20
httpboot: parse https prefix in the uri

lcp:
This commit adds the check for "https://" in the uri to support HTTPs Boot.


vathpela:
Thanks, merged.

PR #74, opened 2017-03-21
Update openssl 1.0.2k

lcp:
As usual, this is the commits to update openssl and Cryptlib. I also replaced several files in Cryptlib with the Null version since we never need those functions.


Conan-Kudo:
@lcp Why not update to openssl 1.1 now?

lcp:
Cryptlib is synchronized with CryptoPkg in edk2 which is using openssl 1.0.2k now. The patches for edk2 to upgrade to 1.1.0 are still under reviewing, and I will update shim as soon as edk2 merges the 1.1.0 patches.

vathpela:
Thanks, pushed to master.

PR #75, opened 2017-03-30
generate_hash(): fix the regression

jiazhang0:
The commit 03b9f800 introduces an issue in case the gap between
SumOfBytesHashed and context->SecDir->VirtualAddress exists.

This would be a typo because a formal PE image always meet
SumOfBytesHashed + hashsize == context->SecDir->VirtualAddress either
the gap exists or not.


vathpela:
Thanks so much! I was just about to start debugging this when I saw your pull.  You've saved me a bunch of time.

This will be in shim-11 (probably later today.)

PR #78, opened 2017-04-11
Update openssl to 1.1.0e

lcp:
Yet another openssl update, but this one is massive.

Due to the change of openssl API, the directory names in edk2, etc., I rewrote the update scripts to make it flexible for the future changes.

Besides the openssl update, I also fixed the gcc warnings from Cryptlib, and it should be much quieter while compiling.

Some efforts were also made to make Cryptlib smaller. I removed MD4 and DES since we don't really need them. The other potential targets are AES and HMAC, but I haven't found a sane way to disable them. If possible, I would like to shrink shim and MokManager more (<1MB), but it's still a long way to go...


vathpela:
As always, thanks for this work.

PR #79, opened 2017-04-24
Fix buffer overrun / damaged options passed to second_stage.

jsjgruber:
start is a UCS-2 character pointer and loader_len is a number of bytes.
Adjust loader_len to count characters before adding to the start pointer.


vathpela:
Thanks; it's in shim 12.

PR #81, opened 2017-05-09
fix compiler errors in security policy

mxre:
When compiling with -Werror security policy can't be compiled because of
type errors. This commit fixes this problem by adding additional casts.


vathpela:

PR #83, opened 2017-06-30
httpboot: fix OVMF crash

jiazhang0:
This is a typical typo. The free operation should be done if uri
was allocated.


lcp:
Oops. Thanks for spotting the bug.

vathpela:
Pushed to master, thanks.

PR #85, opened 2017-07-20
Add spec-based PCR7 support

mjg59:
Add support for measuring the Shim signing state into PCR7 in the same way that the firmware does in Secure Boot mode.

### DIFF
--- a/shim.c
+++ b/shim.c
@@ -962,10 +962,9 @@ static EFI_STATUS verify_mok (void) {
  * Check that the signature is valid and matches the binary
  */
 static EFI_STATUS verify_buffer (char *data, int datasize,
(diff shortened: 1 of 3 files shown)


mjg59:
Done.

martinezjavier:
@mjg59 hey, I just reviewed your patches and looks good to me.

BTW, do you think you could add some documentation on what shim measures and on which PCRs? Thanks!

vathpela:
I've added some PCR docs here: https://github.com/rhboot/shim/pull/89/commits/ab4bfc40e54238975799970696e45ae3b2da754b

martinezjavier:
@vathpela that's great, thanks a lot for adding it.

vathpela:
I've pulled this as part of #89

PR #86, opened 2017-07-24
README.fallback: correct the path of BOOT.CSV in layout example

jiazhang0:
BOOT.CSV should be placed in fedora directory in order to locate the base
directory of files recorded in $FILENAME column.


vathpela:

PR #88, opened 2017-07-31
Make fallback debug printing be dynamic at runtime.

vathpela:
This allows a user to define an EFI variable "FALLBACK_VERBOSE" with any non-zero value in order to enable debug information printing in fallback.efi .


vathpela:
This works for me locally, and Rob tells me it works for him on his test boards.  I've pulled it in to master.

PR #89, opened 2017-07-31
Make fallback aware of tpm measurements, and reboot if tpm is used.

vathpela:
Since booting the entry with fallback in the stack of things that got measured will result in all the wrong PCR values, in the cases where TPM is present and enabled, use ->Reset() instead of loading the Boot#### variable and executing its target.

This PR is layered on top of https://github.com/rhboot/shim/pull/85 .

### DIFF
--- a/README.tpm
+++ b/README.tpm
@@ -1,0 +1,22 @@
+The following PCRs are extended by shim:
+
+PCR4:
+- the Authenticode hash of the binary being loaded will be extended into
+  PCR4 before SB verification.
+
+PCR7:
+- Any certificate in one of our certificate databases that matches a binary
+  we try to load will be extended into PCR7.  That includes:
+  - DBX - the system blacklist, logged as "dbx"
+  - MokListX - the Mok blacklist, logged as "MokListX"
+  - vendor_dbx - shim's built-in vendor blacklist, logged as "dbx"
+  - DB - the system whitelist, logged as "db"
+  - MokList the Mok whitelist, logged as "MokList"
+  - vendor_cert - shim's built-in vendor whitelist, logged as "Shim"
+  - shim_cert - shim's build-time generated whitelist, logged as "Shim"
+- MokSBState will be extended into PCR7 if it is set, logged as
+  "MokSBState".
+
+PCR14:
+- MokList, MokListX, and MokSBState will be extended into PCR14 if they are
+  set.
--- a/shim.c
+++ b/shim.c
@@ -962,10 +962,9 @@ static EFI_STATUS verify_mok (void) {
  * Check that the signature is valid and matches the binary
  */
 static EFI_STATUS verify_buffer (char *data, int datasize,
-			 PE_COFF_LOADER_IMAGE_CONTEXT *context)
+				 PE_COFF_LOADER_IMAGE_CONTEXT *context,
+				 UINT8 *sha256hash, UINT8 *sha1hash)
 {
-	UINT8 sha256hash[SHA256_DIGEST_SIZE];
-	UINT8 sha1hash[SHA1_DIGEST_SIZE];
 	EFI_STATUS status = EFI_SECURITY_VIOLATION;
 	WIN_CERTIFICATE_EFI_PKCS *cert = NULL;
 	unsigned int size = datasize;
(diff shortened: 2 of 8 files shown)


shim.c
```diff
@@ -1226,7 +1226,7 @@ static EFI_STATUS handle_image (void *data, unsigned int datasize,
 		return efi_status;
 
 	/* Measure the binary into the TPM */
-	tpm_log_pe((EFI_PHYSICAL_ADDRESS)(UINTN)data, datasize, sha1hash, 9);
+	tpm_log_pe((EFI_PHYSICAL_ADDRESS)(UINTN)data, datasize, sha1hash, 4);
```
martinezjavier: Should this be just squashed with commit 18b2fa7 since it hasn't been merged yet?

shim.c
```diff
@@ -1921,37 +1921,55 @@ EFI_STATUS init_grub(EFI_HANDLE image_handle)
...
 
 	efi_status = get_variable(L"MokList", &Data, &DataSize, shim_lock_guid);
-	if (efi_status != EFI_SUCCESS)
-		return efi_status;
+	if (!EFI_ERROR(efi_status)) {
+		efi_status = tpm_log_event((EFI_PHYSICAL_ADDRESS)(UINTN)Data,
+					   DataSize, 14, (CHAR8 *)"MokList");
+		FreePool(Data);
+
+		if (EFI_ERROR(efi_status) && !EFI_ERROR(ret))
```
martinezjavier: I don't think the !EFI_ERROR(ret) check is needed here, since nothing could had changed ret at this point.

shim.c
```diff
@@ -1921,37 +1921,55 @@ EFI_STATUS init_grub(EFI_HANDLE image_handle)
...
+			ret = efi_status;
 
-	efi_status = tpm_log_event((EFI_PHYSICAL_ADDRESS)(UINTN)Data,
-				   DataSize, 14, (CHAR8 *)"MokList");
+	} else if (!EFI_ERROR(ret)) {
+		ret = efi_status;
```
martinezjavier: Same here, just else if enough AFAICT.

tpm.c
```diff
@@ -344,3 +345,31 @@ EFI_STATUS tpm_measure_variable(CHAR16 *VarName, EFI_GUID VendorGuid, UINTN VarS
...
+	} else {
+		status = LibLocateProtocol(&tpm_guid, (VOID **)&tpm);
+		if (EFI_ERROR(status))
+			return status;
+
+		if (tpm_present(tpm))
+			return EFI_SUCCESS;
+	}
+	return EFI_NOT_FOUND;
+}
```
martinezjavier: This duplicates part of the logic in tpm_log_event_raw(), I wonder if we could as a follow-up consolidate this and have the shared part in a helper function or something.

shim.c
vathpela: I'd like clarification from https://github.com/mjg59 that this is what he actually intended, since "closer" seems ambiguous, but I think probably yes.

tpm.c
vathpela: Awesomely that helper function is nearly the same as this function, just with another level of indirection and a couple more pointers passed in.  Still, removing duplicated logic is nice, so I'll push an updated version.

shim.c
martinezjavier: @vathpela agreed.

martinezjavier:
@vathpela thanks a lot for addressing the things I pointed out. Patches looks good to me now.

shim.c
mjg59: This looks correct to me

README.tpm
```diff
@@ -0,0 +1,19 @@
...
+
+PCR7:
+- DBX is extended into PCR7 when we find an entry in it that matches a
+  binary we're trying to launch
+- DB is extended into PCR7 if we find an entry in it that matches a
+  binary we're trying to launch
+- Shim's built-in cert list will be extended into PCR7 if it contains a
+  cert that matches a binary we're trying to launch.  It will be logged
+  with the variable name "Shim".
+- MokSBState will be extended into PCR7.
```
mjg59: But only if set.

README.tpm
```diff
@@ -0,0 +1,19 @@
...
+PCR4:
+- the Authenticode hash of the binary being loaded will be extended into
+  PCR4 before SB verification.
+
+PCR7:
+- DBX is extended into PCR7 when we find an entry in it that matches a
+  binary we're trying to launch
+- DB is extended into PCR7 if we find an entry in it that matches a
+  binary we're trying to launch
+- Shim's built-in cert list will be extended into PCR7 if it contains a
```
mjg59: "Shim's vendor cert will be extended into PCR7 if it matches a binary we're trying to launch", and "Shim's build cert will be extended into PCR7 if it matches a binary we're trying to launch"?

vathpela:
Thanks everyone, I've merged this now.

PR #92, opened 2017-08-11
Rework looping in enter_mok_menu(), to allow multiple MOK changes

cyphermox:
Rather than looping once through the possible actions (MokNew, MokDel, etc.), revise the logic so that instead of rebooting immediately we get back to the main menu setting a flag to replace "Continue booting" with a proper reboot.

Getting back to the menu means we can go make other changes before rebooting. For instance, you might want to enable validation, but beforehand you also need to enroll a MOK. You can already do so from userland; except the requests were cleared as soon as one of them was processed.

This involves some extra cleanup of the states to avoid running the same request more than once, removing the option from the menu once it's done, and changing prompting functions to return an EFI_STATUS so we can better track whether the process has succeeded.

### DIFF
--- a/MokManager.c
+++ b/MokManager.c
@@ -786,6 +786,14 @@ static void console_restore_mode (SIMPLE_TEXT_OUTPUT_MODE *SavedMode)
 			  SavedMode->Attribute);
 }
 
+static INTN reset_system ()
+{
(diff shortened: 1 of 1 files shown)


vathpela:
Looks good to me; I've pushed it to master.

PR #93, opened 2017-08-30
FIX 32-on-64 support

miray-tf:
Even with "ALLOW_32BIT_KERNEL_ON_X64" set the "image_is_loadable" method bails out early when a 32bit Kernel is loaded. 
This patch fixes that.

### DIFF
--- a/src/shim.c
+++ b/src/shim.c
@@ -188,7 +188,13 @@ image_is_loadable(EFI_IMAGE_OPTIONAL_HEADER_UNION *PEHdr)
 	if (PEHdr->Pe32.FileHeader.Machine != machine_type) {
 		if (!(machine_type == IMAGE_FILE_MACHINE_I386 &&
 		      PEHdr->Pe32.FileHeader.Machine == IMAGE_FILE_MACHINE_X64 &&
-		      allow_64_bit())) {
+		      allow_64_bit())
+#if defined(ALLOW_32BIT_KERNEL_ON_X64)
+		  && !(machine_type == IMAGE_FILE_MACHINE_X64 &&
+		      PEHdr->Pe32.FileHeader.Machine == IMAGE_FILE_MACHINE_I386 &&
+		      allow_32_bit())
+#endif
+		  ) {
 			return 0;
 		}
 	}


vathpela:
Huh?  This looks like it's supporting 64-bit firmware + 32-bit kernel?  Last I checked the kernel doesn't even support this - and what's the use case anyway?

I'm not sure there's any good reason to support this.

miray-tf:
Yes, I know that the Linux kernel does not support 64-bit firmware + 32-bit kernel, but that is not our use case. 

We use grub2 to boot our own operating system Symobi and we use shim to support SecureBoot. The kernel is still 32-bit only and is booted through the grub2 multiboot protocol.
We extended the grub2 multiboot loader to verify images through shim in SecureBoot mode.
We'll push our changes to github as soon as we have finished rebasing them to the patches in the master-sb branch of the rhboot/grub2 repository.

There might also be other operating systems that are still 32-bit and want to or need to support booting on a SecureBoot system.
As far as we know Reactos is still working on their 64-bit support, so they might be a candidate for this.

Why we need SecureBoot support:
We develop tools to clone, backup and erase hard disks (HDClone and HDShredder). These tools can be run in Windows and they can be booted in a Buildroot based embedded Linux or our own operating system.
The Symobi or Linux based modes are mostly used for recovery. 

How we handle Verification:
We extended the multiboot loader of grub to send the loaded multiboot image to shim for verification when SecureBoot is active.
This is only implemented for mutliboot header mode. ELF images are rejected in SecureBoot mode.
To allow verification we prepend the Symobi kernel image with a PE header followed by a multiboot header. The PE header is only used for verification through shim. For booting GRUB2 uses the mutliboot header.
Because our kernel is 32-bit only we use a 32-bit PE header. We could probably get away with using 64-bit PE header but that just feels wrong...


We have been using this setup for the last 3 years with a shim based on shim-0.9 with a patch to support a 32-bit kernel on a 64-bit userspace without any problems.

TBOpen:
> 
> 
> Huh? This looks like it's supporting 64-bit firmware + 32-bit kernel? Last I checked the kernel doesn't even support this - and what's the use case anyway?
> 
> I'm not sure there's any good reason to support this.

Um, just to be clear, you can boot the 32bit kernel on a x64-bit firmware using 64-bit shim.   Works fine, been doing it from the beginning.   It was the implementation of the patch which was taken from our original patch that was implemented incorrectly which I pointed out and included in another larger patch.

PR #94, opened 2017-08-30
Fix build with ENABLE_SHIM_CERT and ENABLE_SBSIGN

cyphermox:
shim.key is created by the shim.crt target


cyphermox:
I'm pretty sure there's a better way to do this, but this is all I could come up with. Of the requirements for %.efi.signed, it's shim.key which is not found (so build fails, can't find rule to make mmx64.efi.signed); but it is really created by the shim.crt rule, so just running make again would complete the build.

vathpela:
I don't see an obviously better way, so thanks, pulled.

PR #96, opened 2017-09-14
Small fixes for Makefile

cyphermox:
boot.csv was not being cleaned by clean; and while building the mallory tree I had some issues with parallel builds (on Ubuntu, gcc version 6.3.0 20170618 (Ubuntu 6.3.0-19ubuntu1), make 4.1):

make -j4 MAKEFILE_LEVEL=0 EFI_PATH=/usr/lib \             
        ENABLE_SHIM_CERT=1 \
        ENABLE_SBSIGN=1

This fails because shim_cert.h might not yet have been dispatched to be built by the time shim.o/netboot.o/etc. are being built (for those source files referencing shim_cert.h).

.NOTPARALLEL forcing shim_cert.h to not be built in parallel with other targets tends to force it to be built earlier. Building keys earlier should not impact the rest of the build process.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -167,6 +167,7 @@ shim.crt:
 shim.cer: shim.crt
 	$(OPENSSL) x509 -outform der -in $< -out $@
 
+.NOTPARALLEL: shim_cert.h
 shim_cert.h: shim.cer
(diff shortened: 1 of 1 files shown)


vathpela:
I've merged this and pushed it to master.

PR #98, opened 2017-09-29
Make sure we don't leave write() calls with their return values unchecked

cyphermox:


vathpela:
Done, as https://github.com/rhboot/shim/commit/cc08ed0e280f8ba3a26fbd9469e64704aaef75af

PR #99, opened 2017-10-17
Update to openssl 1.1.0e and relax the CA check

lcp:
This series of patches brings back the openssl 1.1.0e update and relaxes the CA check. For those CA certificate without the CA flag or CertSign in the x509v3 extension, a callback is added to allow those CA certificates to be loaded. However, a warning will also shows to notify to user to contact the issuer of the CA certificate to update the certificate. We should remove the workarounod callback after everyone updates the CA certificates.


vathpela:
I've merged this to master, thanks.

PR #100, opened 2017-10-23
Add some static analyzer make targets, run them, do a lot of cleanups.

vathpela:
This patch series adds some static analysis make targets, a bunch of fixes I found from running them (where noted), and some fixes I just decided to add anyway.  The coverity target puts its results here: https://scan.coverity.com/projects/rhboot-shim .  I do not believe this patchset should change any functionality.

Note that we are not currently running the checkers on the OpenSSL code.

I have not actually run the resulting build of this yet.

I also have a branch of https://github.com/rhboot/shim/pull/99 based on this at https://github.com/vathpela/mallory/tree/coverity-openssl110e .


MokManager.c
```diff
@@ -1887,15 +1895,16 @@ static EFI_STATUS mok_hash_enroll(void)
 	EFI_FILE *file = NULL;
 	UINTN filesize;
 	void *data;
+	CHAR16 *selections[] = {
+		L"Select Binary",
+		L"",
+		L"The Selected Binary will have its hash Enrolled",
+		L"This means it will Subsequently Boot with no prompting",
```
cyphermox: Subsequently probably should not be capitalized here.

MokManager.c
```diff
@@ -1887,15 +1895,16 @@ static EFI_STATUS mok_hash_enroll(void)
...
+		L"Remember to make sure it is a genuine binary before Enroling its hash",
```
cyphermox: I learned both 'enrol' and 'enroll' are valid; but we should use the same spelling in both uses in this string.

MokManager.c
```diff
@@ -1961,25 +1970,28 @@ static EFI_STATUS mok_key_enroll(void)
...
-      L"This means any binaries signed with it will be run without prompting",
-      L"Remember to make sure it is a genuine key before Enroling it",
-      NULL
-	      }, L"\\", L"", &file_name);
+	CHAR16 *selections[] = {
+		L"Select Key",
+		L"",
+		L"The selected key will be enrolled into the MOK database",
+		L"This means any binaries signed with it will be run without prompting",
+		L"Remember to make sure it is a genuine key before Enroling it",
```
cyphermox: Same as above: 'Enroling'.

lib/variables.c
```diff
@@ -51,76 +51,74 @@ variable_create_esl(void *cert, int cert_len, EFI_GUID *type, EFI_GUID *owner,
 	return EFI_SUCCESS;
 }
 
-
```
cyphermox: These changes are not GUID-related, but whitespace. Looks like they belonged to some previous commit.

lib/security_policy.c
```diff
@@ -133,8 +134,7 @@ security_policy_authentication (
 	 * EFI_SECURITY_VIOLATION */
 	fail_status = efi_status;
 
-	efi_status = uefi_call_wrapper(BS->LocateDevicePath, 3,
-				       &SIMPLE_FS_PROTOCOL, &DevPath, &h);
+	efi_status = LocateDevicePath(&SIMPLE_FS_PROTOCOL, &DevPath, &h);
```
cyphermox: I don't understand how you are reaching LocateDevicePath here without going through gBS->LocateDevicePath.

MokManager.c
vathpela: These were already that way, of course, but I've fixed them up in the next revision anyway.

lib/variables.c
vathpela: Thanks, fixed in my next version.

Makefile
```diff
@@ -11,127 +11,9 @@ endif
...
-ifneq ($(origin VENDOR_CERT_FILE), undefined)
-	CFLAGS += -DVENDOR_CERT_FILE=\"$(VENDOR_CERT_FILE)\"
-endif
-ifneq ($(origin VENDOR_DBX_FILE), undefined)
-	CFLAGS += -DVENDOR_DBX_FILE=\"$(VENDOR_DBX_FILE)\"
-endif
-
-LDFLAGS		= --hash-style=sysv -nostdlib -znocombreloc -T $(EFI_LDS) -shared -Bsymbolic -L$(EFI_PATH) -L$(LIBDIR) -LCryptlib -LCryptlib/OpenSSL $(EFI_CRT_OBJS) --build-id=sha1 $(ARCH_LDFLAGS)
+include $(TOPDIR)/Make.defaults
+include $(TOPDIR)/Make.rules
```
martinezjavier: The commit adds an empty Make.rules file...

Makefile
```diff
@@ -11,127 +11,9 @@ endif
...
+include $(TOPDIR)/Make.coverity
```
martinezjavier: ... but it doesn't do the same for Make.coverity, so this will leads to the following error:

```
$ make -j5
Makefile:16: /home/javier/devel/shim/Make.coverity: No such file or directory
make: *** No rule to make target '/home/javier/devel/shim/Make.coverity'.  Stop.
```

But I think that's better to postpone adding these empty files and instead add the files and includes in the next commit.

Make.coverity
```diff
@@ -0,0 +1,48 @@
...
+
+cov-int : clean-shim-objs
+	make $(DASHJ) Cryptlib/OpenSSL/libopenssl.a Cryptlib/libcryptlib.a
+	cov-build --dir cov-int make $(DASHJ) all
+
+cov-int-all : clean
+	cov-build --dir cov-int make $(DASHJ) all
+
+cov-clean :
+	@rm -vf $(NAME)-coverity-*.tar.*
```
martinezjavier: Shouldn't `@rm -vf $(COV_FILE)` be enough? I think that makes it easier to follow.

Make.scan-build
```diff
@@ -0,0 +1,23 @@
+include $(TOPDIR)/Make.rules
+include $(TOPDIR)/Make.defaults
```
martinezjavier: Maybe I'm missing something but I don't see why these includes are needed here, since all Make.* are included from the top level Makefile.

include/efiauthenticated.h
```diff
@@ -2,214 +2,208 @@
...
-  ///
-  /// Header before the array of signatures. The format of this header is specified 
-  /// by the SignatureType.
-  /// UINT8           SignatureHeader[SignatureHeaderSize];
-  ///
-  /// An array of signatures. Each signature is SignatureSize bytes in length. 
-  /// EFI_SIGNATURE_DATA Signatures[][SignatureSize];
-  ///
+	/*
+	 * Type of the signature. GUID signature types are defined in below.
```
martinezjavier: Is "defined in below" a correct English expression? I think the correct sentence would be just "defined below".

lib/variables.c
```diff
@@ -51,76 +51,74 @@ variable_create_esl(void *cert, int cert_len, EFI_GUID *type, EFI_GUID *owner,
...
+
+	if (Data == NULL || DataSize == NULL) {
+		return EFI_INVALID_PARAMETER;
+	}
+	//
+	// In Setup mode or Custom mode, the variable does not need to be signed but the
+	// parameters to the SetVariable() call still need to be prepared as authenticated
+	// variable. So we create EFI_VARIABLE_AUTHENTICATED_2 descriptor without certificate
+	// data in it.
+	//
```
martinezjavier: While being there, maybe you can also change this comment to not use C++ style? Although that's probably for a separate commit.

lib/variables.c
```diff
@@ -51,76 +51,74 @@ variable_create_esl(void *cert, int cert_len, EFI_GUID *type, EFI_GUID *owner,
...
+	/* we're expecting an EFI signature list, so don't free the input since
+	 * it might not be in a pool */
 #if 0
-  if (Payload != NULL) {
-    FreePool(Payload);
-  }
+	if (Payload != NULL) {
+		FreePool(Payload);
+	}
 #endif
```
martinezjavier: Also in a separate commit you may want to remove this commented out code?

MokManager.c
```diff
@@ -537,9 +538,8 @@ static void show_efi_hash (EFI_GUID Type, void *Mok, UINTN MokSize)
 
 	while (key_num < hash_num) {
 		int rc;
-		key_num = rc = console_select((CHAR16 *[]){ L"[Hash List]", NULL },
-					 menu_strings, key_num);
 
+		key_num = rc = console_select(selection, menu_strings, key_num);
```
martinezjavier: Much easier to read indeed.

MokManager.c
```diff
@@ -536,10 +536,11 @@ static void show_efi_hash (EFI_GUID Type, void *Mok, UINTN MokSize)
 	menu_strings[i+1] = NULL;
 
 	while (key_num < hash_num) {
-		key_num = console_select((CHAR16 *[]){ L"[Hash List]", NULL },
+		int rc;
+		key_num = rc = console_select((CHAR16 *[]){ L"[Hash List]", NULL },
```
martinezjavier: Weird, Github shows in the list of commits for this PR #100 as if this commit comes after 0c89c6d9a228 ("Get rid of all the places we cast to (CHAR16 *[])"), but I see in the [commits of your coverity branch]( https://github.com/vathpela/mallory/commits/coverity ) that's the opposite.

And now I see that the order is wrong for other commits too but didn't notice because were unrelated.

lib/shell.c
```diff
@@ -33,7 +34,8 @@ argsplit(EFI_HANDLE image, int *argc, CHAR16*** ARGV)
 		}
 	}
 
-	(*argc)++;		/* we counted spaces, so add one for initial */
+	/* we counted spaces, so add one for initial */
+	(*argc)++;
```
martinezjavier: This is an unrelated cleanup.

lib/simple_file.c
```diff
@@ -358,7 +360,7 @@ simple_dir_filter(EFI_HANDLE image, CHAR16 *name, CHAR16 *filter,
 			break;
 		}
 
-	next:		
+	next:
```
martinezjavier: Same. I think all unrelated cleanups should be in separate commits.

Make.coverity
vathpela: That doesn't work as well - it means if you do two runs with an intervening commit and forget to make clean, one of them isn't cleaned up.

vathpela:
This fixes everything anyone brought up, and didn't change anything else except to fix the merge error with with Hans's REQUIRE_TPM fix, so I'm going to go ahead and merge it.

PR #102, opened 2017-10-27
httpboot: Amend the device path matching rule

lcp:
Originally, we check if the last 2 nodes in the device path are
IPv4()/Uri() or IPv6()/Uri() to determine whether httpboot is used or
not. However, since UEFI 2.7, the Dns node will be inserted between the
IP node and the URI node. This commit changes the matching rule to
search IP node and URI node and ignores any node between those two
nodes. Besides, the URI node is not necessary to be the last node now.

### DIFF
--- a/httpboot.c
+++ b/httpboot.c
@@ -105,10 +105,11 @@ find_httpboot (EFI_HANDLE device)
 {
 	EFI_DEVICE_PATH *unpacked;
 	EFI_DEVICE_PATH *Node;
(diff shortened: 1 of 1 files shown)


lcp:
I found that we still need to check whether the URI node is the last or not because when UEFI loads a RAMDISK image from a http server, the device path will become .../IPv4()/.../Uri()/RamDisk(). In this case, shim just needs to load grub from the RAMDISK. I'll send a v2 request to adjust it.

PR #103, opened 2017-10-27
httpboot: Amend the device path matching rule

lcp:
Originally, we check if the last 2 nodes in the device path are
IPv4()/Uri() or IPv6()/Uri() to determine whether httpboot is used or
not. However, since UEFI 2.7, the DNS node will be inserted between the
IP node and the URI node if the server provides the DNS server address.
This commit changes the matching rule to search IP node and URI node
and ignore any node between those two nodes.


vathpela:

PR #104, opened 2017-10-27
Measure binary into the TPM when Shim verify is called

tklengyel:
Currently the only measurement the shim logs in the TPM is that of the EFI
application it directly loads. However, there are no measurements being taken
of application that are being verified through the shim_lock protocol. In this
patch we extend PCR4 for any binary for which Verify is being called through
the shim_lock protocol.

### DIFF
--- a/shim.c
+++ b/shim.c
@@ -2743,9 +2743,6 @@ install_shim_protocols(void)
 	SHIM_LOCK *shim_lock;
 	EFI_STATUS efi_status;
 
-	if (!secure_mode())
-		return EFI_SUCCESS;
-
 	/*
 	 * Did another instance of shim earlier already install the
 	 * protocol? If so, get rid of it.
@@ -2779,6 +2776,9 @@ install_shim_protocols(void)
 		return efi_status;
 	}
 
+	if (!secure_mode())
+		return EFI_SUCCESS;
+
 #if defined(OVERRIDE_SECURITY_POLICY)
 	/*
 	 * Install the security protocol hook
@@ -2794,6 +2794,12 @@ uninstall_shim_protocols(void)
 {
 	EFI_GUID shim_lock_guid = SHIM_LOCK_GUID;
 
+	/*
+	 * If we're back here then clean everything up before exiting
+	 */
+	uefi_call_wrapper(BS->UninstallProtocolInterface, 3, shim_lock_handle,
+			  &shim_lock_guid, &shim_lock_interface);
+
 	if (!secure_mode())
 		return;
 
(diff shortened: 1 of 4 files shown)


shim.c
```diff
@@ -1810,9 +1810,6 @@ EFI_STATUS shim_verify (void *buffer, UINT32 size)
 	loader_is_participating = 1;
 	in_protocol = 1;
 
-	if (!secure_mode())
-		goto done;
-
```
martinezjavier: Why are you moving this check?  Doing so changes the shim_verify() function semantics since previously it would be a no-op if Secure Boot wasn't enabled.

Also, could you please add a commit message that explains the rationale for the change? Your subject line explains what the patch does but there isn't an explanation about why the change is needed.

shim.c
```diff
@@ -1821,7 +1818,14 @@ EFI_STATUS shim_verify (void *buffer, UINT32 size)
 	if (status != EFI_SUCCESS)
 		goto done;
 
+	/* Measure the binary into the TPM */
+	tpm_log_pe((EFI_PHYSICAL_ADDRESS)(UINTN)buffer, size, sha1hash, 4);
```
martinezjavier: The binary is already measured to PCR4 in the handle_image() function. Isn't that enough? If not, please explain why in the commit message and also make sure that the README.tpm file is updated to reflect what the code does.

shim.c
tklengyel: Yes I can move it to a separate commit. This move is necessary to ensure that measurements are being taken even if SecureBoot is disabled. I think this is actually more coherent with how load_image behaves as that also takes a measurement even if SecureBoot is disabled.

shim.c
tklengyel: I've updated the commit message:
```
Currently the only measurement the shim logs in the TPM is that of the EFI
application it directly loads. However, there are no measurements being taken
of application that are being verified through the shim_lock protocol. In this
patch we extend PCR4 for any binary for which Verify is being called through
the shim_lock protocol.

tklengyel:
I've added two more additional tweaks to this PR:

1. 2de71a5  adds `REQUIRE_TPM` compile-flag will treat all TPM related issues as critical (currently these are just discarded)
2. 5c6e3ac Add a fall-back option in case the TPM2 measurement with the `PE_COFF_IMAGE` flag fails. Currently on my Dell ultrabook I get a `EFI_UNSUPPORTED` error when the shim tries to measure either Linux or Xen with this flag set (cc @mjg59)

mjg59:
LGTM.

tklengyel:
I just noticed the shim_lock protocol wasn't getting installed when SecureBoot is disabled so I just pushed an additional commit to change that.

vathpela:
Thanks, I've pushed this to master, and assuming we don't find any issues in testing, it'll be in shim 15.

PR #107, opened 2017-12-13
Uninstall shim protocols before re-installing them

cyphermox:
Make sure if we chainload things, a chainloaded bootloader will be able to use
the latest systab replacements and protocols. They need to match for things
to validate correctly.

### DIFF
--- a/replacements.c
+++ b/replacements.c
@@ -100,19 +100,6 @@ start_image(EFI_HANDLE image_handle, UINTN *exit_data_size, CHAR16 **exit_data)
 	EFI_STATUS status;
 	unhook_system_services();
 
-	/* We have to uninstall shim's protocol here, because if we're
-	 * On the fallback.efi path, then our call pathway is:
-	 *
-	 * shim->fallback->shim->grub
-	 * ^               ^      ^
-	 * |               |      \- gets protocol #0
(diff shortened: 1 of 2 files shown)


replacements.c
```diff
@@ -100,22 +100,8 @@ start_image(EFI_HANDLE image_handle, UINTN *exit_data_size, CHAR16 **exit_data)
...
-	 * |               \- installs its protocol (#1)
-	 * \- installs its protocol (#0)
-	 * and if we haven't removed this, then grub will get the *first*
-	 * shim's protocol, but it'll get the second shim's systab
-	 * replacements.  So even though it will participate and verify
-	 * the kernel, the systab never finds out.
-	 */
 	if (image_handle == last_loaded_image) {
 		loader_is_participating = 1;
-		uninstall_shim_protocols();
```
vorlonofportland: I think you want to keep this call here for now, for compatibility when a new shim is chainloading to an old shim.  When we're chainloading to a new shim, the other change in shim.c is sufficient and looks correct, but I don't think we should regress the case where we've updated the chainloading shim but not the chainloaded shim.

vathpela:
Thanks, I've pushed to master.

PR #108, opened 2017-12-21
Fix Http Boot Enabled build error

cdadmin:
Shim 14 no longer builds with http boot enabled.  It appears it is missing an include to console.h since in_protocol was removed.

In addition, this provides Proxy DHCP support for netboot when the boot options are provided via a proxy dhcp instead of the dhcp server.  The second stage bootloader must also implement proxy offers in order for this to work.

### DIFF
--- a/httpboot.c
+++ b/httpboot.c
@@ -37,6 +37,7 @@
 #include "Http.h"
 #include "Ip4Config2.h"
 #include "Ip6Config.h"
+#include "console.h"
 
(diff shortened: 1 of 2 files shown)


vathpela:
Thanks!  Part of this overlapped with #109 , so I took that patch, and I pulled his first.  I've also pushed your proxy dhcp support patch to master, and it'll be in shim 15.

PR #109, opened 2018-01-04
httpboot: include console.h

lcp:
in_protocol is declared in console.h, so httpboot.c has to include the
header.


vathpela:

PR #112, opened 2018-03-01
Let generate_path handle paths that start with volume labels

einherjer0:
This change fixes the following problem:
```
Launching UEFI Removable image...···
dppath: \EFI\BOOT\bootx64.efi
path:   FS4:\EFI\BOOT\bootx64.efi
Failed to open \EFI\BOOT\FS4:\EFI\BOOT\bootx64.efi - Invalid Parameter
Failed to load image \EFI\BOOT\FS4:\EFI\BOOT\bootx64.efi: Invalid Parameter
start_image() returned Invalid Parameter
Error reported: Invalid Parameter
```

### DIFF
--- a/shim.c
+++ b/shim.c
@@ -1595,6 +1595,16 @@ static EFI_STATUS generate_path(EFI_LOADED_IMAGE *li, CHAR16 *ImagePath,
 	EFI_STATUS efi_status = EFI_SUCCESS;
 	CHAR16 *bootpath;
 
+	// Remove volume label if present, ie "FSx:"
+	if (StrSize(ImagePath) > 4 && 
+		StrnCmp(ImagePath, L"FS", 2) == 0 &&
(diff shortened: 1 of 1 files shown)


vathpela:
How do you get in to a situation where you have the text string "FS4:" in ImagePath to begin with?  It seems like however that happened needs to be addressed instead, especially given that the workaround here gives the illusion that we've parsed the filesystem number like the EFI Shell would, even though we've merely ignored it.

martinezjavier:
@einherjer0 I've proposed PR #147 to fix some issues when invoking shim from the UEFI shell using relative paths. But I believe that it should also fix your issue.

einherjer0:
@martinezjavier Thanks! I will try to verify your fix on my system and report back.

einherjer0:
This fix worked for me. Thanks @martinezjavier

PR #114, opened 2018-03-07
Expose ALLOW_32BIT_KERNEL_ON_X64 via Make.defaults

tklengyel:
The flag ALLOW_32BIT_KERNEL_ON_X64 has already been checked for in shim.c,
in this patch we just expose it through the Makefile.


tklengyel:
This PR is now based on https://github.com/rhboot/shim/pull/93

PR #115, opened 2018-03-07
Add Measure function to the shim lock protocol

tklengyel:
Expand the shim lock protocol to include a "Measure" functionality which can
be used to record hashes into the TPM of arbitrary buffers. The Measure
function will always measure the buffers directly (ie. not with Authenticode),
but for buffers that point to PE images the log will record that it was of a
PE image.

### DIFF
--- a/shim.c
+++ b/shim.c
@@ -1784,6 +1784,25 @@ done:
 	return efi_status;
 }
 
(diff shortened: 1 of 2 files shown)


brendank310:
Any objections to this approach? I'd be interested in using this functionality when merged into the mainline.

PR #116, opened 2018-03-07
shim: Don't overwrite EFI_LOADED_IMAGE's LoadOptions when not needed

tklengyel:
When the firmware is using EFI_LOAD_OPTION to specify options for the secondary
loader, the shim will properly detect that and return in set_second_stage. Later
howerer in handle_image EFI_LOADED_IMAGE is being overwritten with load_option
irrespective of the fact that load_option was never set. This effectively
prevents the EFI_LOAD_OPTION from reaching the secondary loader.

Only overwrite EFI_LOADED_IMAGE's LoadOptions when load_option is not NULL
solves the problem.


vathpela:

PR #117, opened 2018-03-12
Support for transparent LoadImage()/StartImage()

mcb30:
These patches allow applications invoked by shim to use LoadImage() and StartImage() to transparently verify (and invoke) further applications using shim's own root of trust.  Applications no longer need to be modified in order to perform verification against shim's trusted certificate list.


vathpela:
I really like most of this! It's quite close to what I've had on my TODO list for some time as part of version 2 of the shim protocol.  A couple of things stick out though.
The biggest issue is that it actually *can't* be transparent.  There's hidden ABI between LoadImage() and StartImage(), because of the fact that the EFI_LOADED_IMAGE * returned by LoadImage() isn't actually the base of the allocation it makes.  As a result, any implementation of StartImage() needs to be compatible with the firmware's implementation of LoadImage() (and vise versa).
So the thing to do is actually define EFI_LOADED_IMAGE_2 that has the normal exported fields from EFI_LOADED_IMAGE in it, plus a "VOID *PrivateData;" field, and then put everything else from LoadImage()'s implementation in a EFI_LOADED_IMAGE_2_PRIVATE_DATA structure, instead of the way it currently exists.  Then the two aren't compatible, and you'll have to load this one intentionally, but we won't have the memory corruption issue we have from the original EFI_LOADED_IMAGE implementation.  And if we get it right, it should be subclassable.

The other thing I've wanted to implement in shim v2 is the ability for a signed program we've loaded to add its keys to the key list - so we don't have to carry all keys, but there's a strict path dependency on the execution flow for any keys registered from something we load.  There's a spec I wrote for this for UEFI years ago, which we could just implement directly as part of the Loader Protocol here: https://docs.google.com/document/d/1IXIMLhVsTswEhyU6ChbFNZvQe04ctJj6OZnmISu8_AE/edit?usp=sharing

mcb30:
On 14/03/18 17:42, Peter Jones wrote:
> I really like most of this!

Thank you!

> The biggest issue is that it actually /can't/ be transparent. There's 
> hidden ABI between LoadImage() and StartImage(), because of the fact 
> that the EFI_LOADED_IMAGE * returned by LoadImage() isn't actually the 
> base of the allocation it makes. As a result, any implementation of 
> StartImage() needs to be compatible with the firmware's implementation 
> of LoadImage() (and vise versa).

Transparency is already working.

My patch creates a separate (shim-private) LOADER_PROTOCOL which is also 
attached to the image handle for any images that shim has to load using 
its own signature verification and PE loader code.

shim already hooks the system table to override LoadImage(), 
StartImage(), and Exit().  By also hooking UnloadImage(), we get full 
control over all the code paths that might attempt to access the 
firmware's EFI_LOADED_IMAGE private data structure.

My patch causes shim to check for the existence of LOADER_PROTOCOL on 
the image handle.  If LOADER_PROTOCOL is found then we call the 
LOADER_PROTOCOL methods for StartImage(), UnloadImage(), and Exit() 
instead of calling the underlying firmware.

This allows shim to provide global methods for all four of LoadImage(), 
StartImage(), UnloadImage(), and Exit(), all of which will correctly 
handle either shim-loaded images (which will have a LOADER_PROTOCOL 
instance) or firmware-loaded images (which will not have a 
LOADER_PROTOCOL instance).

Do you see any problems with this approach, which already works to 
provide transparency?

Thanks,

Michael

vathpela:
... Though obviously we'd want to implement that as API calls instead of EFI variables, so that it works for second->third generation patches.

mcb30:
On 14/03/18 21:20, Peter Jones wrote:
> ... Though obviously we'd want to implement that as API calls instead of 
> EFI variables, so that it works for second->third generation patches.

It's already implemented in terms of API calls: shim intercepts the boot 
services API calls in the system table (which it was already doing), and 
then shim uses the API calls within the image's individual 
LOADER_PROTOCOL instance to actually perform the StartImage(), 
UnloadImage(), or Exit() for that image.

The API as exposed to other UEFI executables is unmodified: they 
continue to just invoke LoadImage() and StartImage() and everything just 
works transparently.

The LOADER_PROTOCOL API is such that it will already work even if the 
shim that discovers and invokes the LOADER_PROTOCOL method is not the 
same shim that actually installed the LOADER_PROTOCOL.

The image loading and execution is done using the existing shim 
handle_image() call, which already measures the verified code to the TPM.

This pull request isn't an abstract design suggestion; this is code that 
is already implemented and working.  What problems do you see that 
haven't already been addressed in the implementation?

Thanks,

Michael

mcb30:
Sorry; I just realised that your " ... Though obviously we'd want to implement that as API calls instead of EFI variables" was referring to your Google doc link rather than to the patch!  Apologies for the noise.

mcb30:
Any further comments?  You can trivially check that this code **does** work to allow a second-stage bootloader (such as iPXE) to transparently use LoadImage()/StartImage() to boot a UEFI shell (or Linux kernel) signed only with shim's VENDOR_CERT_FILE.  Exiting from the shell back to iPXE also works.

vathpela:
The problem isn't the /API/, the problem is the /ABI/ - there's no way to guarantee that an EFI_LOADED_IMAGE from before shim was executed won't get passed to our StartImage(), and this does actually happen sometimes.  When it does, if the structure of the allocation the EFI_LOADED_IMAGE is in is different between the two, then StartImage() can take a fault (or worse - silently corrupt data) accessing the wrong data structure.  There's no way to be sure this is safe at all, except to not do it.

mcb30:
We already handle that situation correctly.  The EFI_LOADED_IMAGE from before shim was executed won't have a LOADER_PROTOCOL installed on the handle, so my code will already hand that off to the underlying platform's StartImage() call.

vathpela:
But your LOADER_PROTOCOL method only handles /half/ of the problem.  It doesn't address what happens if the built-in version of StartImage() gets passed our EFI_LOADED_IMAGE handles.

mcb30:
We override the StartImage() function pointer within the EFI system table, so the built-in version of StartImage() will never be passed one of our EFI_LOADED_IMAGE handles.

vathpela:
That makes the assumption that nothing has already taken a pointer to it.  This is not a safe assumption.

vathpela:
BTW, I merged 8721bbe6fb1 earlier than I should have, and wound up needing to revert it.  The issue is that it needs to check the blacklists on each certificate within the list.

mcb30:
The reversion of 8721bbe is incorrect.  That commit makes no changes to the blacklist checking behaviour for "shim_cert" and "vendor_cert".

Specifically: "cert" is checked against the blacklists, but neither "shim_cert" nor "vendor_cert" is checked against the blacklists.  This behaviour is identical both before and after commit 8721bbe.  Reverting commit 8721bbe simply breaks useful functionality; it doesn't change anything relating to blacklist checking.

I'm happy to _add_ the new functionality to verify shim_cert and vendor_cert against the blacklists, if that would be considered useful.  Please revert the reversion, and I'll put together a patch.

mjg59:
We're not in control of the behaviour of drivers, and if something an app does triggers a driver into calling LoadImage() it's going to end up with our implementation rather than the system implementation. But since that driver may have been initialised before Shim runs, it's certainly possible that it's grabbed a pointer to StartImage. It doesn't seem hugely likely that it would stash a pointer to one and not the other, but (a) there are a lot of drivers and (b) firmware is, on average, bad, so I wouldn't bet on there being 0 cases of this.

(I /guess/ you could overwrite the host StartImage to jump to your StartImage, but ideally it's not going to be mapped writable so you're going to have to re-map it first and I suspect you are going to have a very bad time)

vathpela:
(Can you make the discussion about 8721bbe a separate PR?)

mcb30:
8721bbe discussion moved to #119

mcb30:
OK.  If you want to cope with that kind of pathological potential behaviour, then we need to use the underlying system LoadImage() to load a stub binary (embedded into shim), and have the stub binary act as a proxy for the real loaded image (e.g. by locating a LOADER_PROTOCOL instance on its own image handle).  Do you see any way for **that** design to fail?

vathpela:
We've also run in to situations where calling the system's LoadImage() on a binary not verifiable by built-in keys will result in UI-blocking dialog boxes and the like.  I'm very much not convinced we actually *must* have this be transparent, nor that there's a significant upside, especially since it requires this kind of effort.

I would significantly rather:
- define EFI_LOADED_IMAGE_PROTOCOL_2 and make it have a Private element, so it can be properly subclassed, so that I can write an ECR for UEFI to add it and LOADER_PROTOCOL (as EFI_LOADER_PROTOCOL) to the spec and have a proper way to subclass it.
- add a call (when I do that ECR work) to add a registration function to make it possible to intercept gBS->LoadImage(), gBS->StartImage(), etc. from behind the system table pointer.  (This also gives us an opportunity to make systab transition to +r-wx page permissions, which we can't currently do because of shim replacements.)

Tangentially, in any case we should also have UnloadImage() in LOADER_PROTOCOL.

mcb30:
Unfortunately solving it the "right" way will involve a 5-10 year lag before the solution becomes meaningfully usable for anyone, since it will take at least that long for the change to propagate out to most deployed systems.  That's not viable for most users.

Making it transparent is important because it allows decisions about security policy changes to be isolated within shim.  Other code such as the UEFI shell, iPXE, etc, should not decide upon security policy: they should simply use the platform standard API for loading and starting images, and thereby delegate security policy decisions to the platform.

Also, consider the situation: platform -> shim -> image A -> image B:

We could plausibly add a non-transparent API for loading and starting images in order to allow image A to load and start arbitrary unrelated image B.  If image A is aware of this non-transparent API then it can use it: so far, so good.

The problem arises when image B (which has no knowledge of this non-transparent API) tries to find its own EFI_LOADED_IMAGE_PROTOCOL instance, in order to locate its load options or install an Unload() method.  This has to work, since it's guaranteed by the UEFI spec.

Therefore, even a non-transparent API would still end up having to install an EFI_LOADED_IMAGE_PROTOCOL instance.  There is no way around this.  The current shim code hacks support for this by modifying its own EFI_LOADED_IMAGE_PROTOCOL instance (see the "grub needs to know its location and size in memory" portion of handle_image()): this works for grub but not for arbitrary second-stage loaders since the second-stage loader may want to call Exit().

So, we know that whatever API we use (transparent or non-transparent) **must** install an EFI_LOADED_IMAGE_PROTOCOL.

To accommodate mjg59's comment about a theoretically possible way for the system StartImage() to end up being passed a handle created by shim, there needs to be an EFI_LOADED_IMAGE_PROTOCOL created by the system LoadImage().

Combining these two: whatever API we use (transparent or non-transparent) **must** use the system LoadImage() to create an EFI_LOADED_IMAGE_PROTOCOL.

The easiest way to do this is to have a trivial proxy EFI application signed with the same signature as shim and embedded as a payload within shim.  The proxy application simply passes control to the LOADER_PROTOCOL installed on the image handle.

We can design LOADER_PROTOCOL to allow for it eventually replacing EFI_LOADED_IMAGE_PROTOCOL; I'm happy to do this.  We can also provide the proxy application within shim as a practical solution for users who don't want to wait 5-10 years.

Thoughts?

mcb30:
...and LOADER_PROTOCOL already has an UnloadImage().  :)

mcb30:
> The easiest way to do this is to have a trivial proxy EFI application signed with the same signature as shim and embedded as a payload within shim. The proxy application simply passes control to the LOADER_PROTOCOL installed on the image handle.

@mjg59 for future reference: there is now a proxy application that does precisely this, at https://github.com/ipxe/ProxyLoaderPkg.  It has been signed for Secure Boot.

This has all been working since November 2018, but I have given up on attempting to get anything useful upstreamed into shim since even accepted bugfix patches seem to get ignored.

PR #119, opened 2018-03-23
Revert "Revert "Allow shim to handle multiple trusted certificates""

mcb30:
The reasoning given in "Revert "Allow shim to handle multiple trusted
certificates"" is incorrect.  The original commit "Allow shim to
handle multiple trusted certificates" makes no changes to the
behaviour of shim in terms of blacklisted certificate checking.

If it is desired to add support for checking shim_cert and/or
vendor_cert against the certificate blacklists, then this
functionality should be added separately.  It is entirely unrelated to
the ability of shim to handle multiple trusted certificates.

This reverts commit cdbfb5a69ea16142d132517d39c9e96d154c8345.

Signed-off-by: Michael Brown <mbrown@fensystems.co.uk>

### DIFF
--- a/shim.c
+++ b/shim.c
@@ -991,6 +991,35 @@ done:
 }
 
 /*
+ * Check that a trusted certificate signed the binary
+ */
+static BOOLEAN verify_trusted_cert(const WIN_CERTIFICATE_EFI_PKCS *cert,
+				   const UINT8 *sha256hash,
+				   const UINT8 *trusted_cert,
+				   size_t trusted_cert_len)
+{
+	const UINT8 *tmp;
+	ASN1_TYPE *asn1;
+
+	while (trusted_cert_len) {
+		if (AuthenticodeVerify(cert->CertData,
+				       cert->Hdr.dwLength - sizeof(cert->Hdr),
+				       trusted_cert, trusted_cert_len,
+				       sha256hash, SHA256_DIGEST_SIZE))
+			return TRUE;
+		tmp = trusted_cert;
+		asn1 = d2i_ASN1_TYPE(NULL, &tmp, trusted_cert_len);
+		if (!asn1)
+			break;
+		ASN1_TYPE_free(asn1);
+		trusted_cert_len -= (tmp - trusted_cert);
+		trusted_cert = tmp;
+	}
+
+	return FALSE;
+}
+
+/*
  * Check that the signature is valid and matches the binary
  */
 static EFI_STATUS verify_buffer (char *data, int datasize,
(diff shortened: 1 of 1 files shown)


mcb30:
BTW, the Travis CI build (and my local build) are both failing due to f391e44.

vathpela:
Travis should be fixed if you rebase to after 23322362f58.

vathpela:
Aside from the Travis issue, which is my problem, I'm conceptually okay with this for shim 16 (sorry - I'm in the middle of trying to get 15 out the door, and it's deadline season.)

You are, of course, correct about it not currently checking the blacklist on those - I was mistaken that we were already doing that.  That said, I think we do need to do such checking for vendor_cert, but not for shim_cert, since it is nominally ephemeral, existing only during the build process.  The fact that we're not currently doing so is a bug, though a fairly minor one.

mcb30:
Thanks.  Rebased, and Travis is now building ok.  Adding support for validating vendor_cert (and optionally shim_cert) will be easier if the patch is reverted, and I'd also be much happier since there wouldn't then be a public commit message in the shim log erroneously claiming that I'd introduced a security hole.

As I said, I'm happy to write the subsequent patch to add blacklist checks of vendor_cert (and, since it's no extra code, shim_cert as well); would you like me to do this?

vathpela:
Please, yes.  I am sorry about the public log, and I do apologize.  It was my error.

mcb30:
Thanks.  I'll get the blacklist check code added to this PR by Monday.

mcb30:
I'm trying to figure out precisely what to check for a blacklisted vendor_cert.  The existing code in check_blacklist() calls check_db_cert_in_ram(). This tests each X.509 certificate in dbx (or other blacklist) to see if it can be used to verify the binary's sha256 hash via the binary's PKCS7 signature.  The binary's sha256 hash should be irrelevant to the outcome of the question "is certificate X blacklisted", so this is slightly puzzling.

For vendor_cert, should we be testing each X.509 certificate in dbx (or other blacklist) to see if it is an exact copy of vendor_cert?  If not, what test should we use?

Thanks, and sorry for the delay.

mcb30:
In the absence of any response: please merge this PR as-is, to revert the existing incorrect reversion.

I'm still happy to add the certificate blacklist checking feature if desired, once the above question has been answered.

mcb30:
Rebased onto current master and ready (again) to be merged.

mcb30:
Rebased again to current master.  Still ready to merge.  No substantive change since all comments were resolved in March.

PR #135, opened 2018-05-24
Add a countdown menu to fallback

lcp:
When TPM presents, to make the measurement result consistent, fallback should reset the system. However, some machine with faulty firmware may keep loading \EFI\Boot\bootx64.efi -> fbx64.efi, and the system became unusable. To give the user a chance to stop system reset, a 5-seconds countdown menu is added so that the user can interrupt system reset and choose to boot the restored boot option. Besides, the "Always continue boot" option will create a variable, FB_NO_REBOOT, and fallback will just load the restored boot option if the variable exists, so that we can work around the rebooting issue in the faulty machine.

### DIFF
--- a/MokManager.c
+++ b/MokManager.c
@@ -729,30 +729,6 @@ done:
 	return efi_status;
 }
 
-static void console_save_and_set_mode(SIMPLE_TEXT_OUTPUT_MODE * SavedMode)
-{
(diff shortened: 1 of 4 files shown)


vathpela:
Thanks; I've fixed the conflict here and pushed it to master.

PR #140, opened 2018-07-21
Implement vendor EFI Signature List (ESL)

puiterwijk:


puiterwijk:
Note that I've spoken with @vathpela, and we agreed on that I should rework this.
I will hopefully do so sometime soon.

vathpela:
I think this has been obviated by #203, and I don't remember what we agreed on that needed reworking, so I'm closing it now.  If you happen to remember anything else we wanted to do, feel free to re-open with a note.

PR #142, opened 2018-08-07
Fix for "Section 0 has negative size" error when loading fbaa64.efi

maranwilson:
The current code is incorrectly failing to load the fbaa64.efi image found
in Arm servers even though the UEFI shell code is able to properly load
and execute the same image.

Aside from my own test cases involving KVM guests running on Arm, I'm
pretty sure this addresses the root cause for problems others have reported
as well such as:
https://bugzilla.redhat.com/show_bug.cgi?id=1527283
https://bugzilla.redhat.com/show_bug.cgi?id=1489604

I've tested this patch on an Arm64 server and confirmed the problem
is resolved. I've also tested the patches on an x86 machine to confirm
there were no regressions.

### DIFF
--- a/shim.c
+++ b/shim.c
@@ -1377,6 +1377,11 @@ static EFI_STATUS handle_image (void *data, unsigned int datasize,
 	 */
 	Section = context.FirstSection;
 	for (i = 0; i < context.NumberOfSections; i++, Section++) {
+		/* Don't try to copy discardable sections with zero size */
+		if ((Section->Characteristics & EFI_IMAGE_SCN_MEM_DISCARDABLE) &&
+		    !Section->Misc.VirtualSize)
+			continue;
+
 		base = ImageAddress (buffer, context.ImageSize,
 				     Section->VirtualAddress);
 		end = ImageAddress (buffer, context.ImageSize,


vathpela:
This seems really weird for me - it seems like you've got an invalid section header, but it's very hard to know without seeing it.  So with that in mind, can you include the objdump output of the headers in the commit message?  You'll need to change the byte at 0x4c from 0xAA to 0x86 and then do something like: `objdump -x -m aarch64 ./boot/efi/EFI/BOOT/fbaa64.efi` (because bfd doesn't know the right info).

maranwilson:
Thanks for taking a look. The commit message has been updated as requested. The fbaa64.efi file in question is simply being compiled on a 64 bit Arm machine from latest sources in this repo.  Both by myself and by others who also compile and release aarch64 RPMs based on this code. I've been assuming it's not a problem that needs to be fixed on the compilation side of the fbaa64.efi binary due to the fact that the UEFI shell code has no trouble parsing and executing the same file.

maranwilson:
Friendly ping... Just a reminder that I've updated the commit message as requested. Please let me know if there is any other questions I can answer or additional information I can provide to help move this bug fix along. I'm also open to coding up and testing out alternative ideas if there is a different solution that would be preferred by the community.

vathpela:
I think everything here got pushed quite some time back, and this just didn't get closed.  Closing it now.

Issue #144, opened 2018-08-21
Mkdir wrong when compile


tanminger:
I am using Ubuntu 14.04 x86_64. The version of make is 3.81.
I get the following error when compile:
Fatal error: can't create Hash/CryptMd4Null.o: No such file or directory

Analyse the reason:
ls -l Cryptlib/
total 4
drwxrwxr-x 2 tanminger tanminger 4096 Aug 21 02:34 {Hash,Hmac,Cipher,Rand,Pk,Pem,SysCall}
$

In Makefile line 105:
    mkdir -p Cryptlib/{Hash,Hmac,Cipher,Rand,Pk,Pem,SysCall}

This command should has issue.

tanminger:
I create a patch for this issue: https://github.com/rhboot/shim/pull/145

PR #145, opened 2018-08-21
Fix the compile error of mkdir wrong directory.

tanminger:
In Ubuntu 14.04, the following code in old Makefile:
  mkdir -p Cryptlib/{Hash,Hmac,Cipher,Rand,Pk,Pem,SysCall}
will create a directory named "{Hash,Hmac,Cipher,Rand,Pk,Pem,SysCall}".

Signed-off-by: Ming Tan <ming.tan@intel.com>


tanminger:
This patch is for fix issue https://github.com/rhboot/shim/issues/144

PR #153, opened 2018-11-05
Make sure that MOK variables always get mirrored

puiterwijk:
Without this, if a Mok variable doesn't exist in Boot Services, it will also
not be copied to Runtime, even if we have data to be added to it (vendor cert).
This patch makes sure that if we have extra data to append, we still mirror
the variable.

### DIFF
--- a/mok.c
+++ b/mok.c
@@ -223,11 +223,26 @@ EFI_STATUS import_mok_state(EFI_HANDLE image_handle)
 		UINT32 attrs = 0;
 		BOOLEAN delete = FALSE, present, addend;
 
+		addend = (v->addend_source && v->addend_size &&
+			  *v->addend_source && *v->addend_size)
+			? TRUE : FALSE;
+
 		efi_status = get_variable_attr(v->name,
 					       &v->data, &v->data_size,
 					       *v->guid, &attrs);
-		if (efi_status == EFI_NOT_FOUND)
+		if (efi_status == EFI_NOT_FOUND) {
+			if (v->rtname && addend) {
+				efi_status = mirror_one_mok_variable(v);
+				if (EFI_ERROR(efi_status) &&
+				    ret != EFI_SECURITY_VIOLATION)
+					ret = efi_status;
+			}
+			/*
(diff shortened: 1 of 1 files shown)


puiterwijk:
To see this in action, on a system without any custom entries in the non-volatile MokList, run "mokutil --list-enrolled".

With this bug, it will say: `No such file or directory: MokListRT`.
With this fix, it will list the vendor certificate.

lcp:
The patch looks good.

mok.c
```diff
@@ -223,11 +223,26 @@ EFI_STATUS import_mok_state(EFI_HANDLE image_handle)
...
+		addend = (v->addend_source && v->addend_size &&
+			  *v->addend_source && *v->addend_size)
+			? TRUE : FALSE;
+
 		efi_status = get_variable_attr(v->name,
 					       &v->data, &v->data_size,
 					       *v->guid, &attrs);
-		if (efi_status == EFI_NOT_FOUND)
+		if (efi_status == EFI_NOT_FOUND) {
+			if (v->rtname && addend) {
```
LennySzubowicz: Is is ok to count on the value of get_variable_attr() output parameters on an error return?
Specifically, for an EFI_NOT_FOUND error return, v->data_size won't be touched as the code is currently. And it's 0 on input because it's in a static struct. However, the value in v->data_size is used by mirror_one_mok_variable(). So I'm thinking it might be safer to guard against future changes and to make this assumption explicit by clearing it on this special case path.

`v->data_size = 0;`

lcp: https://github.com/rhboot/shim/blob/master/lib/variables.c#L207-L231

By design, get_variable_attr() sets len to 0 to query the size of the variable.

LennySzubowicz: Yes, it's completely obvious once you pointed that out. Don't know how I missed it.
The patch looks very good.

vathpela:
Thanks - I've pulled both @puiterwijk's patch and the one from @lcp's patchset, resolving the conflict, and then moved the logic from the goto into a helper function in a third commit.

PR #157, opened 2018-12-06
Make EFI variable copying fatal only on secureboot enabled systems

puiterwijk:
I have come across systems that are unwilling to reserve enough memory for
a MokListRT big enough for big certificates.
This seems to be the case with firmware implementations that do not support
secureboot, which is probably the reason they went with much lower variable
storage.

This patch set makes sure we can still boot on those systems, by only
making the copy action fatal if the system has secure boot enabled, or if
the error was anything other than EFI_INVALID_PARAMETER.

### DIFF
--- a/shim.c
+++ b/shim.c
@@ -2628,7 +2628,18 @@ efi_main (EFI_HANDLE passed_image_handle, EFI_SYSTEM_TABLE *passed_systab)
 	 * boot-services-only state variables are what we think they are.
 	 */
 	efi_status = import_mok_state(image_handle);
-	if (EFI_ERROR(efi_status)) {
+	if (!secure_mode() && efi_status == EFI_INVALID_PARAMETER) {
+		/*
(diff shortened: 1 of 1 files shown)


arrfab:
FWIW, it affects some CentOS Users and we consolidated all feedback in the following bug report : 
https://bugs.centos.org/view.php?id=15522
Some people confirm that the .patch in this commit permit people to boot fine (not considered critical error anymore)

lcp:
If it's the generation of MokListRT causing the failure, it'd be better to change the condition check right after import_mok_state(). The code after the "die" label means the program must be terminated after showing the error messages.

puiterwijk:
@lcp thanks for that feedback, I ended up adding an additional `if` clause which only triggers in this specific case.

lcp:
Looks good to me :)

vathpela:
Maybe we ought to consider being able to mirror only the sha2 keyids of large certs?

PR #158, opened 2018-12-11
Remove call to TPM2 get_event_log()

mjg59:
Calling the TPM2 get_event_log causes the firmware to start logging
events to the final events table, but implementations may also continue
logging to the boot services event log. Any OS that wishes to
reconstruct the full PCR state must already look at both the final
events log and the boot services event log, so if this call is made
anywhere other than immediately before ExitBootServices() then the OS
must deduplicate events that occur in both, complicating things
immensely.

Linux already has support for copying up the boot services event log
across the ExitBootServices() boundary, so there's no reason to make
this call. Remove it.

Signed-off-by: Matthew Garrett <mjg59@google.com>

### DIFF
--- a/tpm.c
+++ b/tpm.c
@@ -70,41 +70,6 @@ static BOOLEAN tpm2_present(EFI_TCG2_BOOT_SERVICE_CAPABILITY *caps,
 	return FALSE;
 }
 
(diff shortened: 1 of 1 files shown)


vathpela:

PR #159, opened 2018-12-18
shim: only include shim_cert.h in shim.c

lcp:
Currently shim_cert.h was included in shim.h. This made every user of
shim.h create its own shim_cert array and growed the overall file size.
Since shim_cert is only used in shim.c, it should be only included in
shim.c.

### DIFF
--- a/shim.c
+++ b/shim.c
@@ -35,6 +35,10 @@
 
 #include "shim.h"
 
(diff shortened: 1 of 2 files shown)


lcp:
There are more to be fixed. Will send another PR.

Issue #154, opened 2018-11-20
Reset MOK function does not cause MokListRT to be repopulated with the embedded VENDOR_CERT_FILE


cyphermox:
I ran mokutil --reset with a very recent mokutil, using SHIM 15+ snapshot (3beb971b).

After rebooting and using the Reset MOK option in MokManager, the embedded certificate is no longer listed in mokutil --list-enrolled; in /proc/keys, etc.

MokListRT is empty.

lcp:
Would you mind to try my fix?
https://github.com/lcp/shim/commit/e6ce8788f4a622da1ba5421a5eb11df163a56727

martinezjavier:
@cyphermox I think PR #153 from @puiterwijk should fix this.

@lcp, this is very similar to your patch. Could you please review it?

lcp:
Yeah, PR#153 should fix this bug.

cyphermox:
Yup :)

PR #160, opened 2018-12-20
Fix MOK variables mirroring and duplicate shim_cert

lcp:
As mentioned in https://github.com/rhboot/shim/issues/154 and https://github.com/rhboot/shim/pull/153, the vendor certificate wasn't mirrored correctly. Besides, a faulty check also stopped several MOK variables from being mirrored.

During my testing, I also found that the auto-generated build certificate wasn't handled properly. Every user of shim_cert.h created its own duplicate shim_cert array. The build certificate was also not mirrored to MokListRT.

This PR should fix the mirroring issues above altogether.

### DIFF
--- a/mok.c
+++ b/mok.c
@@ -68,6 +68,10 @@ struct mok_state_variable {
 	 */
 	UINT8 **addend_source;
 	UINT32 *addend_size;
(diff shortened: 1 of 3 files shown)


vathpela:

PR #162, opened 2019-01-11
Fix arm build

jacmet:
The build is currently broken for 32bit ARM. Minor fixups to fix that.


vathpela:

PR #163, opened 2019-01-14
Makefile: do not run git on clean if there's no .git directory

bluca:
When building in minimal chroot on build workers, like in Debian (where
make clean is called at the beginning of the build process), git will
not be available. Skip the git clean.

Signed-off-by: Luca Boccassi <bluca@debian.org>


vathpela:

PR #169, opened 2019-02-15
Makefile: use fixed build host if SOURCE_DATE_EPOCH is defined

bluca:
If SOURCE_DATE_EPOCH is defined then we can be reasonably sure the
user wants the build to be fully reproducible, so use a fixed string.
In case of a cross build, using uname -s -m -p -i o will still report
the host's kernel architecture, which will trip some CIs like
Debian's.

Signed-off-by: Luca Boccassi <bluca@debian.org>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,6 +46,12 @@ ifneq ($(origin ENABLE_HTTPBOOT), undefined)
 	SOURCES += httpboot.c include/httpboot.h
 endif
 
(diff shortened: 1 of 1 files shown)


bjornfor:
Friendly bump. Will you merge this?

bluca:
This was merged directly in master via https://github.com/rhboot/shim/commit/11fd3197d21f94b491ccfc1da6d38b14060e62d7 closing. Thanks!

PR #171, opened 2019-03-05
shim: boot default loader when second stage file not found

Ivanhu5866:
When load options specify the second stage loader, but the file doesn't exist,
try to boot from default loader.

### DIFF
--- a/shim.c
+++ b/shim.c
@@ -2010,6 +2010,16 @@ EFI_STATUS init_grub(EFI_HANDLE image_handle)
 					 use_fb ? FALLBACK : second_stage);
 	}
 
+	/*
+	 * When load options specify the second stage loader, but the file
+         * doesn't exist, try to boot from default loader.
+	 */
+	if (efi_status == EFI_NOT_FOUND && !use_fb) {
+		console_print(L"Boot from default loader: %s\n", DEFAULT_LOADER);
+		msleep(2000000);
+		efi_status = start_image(image_handle, DEFAULT_LOADER);
+	}
+
 	if (EFI_ERROR(efi_status)) {
 		console_print(L"start_image() returned %r\n", efi_status);
 		msleep(2000000);


lcp:
If the machine is in Setup mode, there is no point to use MOK since the database cannot be trusted.

Ivanhu5866:
We consider that in factory processes, at the beginning, SecureBoot is disabled or in setup mode, and then SecureBoot will be enabled. At this time no users can enroll key with inputting password, and we need a key to be enrolled automatically, because after SecureBoot is on, some modules like DKMS will need to be verified.

lcp:
I understand that the factory would need some kind of auto-enrolling mechanism. However, writing MokList in shim makes me nervous. Would it be possible to create an one-shot boot option to load a special auto-enrolling EFI image? If it's for setup mode, you don't really need shim anyway.

Ivanhu5866:
We actually proposed the standalone UEFI application with one-shot boot option, like what you meat. https://github.com/Ivanhu5866/MokEnrollKey. However, due to some factory limitations, it is not allow to create extra boot option to do that.
It is not for setup mode, it is for getting ready for enabling secureboot, so need the key to be enrolled.
Base on UEFI specification, while SetupMode the platform firmware shall not require authentication in order to modify the keys and databases. So we think it might be reasonable to enroll key to MOK without not enforcing interactive password in SetupMode.

vathpela:
I don't see any problem with falling back to the default - I'm really not sure how that's even *relevant* to MoK enrollment, is github doing something weird here?

On the theory that this thread has gotten mangled in some weird way, I've pushed this as 4d8f27c85b16e4b48e6807b537b8c375e5226326 as well as 8ba7670e38bd2f83fbf4ebd1c4ff09d81e80d82f to make missing MokManager also fall back to default.

We probably ought to also fall back to doing a search for mm.efi in other directories like we do for BOOT.CSV in fallback, but I don't have the time to do that right now.

PR #173, opened 2019-03-08
VLogError(): Avoid NULL pointer dereferences in (V)SPrint calls

dannf:
VLogError() calculates the size of format strings by using calls to
SPrint and VSPrint with a StrSize of 0 and NULL for an output buffer.
Unfortunately, this is an incorrect usage of (V)Sprint. A StrSize
of "0" is special-cased to mean "there is no limit". So, we end up
writing our string to address 0x0. This was discovered because it
causes a crash on ARM where, unlike x86, it does not necessarily
have memory mapped at 0x0.

Replace the (V)SPrint calls with (V)PoolPrint calls, and we can use
StrLen() to calculate the size. Downside is that PoolPrint will
allocate pool memory to do this, so we need to release it.

Fixes: 25f6fd08cd26 ("try to show errors more usefully.")

### DIFF
--- a/errlog.c
+++ b/errlog.c
@@ -14,29 +14,20 @@ EFI_STATUS
 VLogError(const char *file, int line, const char *func, CHAR16 *fmt, va_list args)
 {
 	va_list args2;
-	UINTN size = 0, size2;
 	CHAR16 **newerrs;
-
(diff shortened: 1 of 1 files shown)


dannf:
@vathpela has a similar, but more efficient version. It works in my testing, so I'll update this PR to use his version.

PR #174, opened 2019-03-08
Fix OBJ_create() to tolerate a NULL sn and ln

dannf:
From: https://github.com/openssl/openssl/commit/f13615c5b828aeb8e3d9bf2545c803633d1c684f

Apply an upstream patch from OpenSSL to tolerate a NULL sn. This avoids
a NULL pointer reference in shim.c:verify_eku(). This was discovered
because it causes a crash on ARM where, unlike x86, it does not necessarily
have memory mapped at 0x0.

Fixes: 6c180c6004ac ("shim: verify Extended Key Usage flags")

### DIFF
--- a/Cryptlib/OpenSSL/crypto/objects/obj_dat.c
+++ b/Cryptlib/OpenSSL/crypto/objects/obj_dat.c
@@ -685,7 +685,8 @@ int OBJ_create(const char *oid, const char *sn, const char *ln)
     int ok = 0;
 
     /* Check to see if short or long name already present */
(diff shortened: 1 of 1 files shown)


vathpela:
Right now in my development tree, I have:
- https://github.com/vathpela/mallory/commit/ca18081a036fae1d7fc9888deca1b6b4fa298ac8
   - to always provide non-null values in case OpenSSL is broken
- https://github.com/vathpela/mallory/commit/356b8f887cb555157cc54ddb0c13ca3f0c516225
  - rebase openssl to 1.1.0h

vathpela:
This got pushed as 46b76a017173e7252469145af1be4933ee80954d quite some time ago.

PR #176, opened 2019-04-26
Use portable shebangs: /bin/bash -> /usr/bin/env bash

bjornfor:


vathpela:
I've pushed this to master as 6a73ca814af68fc8050cd56e2cf9a26c12bd4e71

bjornfor:
It took a while, but thanks for merging! 😀

PR #180, opened 2019-07-28
Remove undesired output dppath and path variables

howl:
When passing arguments to shim, it outputs the contents of dppath and path
variables. Seems to be a debug leftover.

Signed-off-by: David Santamaría Rogado <howl.nsp@gmail.com>

### DIFF
--- a/shim.c
+++ b/shim.c
@@ -2146,8 +2146,6 @@ static int is_our_path(EFI_LOADED_IMAGE *li, CHAR16 *path)
 		goto done;
 	}
 
-	dprint(L"dppath: %s\n", dppath);
-	dprint(L"path:   %s\n", path);
 	if (StrnCaseCmp(dppath, PathName, strlen(dppath)))
 		ret = 0;
 


cyphermox:
The path isn't a bad piece of information at all when you might be booting from various sources (the network, local disk, which disk, etc.). This only shows when SHIM_VERBOSE is enabled, too.

What is having this information shown on screen breaking? Seems like in a typical case, you would probably not want to have systems booting with SHIM_VERBOSE unless you're trying to debug things, in which case knowing that dppath and path are correct seems like a good thing.

howl:
I'm using the shim.efi for Fedora as is, so I should then compile myself.

I know it could be interesting to know it but I down know what kind of information could bring it. For example https://github.com/rhboot/shim/issues/181 I could find that it's incorrectly parsed because shim also outputs if it failed to execute itself followed by the argument.

cyphermox:
SHIM_VERBOSE is a runtime variable in UEFI firmware. You can run:

mokutil --set-verbosity false

No need to recompile for that.

howl:
> SHIM_VERBOSE is a runtime variable in UEFI firmware. You can run:
> 
> mokutil --set-verbosity false
> 
> No need to recompile for that.

Thanks.

howl:
@cyphermox I have tried mokutil but shim remains echoing both variables. I have a shim copy in hackbgrt folder in efi system partition, but I think that should matter as that variable is stored in nvram, or I'm wrong?

vathpela:
> @cyphermox I have tried mokutil but shim remains echoing both variables. I have a shim copy in hackbgrt folder in efi system partition, but I think that should matter as that variable is stored in nvram, or I'm wrong?

You can also do:
```
[root@trillian ~]# cd /sys/firmware/efi/efivars/
[root@trillian efivars]# chattr -i SHIM_VERBOSE-605dab50-e046-4300-abb6-3dd810dd8b23 
[root@trillian efivars]# rm SHIM_VERBOSE-605dab50-e046-4300-abb6-3dd810dd8b23 
rm: remove regular file 'SHIM_VERBOSE-605dab50-e046-4300-abb6-3dd810dd8b23'? y
```

PR #182, opened 2019-08-08
MokManager: console mode modification for hi-dpi screen devices

Ivanhu5866:
There are lots of hi-dpi laptops nowadays, as doing mok enrollment, the font
is too small to see.
https://bugs.launchpad.net/ubuntu/+source/shim/+bug/1822043

This patch checks if the resolution is larger than Full HD (1920x1080) and
current console output columns and rows is in a good mode. Then switch the
console output to a better mode.

### DIFF
--- a/MokManager.c
+++ b/MokManager.c
@@ -2491,6 +2491,8 @@ EFI_STATUS efi_main(EFI_HANDLE image_handle, EFI_SYSTEM_TABLE * systab)
 
 	setup_rand();
 
+	console_mode_handle();
+
 	efi_status = check_mok_request(image_handle);
 
 	console_fini();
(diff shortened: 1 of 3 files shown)


lcp:
Thanks for the work! I want to get this fixed long ago but I don't have any hi-dpi device at hand. This will improve MokManager a lot!

Ivanhu5866:
The patch is trying to switch to a better mode, mode number 2, which is the first non standard mode, which is provided by the device manufacturer, so it should be a good mode. And if it is failed, it will switch to the mode 0, which was "must" supported mode(80x25) defined by the UEFI spec.
In this case, mode 2 should be the best mode, since most of the firmwares only support 4 modes,
mode 0 (80x25)
mode 1(80x50)
mode 2(100x31)
mode 3(480x113)

vathpela:
I pushed this to master as cf05af6d899d4c00a70a61177b2a4932d5312129

PR #183, opened 2019-09-05
Fix "address of packed member" compilation error with GGC 9

diabonas:
When compiling with GCC 9, there are a couple of errors of the form
```
MokManager.c: In function ‘write_back_mok_list’:
MokManager.c:1056:19: error: taking address of packed member of ‘struct <anonymous>’ may result in an unaligned pointer value [-Werror=address-of-packed-member]
 1056 |   if (CompareGuid(&(list[i].Type), &X509_GUID) == 0)
      |                   ^~~~~~~~~~~~~~~
```
Copying the member of the packed struct to a temporary variable and pointing to that variable solves the problem.

### DIFF
--- a/MokManager.c
+++ b/MokManager.c
@@ -1038,6 +1038,7 @@ static EFI_STATUS write_back_mok_list(MokListNode * list, INTN key_num,
 	EFI_STATUS efi_status;
 	EFI_SIGNATURE_LIST *CertList;
 	EFI_SIGNATURE_DATA *CertData;
+	EFI_GUID type;
 	void *Data = NULL, *ptr;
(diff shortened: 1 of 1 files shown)


smcv:
https://github.com/rhboot/shim/commit/58532e12e9aa207e3c14096f4f5d6f7900130c52 appears to be an alternative solution to this.

diabonas:
@smcv 58532e12e9aa207e3c14096f4f5d6f7900130c52 is the patch proposed in this pull request, rebased onto the current master. I'm not sure why the PR wasn't closed in the process of cherry-picking the patch, doing so manually now.

PR #187, opened 2019-09-26
tpm: Don't log duplicate identical events

chrisccoulson:
According to the comment in tpm_measure_variable ("Don't measure something that we've already measured"), shim shouldn't measure duplicate events if they are identical, which also aligns with section 2.3.4.8 of the TCG PC Client Platform Firmware Profile Specification ("If it has been measured previously, it MUST NOT be measured again"). This is currently broken because tpm_data_measured() uses the return value of CompareGuid() incorrectly.


vathpela:
Pushed to master as 103adc89ce578a23cbdbd195c5dc5e329b85b854

PR #190, opened 2019-12-03
Update to openssl 1.1.1d

lcp:
This patch series updates the built-in openssl code to 1.1.1d. I didn't use the update script completely since the source of patches in the "openssl-rebase-helper" is not clear to me though I cherry-picked a few patches from it to fix the compilation errors.


vathpela:
For purely selfish reasons I don't want to go to 1.1.1d until RHEL does so for the normal openssl package, and I'm not yet sure when that'll be. You're defintely right about the rebase helper not being as helpful as I'd hoped.  What do you think about the strategy I've tried over in https://github.com/vathpela/mallory/tree/shim-16 which brings us to 1.1.1c, and switches us to using the upstream git versions more directly, so it's mostly just actual patch rebasing rather than all these crazy scripts?

vathpela:
Note that to make that branch work, you'll need a different layout for gnu-efi's installation, which the patches on this branch do: https://github.com/vathpela/gnu-efi/tree/fedora
And in order to use the OpenSSL tree it's pointing at and not crash immediately you'll specifically need this patch (which is also in that branch): https://github.com/vathpela/gnu-efi/commit/f08f27f5939559997c3177692348fdb7fcde7c89

lcp:
If we can use edk2 directly, it's definitely a better solution, and we can base on a specific edk2 stable release for the shim release. Will check the branch and see how it works.

lcp:
I cloned the mallory repo and tried to build shim-16. However, it seems efi.mk is missing. Also, would it be possible to make coverity optional? It's not a necessary service for the build service.

vathpela:
> I cloned the mallory repo and tried to build shim-16. However, it seems efi.mk is missing.

efi.mk is installed by the gnu-efi tree I mentioned above.

>  Also, would it be possible to make coverity optional? It's not a necessary service for the build service.
I don't understand - it's already not a default make target, what are you asking for?

lcp:
> > I cloned the mallory repo and tried to build shim-16. However, it seems efi.mk is missing.
> 
> efi.mk is installed by the gnu-efi tree I mentioned above.
> 
Sorry, my bad. I thought it's part of mallory.

> > Also, would it be possible to make coverity optional? It's not a necessary service for the build service.
> 
> I don't understand - it's already not a default make target, what are you asking for?

In include/coverity.mk, it checks whether cov-build exists in the system path or not, and this looks optional to me.

lcp:
I got some build errors while building mallory. It turned out that "space" wasn't defined in efi.mk. Anyway, I built shimx64.efi from mallory and will start to do some tests.

lcp:
I found Make.coverity and coverity.mk are identical. Just wonder why the check for COV_BUILD always happens. Anyway, here is the error message:

make TOPDIR=.././ BUILDDIR=.././/build-x64 -C .//build-x64 -f Makefile all
make[1]: Entering directory '/home/linux/git/mallory/build-x64'
.././/include/coverity.mk:43: *** cov-build not found.  Stop.
make[1]: Leaving directory '/home/linux/git/mallory/build-x64'
make: *** [Makefile:25: all] Error 2

A similar error also happens to scan-build.

lcp:
I figured out the root cause of the build errors and submitted a PR for mallory:
https://github.com/vathpela/mallory/pull/1

lcp:
I tested shimx64.efi from mallory, but somehow it failed to verify the signature. It failed at 'PKCS7_bio_add_digest()' because the NID of the digest algorithm was 0 (undefined). I wonder if something went wrong when constructing PKCS7 structure.

Besides, when enabling "-Os", shim crashed at the very beginning. I switched back to "-O0" to workaround it so that I can continue other tests.

vathpela:
Alright, so I need to figure out why verification isn't working, and also why -Os is weird, for sure.

1) Aside from those, are you okay with the submodule approach?
2) what compiler and linker versions were you seeing the -Os issue with?

lcp:
> Alright, so I need to figure out why verification isn't working, and also why -Os is weird, for sure.
> 
>     1. Aside from those, are you okay with the submodule approach?
> 
It looks fine to me. At least it's easier to maintain the patch for edk2 modules.

>     2. what compiler and linker versions were you seeing the -Os issue with?

gcc version 7.4.1 20190905 [gcc-7-branch revision 275407] (SUSE Linux) 
GNU ld (GNU Binutils; openSUSE Leap 15.1) 2.32.0.20190909-lp151.3.3

lcp:
I saw the shim-16 branch was merged. Will the change in gnu-efi(\*) be merged into gnu-efi upstream?
(\*) https://github.com/vathpela/gnu-efi/tree/fedora

vathpela:
This was auto-closed because I can't re-assign it to rhboot:main and for some reason it didn't follow when the default branch was renamed. @lcp can you please edit this to make it be against rhboot:main and re-open it?

PR #191, opened 2020-01-28
translate_slashes(): don't write to string literals

lersek:
Currently, all three invocations of the translate_slashes() function may
lead to writes to the string literal that is #defined with the
DEFAULT_LOADER_CHAR macro. According to ISO C99 6.4.5p6, this is undefined
behavior ("If the program attempts to modify such an array, the behavior
is undefined").

This bug crashes shim on e.g. the 64-bit ArmVirtQemu platform ("Data
abort: Permission fault"), where the platform firmware maps the .text
section (which contains the string literal) read-only.

Modify translate_slashes() so that it copies and translates characters
from an input array of "char" to an output array of "CHAR8".

While at it, fix another bug. Before this patch, if translate_slashes()
ever encountered a double backslash (translating it to a single forward
slash), then the output would end up shorter than the input. However, the
output was not NUL-terminated in-place, therefore the original string
length (and according trailing garbage) would be preserved. After this
patch, the NUL-termination on contraction is automatic, as the output
array's contents are indeterminate when entering the function, and so we
must NUL-terminate it anyway.

Fixes: 8e9124227d18475d3bc634c33518963fc8db7c98
Fixes: e62b69a5b0b87c6df7a4fc23906134945309e927
Fixes: 3d79bcb2651b9eae809b975b3e03e2f96c067072
Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1795654
Signed-off-by: Laszlo Ersek <lersek@redhat.com>

### DIFF
--- a/httpboot.c
+++ b/httpboot.c
@@ -743,14 +743,14 @@ httpboot_fetch_buffer (EFI_HANDLE image, VOID **buffer, UINT64 *buf_size)
 {
 	EFI_STATUS efi_status;
 	EFI_HANDLE nic;
-	CHAR8 *next_loader = NULL;
+	CHAR8 next_loader[sizeof DEFAULT_LOADER_CHAR];
 	CHAR8 *next_uri = NULL;
 	CHAR8 *hostname = NULL;
 
 	if (!uri)
 		return EFI_NOT_READY;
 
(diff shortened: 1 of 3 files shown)


vathpela:
This has been pushed to master as 9813e8bc8b3295f343809fac43298a73a93ffc97

PR #193, opened 2020-02-18
Include information about PE/COFF images in the TPM Event Log

martinezjavier:
The [TCG PC Client Specific Platform Firmware Profile Specification](https://trustedcomputinggroup.org/wp-content/uploads/TCG_PCClientSpecPlat_TPM_2p0_1p04_pub.pdf) says that when measuring a PE/COFF image, the TCG_PCR_EVENT2 structure Event field MUST contain a UEFI_IMAGE_LOAD_EVENT structure.
    
Currently an empty UEFI_IMAGE_LOAD_EVENT structure is passed so users only have the hash of the PE/COFF image, but not information such the file path of the binary.

The first commit just fixes a bug, and is that the EFI_LOADED_IMAGE FilePath field wasn't updated with the second stage loader path. So the wrong information would be logged if the value is used to fill the EFI_IMAGE_LOAD_EVENT DevicePath field.

The second commit fills the EFI_IMAGE_LOAD_EVENT structure to include the information about the PE/COFF image present in the TPM Event Log.

### DIFF
--- a/include/tpm.h
+++ b/include/tpm.h
@@ -10,8 +10,9 @@ EFI_STATUS tpm_log_event(EFI_PHYSICAL_ADDRESS buf, UINTN size, UINT8 pcr,
 			 const CHAR8 *description);
 EFI_STATUS fallback_should_prefer_reset(void);
 
-EFI_STATUS tpm_log_pe(EFI_PHYSICAL_ADDRESS buf, UINTN size, UINT8 *sha1hash,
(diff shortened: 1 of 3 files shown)


mjg59:
LGTM. Long term I guess we should figure out whether we have SHA256 banks and do something useful there.

vathpela:
I've pushed this to master as c252b9ee94c

PR #198, opened 2020-05-03
Add RISC-V (riscv32/riscv64) PE machine IDs

rfinnie:
c/t grub2 include/grub/efi/pe32.h


rfinnie:
FYI, admittedly this doesn't do much other than acknowledge RISC-V exists.  I've got a more complete (but very WIP) branch at https://github.com/rfinnie/shim/tree/riscv-more

PR #202, opened 2020-06-29
Add support for vendor_db built-in shim whitelist

aburmash:
Potential new signing strategies ( for example signing grub, fwupdate and vmlinuz with separate certificates )
require shim to support a vendor provided bundle of trusted certificates and hashes, which allows shim to "whitelist"
EFI binaries matching either certificate by signature, or hash in the vendor_db.
Functionality is similar to vendor_dbx ( vendor blacklist ).
Patch is a polished version of code, authored by P. Jones.


vathpela:
This is superceded by pr #203, so I'm closing it here.

PR #204, opened 2020-07-10
Support loading additional certificates

mjg59:
Based on an idea by @vathpela, support loading certificates from additional EFI binaries as long as those binaries are themselves signed by an already trusted certificate.

### DIFF
--- a/src/shim.c
+++ b/src/shim.c
@@ -1793,25 +1793,20 @@ static EFI_STATUS shim_read_header(void *data, unsigned int datasize,
 /*
  * Load and run an EFI executable
  */
-EFI_STATUS start_image(EFI_HANDLE image_handle, CHAR16 *ImagePath)
+EFI_STATUS read_image(EFI_HANDLE image_handle, CHAR16 *ImagePath,
+		      EFI_LOADED_IMAGE **li, CHAR16 *PathName, void **data,
+		      int *datasize)
 {
 	EFI_STATUS efi_status;
-	EFI_LOADED_IMAGE *li, li_bak;
-	EFI_IMAGE_ENTRY_POINT entry_point;
-	EFI_PHYSICAL_ADDRESS alloc_address;
-	UINTN alloc_pages;
-	CHAR16 *PathName = NULL;
 	void *sourcebuffer = NULL;
 	UINT64 sourcesize = 0;
-	void *data = NULL;
-	int datasize;
 
 	/*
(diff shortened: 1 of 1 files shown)


src/shim.c
```diff
@@ -1825,10 +1875,10 @@ EFI_STATUS start_image(EFI_HANDLE image_handle, CHAR16 *ImagePath)
 			       efi_status);
 			return efi_status;
 		}
-		data = sourcebuffer;
-		datasize = sourcesize;
+		*data = sourcebuffer;
+		*datasize = sourcesize;
 #if  defined(ENABLE_HTTPBOOT)
-	} else if (find_httpboot(li->DeviceHandle)) {
+	} else if (find_httpboot(*li->DeviceHandle)) {
```
aburmash, line 1881: } else if (find_httpboot((*li)->DeviceHandle)) {
 brackets missing ?

mjg59: Oh gosh, good catch.

src/shim.c
```diff
@@ -1880,7 +1895,7 @@ EFI_STATUS start_image(EFI_HANDLE image_handle, CHAR16 *ImagePath)
 	 */
 	efi_status = handle_image(data, datasize, li, &entry_point,
 				  &alloc_address, &alloc_pages);
-	if (EFI_ERROR(efi_status)) {
+	if (efi_status != EFI_SUCCESS) {
```
lcp, line 1954: This change looks not necessary.

mjg59: It's explained in the commit message - without this change, build fails with:

.././src/shim.c: In function ‘start_image’:
.././src/shim.c:1966:15: error: ‘entry_point’ may be used uninitialized in this function [-Werror=maybe-uninitialized]

lcp: Ah, thanks. Didn't realize this it related to the error mentioned in the commit message.

mjg59: I'll rewrite the commit message to clarify that, thank you!

vathpela:
This was auto-closed because I can't re-assign it to rhboot:main and for some reason it didn't follow when the default branch was renamed. @mjg59 can you please edit this to make it be against rhboot:main and re-open it?

mjg59:
I've moved on from Google and am not actively working on this at the moment, but for anyone else working on it: Google retains copyright to this code, but I was authorised to release it under the license terms covering the rest of Shim.

Issue #181, opened 2019-07-28
Arguments from EFI Shell incorrectly parsed


howl:
If we have shim in for example /EFI/test/shim.efi, executing cd /EFI/test; shim.efi custom_name.efi; makes shim to take "/EFI/test/shim.efi custom_name.efi" as parameter so it fails to load custom_name.efi as would be expected to do.

In the code is commented that EFI Shell passes this as strings separated with NULL character but in this situation space character is passed instead '20 00'. I think this situation should check for it and remove the own path leaving only "custom_name.efi as parameter".

It doesn't matter if we specify the preceding / or use the full path /EFI/test/custom_name.efi, the behavior is the same.

pcmoore:
@howl, can I ask on what system you ran into this problem?  I had a similar issue with a set of systems and put together a rough patch that so far has worked for me.  I'm planning on posting it here for possible inclusion, but if you are willing to try it out on your system (assuming it is different from my test system) I think that would be helpful.

pcmoore:
I'm not sure if you are still watching this @howl, but if you are interested I posted a fix for my test system in PR #225.

PR #225, opened 2020-10-05
BUG: attempt to improve the argument handling

pcmoore:
"So, load options are a giant pain in the ass." - Peter Jones

This patch attempts to workaround two LoadOption problems seen on
my test system: arguments separated with only whitespace (L" ") and
strings that aren't properly NULL terminated.

In addition to my test system, it appears at least one other person
has run into a similar problem:
* https://github.com/rhboot/shim/issues/181

### DIFF
--- a/shim.c
+++ b/shim.c
@@ -2337,6 +2337,10 @@ EFI_STATUS set_second_stage (EFI_HANDLE image_handle)
 		return efi_status;
 	}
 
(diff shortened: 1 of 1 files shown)


pcmoore:
Hmm.  Closed without merging and no comment?  I'm guessing this was either a mistake or an unintended consequence of some other action?  @vathpela?

PR #226, opened 2020-10-20
BUG: attempt to improve the argument handling

pcmoore:
"So, load options are a giant pain in the ass." - Peter Jones

This patch attempts to workaround two LoadOption problems seen on
my test system: arguments separated with only whitespace (L" ") and
strings that aren't properly NULL terminated.

In addition to my test system, it appears at least one other person
has run into a similar problem:
* https://github.com/rhboot/shim/issues/181

### DIFF
--- a/shim.c
+++ b/shim.c
@@ -1412,6 +1412,10 @@ EFI_STATUS set_second_stage (EFI_HANDLE image_handle)
 		return efi_status;
 	}
 
+	/* Sanity check since we make several assumptions about the length */
+	if (li->LoadOptionsSize % 2 != 0)
+		return EFI_INVALID_PARAMETER;
+
 	/* So, load options are a giant pain in the ass.  If we're invoked
 	 * from the EFI shell, we get something like this:
 
@@ -1505,6 +1509,26 @@ EFI_STATUS set_second_stage (EFI_HANDLE image_handle)
 	 */
 	UINTN strings = count_ucs2_strings(li->LoadOptions,
 					   li->LoadOptionsSize);
+
+	/*
+	 * In the case where strings == 1 check to see if L' ' is being used as
+	 * a delimiter, if so replace it with NULLs since the code already
+	 * handles that case.
+	 */
+	if (strings == 1) {
+		UINT16 *cur = li->LoadOptions;
+
+		/* replace L' ' with L'\0' if we find any */
+		for (i = 0; i < li->LoadOptionsSize / 2; i++) {
+			if (cur[i] == L' ')
+				cur[i] = L'\0';
+		}
+
+		/* redo the string count */
+		strings = count_ucs2_strings(li->LoadOptions,
+					     li->LoadOptionsSize);
+	}
+
 	/*
 	 * If it's not string data, try it as an EFI_LOAD_OPTION.
 	 */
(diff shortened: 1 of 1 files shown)


pcmoore:
This is a resubmit of a patch/PR that was originally submitted in PR #225.  It appears that the original PR was accidentally deleted when the maintainers removed/renamed a branch.

vathpela:
I think this look good, though is it possible to add a hex dump of the actual data in either a comment or the commit message?  I've been working on making unit tests, and when I add one for this code, I'd like to be able to test all the variations we know.

pcmoore:
> is it possible to add a hex dump of the actual data in either a comment or the commit message?

Sure, give me a day or two and I'll update the the PR.  Do you have a preference between the code, via comment, or patch description?  Both?

pcmoore:
Just in case, a quick update to the patch description, let me know if you are looking for something different.

```
commit c196097e2880fd276b3d219ca2f8710a0c30ab4d
Author: Paul Moore <pmoore2@cisco.com>
Date:   Mon Aug 24 13:17:58 2020 -0400

    shim: attempt to improve the argument handling
    
      "So, load options are a giant pain in the ass." - Peter Jones
    
    This patch attempts to workaround two LoadOption problems seen on
    my test system: arguments separated with only whitespace (L" ") and
    strings that aren't properly NULL terminated.
    
    In addition to my test system, it appears at least one other person
    has run into a similar problem and I can reproduce the whitespace
    delimiter issue with QEMU+OVMF (OEMU v5.2.0, OVMF v202011).
    
     * https://github.com/rhboot/shim/issues/181
    
    For reference, using QEMU+OVMF and a simple test application,
    "test.efi", which does a hexdump of LoadOptions I see the following
    when run via the UEFI shell:
    
     FS0:\> test.efi one two three
     LoadOptions:
     0000: 74 00 65 00 73 00 74 00 2E 00 65 00 66 00 69 00 t.e.s.t...e.f.i.
     0016: 20 00 6F 00 6E 00 65 00 20 00 74 00 77 00 6F 00 ..o.n.e...t.w.o.
     0032: 20 00 74 00 68 00 72 00 65 00 65 00 00 00       ..t.h.r.e.e...
    
```

shim.c
```diff
@@ -2442,68 +2466,47 @@ EFI_STATUS set_second_stage (EFI_HANDLE image_handle)
...
-	if (strings < 1)
-		return EFI_SUCCESS;
+		while (start[loader_len++] != L'\0');
+		loader_len *= 2;
 
-	/*
-	 * And then I found a version of BDS that gives us our own path in
-	 * LoadOptions:
+		remaining_size -= loader_len;
+	} else if (strings == 1 && is_our_path(li, start))
```
martinezjavier, line 2498: can you please add braces here ? It will make the code less error prone.

shim.c
```diff
@@ -2442,68 +2466,47 @@ EFI_STATUS set_second_stage (EFI_HANDLE image_handle)
...
 77162C58                           5c 00 45 00 46 00 49 00          |\.E.F.I.|
 77162C60  5c 00 42 00 4f 00 4f 00  54 00 5c 00 42 00 4f 00  |\.B.O.O.T.\.B.O.|
 77162C70  4f 00 54 00 58 00 36 00  34 00 2e 00 45 00 46 00  |O.T.X.6.4...E.F.|
 77162C80  49 00 00 00                                       |I...|
 
-	 * which is just cruel... So yeah, just don't use it.
-	 */
+		* which is just cruel... So yeah, just don't use it.
+		*/
 	if (strings == 1 && is_our_path(li, start))
```
martinezjavier, line 2510: Don't you need to remove this line since you are already checking for this exact condition in the `else if` clause ?

shim.c
pcmoore: Sure thing, fixed in the next force push.

martinezjavier (approved):
@pcmoore looks good to me now, thanks!

PR #227, opened 2020-10-20
BUG: add some basic $EFI_PATH checking

pcmoore:
Not all distributions put the crt0-efi-$(ARCH).o file under
$LIB_DIR/gnuefi, some stash it directly in $LIB_DIR.  In an effort
to make the build a bit more user friendly, check if $LIB_DIR/gnuefi
exits before setting $EFI_PATH to that value; if $LIB_DIR/gnuefi does
not exist, fallback to $LIB_DIR for $EFI_PATH.


vathpela:
I've pushed this to main, rather than to 15.2.

PR #229, opened 2020-10-21
Add a stub for SBAT

vathpela:


vathpela:
For now we can work on the documentation for this feature here; if it grows out of this that's fine but it seems like a good way to start.

PR #232, opened 2020-10-26
Apply the EKU checks

lcp:
Implement the CodeSign EKU check to fulfill the requirements of NIAP OS_PP.
Also modify the ModSign EKU check to use VerifyEKUsInPkcs7Signature() to check the signer certificate instead of the certificate directly from the key database.

### DIFF
--- a/Cryptlib/Pk/CryptPkcs7VerifyCommon.c
+++ b/Cryptlib/Pk/CryptPkcs7VerifyCommon.c
@@ -24,6 +24,7 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 
 UINT8 mOidValue[9] = { 0x2A, 0x86, 0x48, 0x86, 0xF7, 0x0D, 0x01, 0x07, 0x02 };
 CHAR8 mOidCodeSign[] = "1.3.6.1.5.5.7.3.3";
+CHAR8 mOidModSign[] = "1.3.6.1.4.1.2312.16.1.2";
 
 BOOLEAN ca_warning;
 
@@ -816,6 +817,7 @@ Pkcs7Verify (
   UINTN       SignedDataSize;
   BOOLEAN     Wrapped;
   CONST CHAR8 *RequiredEku[1];
+  CONST CHAR8 *RejectedEku[1];
   EFI_STATUS  EFI_Status;
 
   //
@@ -831,6 +833,7 @@ Pkcs7Verify (
   Cert      = NULL;
   CertStore = NULL;
   RequiredEku[0] = mOidCodeSign;
+  RejectedEku[0] = mOidModSign;
 
   //
   // Register & Initialize necessary digest algorithms for PKCS#7 Handling
@@ -928,6 +931,15 @@ Pkcs7Verify (
   // Bypass the certificate purpose checking by enabling any purposes setting.
   //
   X509_STORE_set_purpose (CertStore, X509_PURPOSE_ANY);
+
+  //
+  // Reject the signature if the signer certificate contains the rejected EKUs
+  // such as ModSign.
+  //
+  EFI_Status = VerifyEKUsInPkcs7Signature(P7Data, P7Length, RejectedEku, 1, FALSE);
+  if (EFI_Status == EFI_SUCCESS) {
+    goto _Exit;
+  }
 
   //
   // Make sure the signer certificate contains the required EKUs such as CodeSign.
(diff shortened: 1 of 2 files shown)


Cryptlib/Pk/CryptPkcs7VerifyCommon.c
```diff
@@ -814,6 +815,8 @@ Pkcs7Verify (
   CONST UINT8 *Temp;
   UINTN       SignedDataSize;
   BOOLEAN     Wrapped;
+  CONST CHAR8 *RequiredEku[1];
```
out0xb2, line 819: Should EKU be a parameter to this function instead, along with a size (for multiple EKUs required), or push the EKU verification up a level ?  The parameter path is probably cleaner, but deviates from the API surface used in UEFI BaseCrypt (where these are separate functions).  Cleanest might be to add a Pkcs7VerifyWithEku() that calls Pkcs7Verify() and VerifyEKUsInPkcs7Signature(), and the supplied EKU list can optionally be NULL (but probably won't be in your case).  Perhaps reference https://github.com/microsoft/mu_plus/blob/release/202008/MfciPkg/Private/Library/MfciPolicyParsingLib/MfciPolicyParsingLib.c function ValidateSignature() (which limits to 1 EKU currently).

Cryptlib/Pk/CryptPkcs7VerifyCommon.c
```diff
@@ -925,6 +932,23 @@ Pkcs7Verify (
   //
   X509_STORE_set_purpose (CertStore, X509_PURPOSE_ANY);
 
+  //
+  // Reject the signature if the signer certificate contains the rejected EKUs
+  // such as ModSign.
+  //
+  EFI_Status = VerifyEKUsInPkcs7Signature(P7Data, P7Length, RejectedEku, 1, FALSE);
+  if (EFI_Status == EFI_SUCCESS) {
+    goto _Exit;
```
out0xb2, line 941: Should EFI_Status be changed to EFI_SECURITY_VIOLATION?

Cryptlib/Pk/CryptPkcs7VerifyCommon.c
```diff
@@ -925,6 +932,23 @@ Pkcs7Verify (
...
+  if (EFI_Status == EFI_SUCCESS) {
+    goto _Exit;
+  }
+
+  //
+  // Make sure the signer certificate contains the required EKUs such as CodeSign.
+  //
+  EFI_Status = VerifyEKUsInPkcs7Signature(P7Data, P7Length, RequiredEku, 1, TRUE);
+  if (EFI_Status != EFI_SUCCESS) {
+    goto _Exit;
```
out0xb2, line 949: EFI_Status may be EFI_NOT_FOUND.  Should change to EFI_SECURITY_VIOLATION ?

src/shim.c
```diff
@@ -360,25 +360,23 @@ static CHECK_STATUS check_db_cert_in_ram(EFI_SIGNATURE_LIST *CertList,
 			CertSize = CertList->SignatureSize - sizeof(EFI_GUID);
 			dprint(L"trying to verify cert %d (%s)\n", i++, dbname);
 			if (verify_x509(Cert->SignatureData, CertSize)) {
-				if (verify_eku(Cert->SignatureData, CertSize)) {
```
out0xb2, line 363: I prefer the EKU check being here versus pushed down and hidden inside the crypto library, business logic should be at this level, not hidden in the crypto library IMHO...

out0xb2: Perhaps verify_x509() could have an EKU parameter?  Defer to @vathpela and @mjg59

vathpela:
@lcp, can you please edit this to make it be against rhboot:main and re-open it?

Cryptlib/Pk/CryptPkcs7VerifyCommon.c
julian-klode: Shouldn't this be `!= EFI_SUCCESS` too?

lcp: This part checks the rejected EKU, and `Pkcs7Verify()` should return `FALSE` if the rejected EKU exists.
Besides, since `EFI_Status` isn't returned to the caller, we just need to make sure `Status` is still `FALSE`.

lcp:
This patch series could break the existing certificates, so I hold it as a SUSE-ONLY patch for now.

PR #233, opened 2020-10-29
RFE: support a stable PCR across firmware updates, enable booting a kernel directly

pcmoore:
This PR is focused on creating a PCR which measures the kernel's signing authority while remaining stable across firmware updates.  While PCR7 is close to what we need, it measures the `db` and `dbx` variables which can change during a firmware update.  As noted in one of the patch descriptions:

    The problem is that PCR7 measures the db and dbx variables which can
    change during a UEFI/firmware update as new values are added, moved,
    and removed.  For those use cases which require PCR stability across
    firmware updates, we need a different solution.
    
    This patch attempts to satisfy this need by extending PCR8 in a
    similar manner, measuring most of what is measured into PCR7 but not
    the db and dbx variables.  PCR8 is extended with the SecureBoot, PK,
    KEK, and the db certificate that authorized the second stage loader.
    While the db and dbx variables are not measured, measuring the PK
    and KEK helps ensure that db and dbx updates (if any) have been
    authorized.

Using PCR8 is not without its challenges, the largest being that the GRUB2 which ships with most Linux distributions makes use of PC8 (and PCR9 for that matter).  If we are to leverage PCR8 we need to do so without GRUB2, or at least without a GRUB2 build that extends PCR8.  Thankfully UEFI provides its own bootloader functionality which is sufficient for many systems.  The only drawback to booting a Linux Kernel directly from shim is the `ExitBootServices()` hook which requires that the second stage loader (or kernel in the case described here) calls into the shim verification protocol.  However, in our case we can safely skip this check:

    On systems where a second stage bootloader is not used, and the Linux
    Kernel is booted directly from shim, shim's ExitBootServices() hook
    can cause problems as the kernel never calls the shim's verification
    protocol.  In this case calling the shim verification protocol is
    unnecessary and redundant as shim has already verified the kernel
    when shim loaded the kernel as the second stage loader.

The concepts in this PR were discussed, albeit briefly, on the EFI mailing list at the link below:

* https://lists.einval.com/pipermail/efi/2020-October/000011.html

### DIFF
--- a/Make.defaults
+++ b/Make.defaults
@@ -109,6 +109,10 @@ ifneq ($(origin ENABLE_PCR8_SIGNER), undefined)
 	CFLAGS  += -DENABLE_PCR8_SIGNER
 endif
 
+ifneq ($(origin ENABLE_PCR9_LOADER), undefined)
+	CFLAGS  += -DENABLE_PCR9_LOADER
+endif
+
 LIB_GCC		= $(shell $(CC) $(ARCH_CFLAGS) -print-libgcc-file-name)
 EFI_LIBS	= -lefi -lgnuefi --start-group Cryptlib/libcryptlib.a Cryptlib/OpenSSL/libopenssl.a --end-group $(LIB_GCC)
 FORMAT		?= --target efi-app-$(ARCH)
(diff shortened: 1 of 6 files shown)


pcmoore:
There still may be a valid reason to bypass the `ExitBootServices()` check, but the PCR8 and PCR9 extensions are a no-go for reasons described in the mailing list thread below, I'm self-NACK'ing this PR as a result.
* https://lists.einval.com/pipermail/efi/2020-October/000011.html

PR #235, opened 2020-10-31
Native UEFI shim

rusnikola:
Updated my fork (which generates true PE/COFF images) to the most recent version. Since edk2 headers are also adopted by the original shim now, code changes are reduced. Latest clang/llvm (with required flags) are already in common Linux distributions. Would it make sense to consider those changes for upstreaming? It is also possible to provide it as an additional option (in case if gcc compatibility is desired).


vathpela:
This was auto-closed because I can't re-assign it to rhboot:main and for some reason it didn't follow when the default branch was renamed. @rusnikola can you please edit this to make it be against rhboot:main and re-open it?

That said, I think other developers had some significant misgivings about this approach - mostly that edk2's build system is terrible - but I don't want to lose track of it.

rusnikola:
@vathpela thanks for letting me know. btw, edk2 is not that bad if you just use their headers only but not Makefiles, etc :)
I can relocate the changes but I have a question about repositories because I am not sure if I need to use "devel" or "main".

What happened to the "master" branch and those experimental changes (e.g., edk2)? Is it just completely abandoned?
The reason why I ask is because there is also the "devel" branch which has edk2 but is behind "main". (btw, what is the difference between "devel" and old "master"?)

vathpela:
> @vathpela thanks for letting me know. btw, edk2 is not that bad if you just use their headers only but not Makefiles, etc :)
> I can relocate the changes but I have a question about repositories because I am not sure if I need to use "devel" or "main".
> 
> What happened to the "master" branch and those experimental changes (e.g., edk2)? Is it just completely abandoned?
>
> The reason why I ask is because there is also the "devel" branch which has edk2 but is behind "main". (btw, what is the difference between "devel" and old "master"?)

"devel" is the old master, and I'm working on backporting and merging to a new main while some other development happens.  If using the edk2 submodule from devel makes it easier, do it against that - either is fine with me.

rusnikola:
@vathpela yes, I already relocated my last version several months ago for the old master (which is now "devel"). So basically my version should already work on top of devel (including devel's edk2 module). Should I just update the pull request for devel or is it going to be abandoned eventually (e.g., the edk2 module going to be abandoned, etc)

So, in other words, will you have edk2 in the "main" repo eventually? If so, probably it makes sense for me to keep it for devel for now and when those changes will be committed to main, I can relocate my version as well.

PR #237, opened 2020-11-10
fix EV_EFI_VARIABLE_AUTHORITY event in eventlog

huang195:
Currently, for an EV_EFI_VARIABLE_AUTHORITY event, the shim puts only EFI_SIGNATURE_DATA.SignatureData in the VariableData field, but omits EFI_SIGNATURE_DATA.SignatureOwner. According to reference implementation in EDK2, the entire EFI_SIGNATURE_DATA is put into the VariableData field, shown here:
https://github.com/tianocore/edk2/blob/master/SecurityPkg/Library/DxeImageVerificationLib/DxeImageVerificationLib.c#L1032


vathpela:
I've pushed this to main, rather than to 15.2.

PR #248, opened 2021-01-14
RFE: compile time option to bypass the ExitBootServices() check

pcmoore:
On systems where a second stage bootloader is not used, and the Linux Kernel is booted directly from shim, shim's ExitBootServices() hook can cause problems as the kernel never calls the shim's verification protocol.  In this case calling the shim verification protocol is unnecessary and redundant as shim has already verified the kernel when shim loaded the kernel as the second stage loader.

This functionality is disabled by default and must be enabled via the DISABLE_EBS macro/define at build time.

### DIFF
--- a/Make.defaults
+++ b/Make.defaults
@@ -105,6 +105,10 @@ ifneq ($(origin REQUIRE_TPM), undefined)
 	CFLAGS  += -DREQUIRE_TPM
 endif
 
+ifneq ($(origin DISABLE_EBS), undefined)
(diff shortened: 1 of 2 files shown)


vathpela:
I've pulled this as 4b0a61dc9a952028c7b1ccc8914835d989ce661e with one minor change - the phrasing of the option was confusing to me, so I changed it to DISABLE_EBS_PROTECTION.

PR #253, opened 2021-01-31
Sbat copy edit

steve-mcintyre:
Can't stop myself from copy-editing this kind of doc... :-)


vathpela:
I *think* I got everything here when merging https://github.com/rhboot/shim/pull/262 .  It should all be reflected in https://github.com/rhboot/shim/commit/4eef10a6b345fad26578c985b8b6ab38ca2025e7 , but the documents had diverged a bit.  Best to give it another once-over and re-open this if I'm wrong.

PR #254, opened 2021-02-03
Add the SBAT example workflow that we discussed

vathpela:


SBAT.example.md
```diff
@@ -0,0 +1,224 @@
...
+- When validating SBAT, any `component_name` that's in both `SBAT` and `.sbat` gets compared
+- `component_generation` in `.sbat` must be >= the highest `component_generation` of an identical `component_name` in `SBAT`
+- that version comparison includes the `"sbat"` entry
+
+Example universe starting point
+-------------------------------
+For grub, a build from a fresh checkout of upstream might have the following in
+`.sbat`:
+```
+sbat,0,UEFI shim,0,https://github.com/rhboot/shim/blob/main/SBAT.md
```
martinezjavier, line 45: This entry is missing a `vendor_package_name`. Should it be `sbat` again or ?

vathpela: yep.  fixed in latest push.

jsetje (commented):
sbat version should be called out as the sbat version and be set to 1:

Instead of:

sbat,0,UEFI shim,sbat,0,https://github.com/rhboot/shim/blob/main/SBAT.md

(which could be confused with the shim generation number)

I went with:

sbat,1,SBAT version,sbat,0,https://github.com/rhboot/shim/blob/main/SBAT.md

SBAT.example.md
```diff
@@ -0,0 +1,217 @@
...
+  { 0, 5, "shim" },
+  { 0, 5, "grub" },
+  { 0, 12, "grub.fedora" },
+}
+```
+
+Along comes bug 0
+-----------------
+Let's say we find a bug in Fedora that's serious, but it's only in Fedora's
+patches, because we're the dumb ones.  Fedora issues a new shim build with the
```
christopherco, line 104: Would Fedora issue a new `shim` build or `grub` build with the new `.sbat` info?

lcp:
Just out of curiosity, will there be any example covering the SBAT version bumping in SBAT variable?

SBAT.example.md
vathpela: That would be a grub update.

vathpela:
> Just out of curiosity, will there be any example covering the SBAT version bumping in SBAT variable?

I don't think that we know enough about /why/ we would do that to write meaningful examples, though in my mind any format update will either be because of one of:
- we decide there's another field we need to require and add to the csv.  in this case we add data fields.
- upstream linux rejects our plans on how to do this and we come up with something else, and we're moving to that.  in this case we should probably pick a new name and not bump the version.
- upstream EDK2 rejects our plans on how to do this and we come up with something else, and we're moving to that.  in this case we should probably pick a new name and not bump the version.

PR #255, opened 2021-02-05
SBAT pull-request CI builds

vathpela:
This series of patches does two things:
- it re-works the PR CI builder so all the test targets use the same build steps
- it adds the sbat branch to the list of branches to build PRs against.


vathpela:
Since all the CI worked, and even tested the thing I thought it would be harder to test (namely, did adding CI work), I'm gonna merge https://github.com/rhboot/shim/pull/255 unless someone has objections within about 15 minutes

PR #256, opened 2021-02-09
Some minor fixes for SBAT.md

lcp:
* Remove the duplicate sentence
* Replace `debian.grub` with `grub.debian`


vathpela:
I've merged this in with https://github.com/rhboot/shim/pull/262 in with this and pushed them both to main.

PR #258, opened 2021-02-09
We need some strchr/strchrnul functions for SBAT parsing, so add them here.

vathpela:
This is why we can't have nice things?


martinezjavier (approved):
LGTM

PR #259, opened 2021-02-12
Add a .sbat section to EFI binaries

martinezjavier:
The Secure Boot Advanced Targeting (SBAT) [0] is a Generation Number Based
Revocation mechanism that is meant to replace the DBX revocation file list.

Binaries must contain a .sbat data section that has a set entries, each of
them consisting of UTF-8 strings as comma separated values. Allow to embed
this information into the fwupd EFI binary at build time.

The SBAT metadata must contain at least two entries. One that defines the
SBAT version used and another one that defines the component generation.

Downstream users can add additional entries if have changes that make them
diverge from the upstream code and potentially add other vulnerabilities.

The reason why an intermediate step is needed, is because the objcopy tool
only has support for pei-x86_64 and pei-i386 targets. So the .sbat section
has to be added to the ELF shared objects instead of the PE32+ binaries.

Signed-off-by: Javier Martinez Canillas <javierm@redhat.com>

### DIFF
--- a/Make.defaults
+++ b/Make.defaults
@@ -22,6 +22,7 @@ DEBUGSOURCE	?= $(prefix)/src/debug/
 OSLABEL		?= $(EFIDIR)
 DEFAULT_LOADER	?= \\\\grub$(ARCH_SUFFIX).efi
 DASHJ		?= -j$(shell echo $$(($$(grep -c "^model name" /proc/cpuinfo) + 1)))
+SBATPATH	?= data/sbat.csv
 
 ARCH		?= $(shell $(CC) -dumpmachine | cut -f1 -d- | sed s,i[3456789]86,ia32,)
 OBJCOPY_GTE224	= $(shell expr `$(OBJCOPY) --version |grep ^"GNU objcopy" | sed 's/^.*\((.*)\|version\) //g' | cut -f1-2 -d.` \>= 2.24)
--- a/Makefile
+++ b/Makefile
@@ -33,12 +33,12 @@ CFLAGS += -DENABLE_SHIM_CERT
 else
 TARGETS += $(MMNAME) $(FBNAME)
 endif
-OBJS	= shim.o mok.o netboot.o cert.o replacements.o tpm.o version.o errlog.o
+OBJS	= shim.o mok.o netboot.o cert.o replacements.o tpm.o version.o errlog.o sbat.o
 KEYS	= shim_cert.h ocsp.* ca.* shim.crt shim.csr shim.p12 shim.pem shim.key shim.cer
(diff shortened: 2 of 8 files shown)


Make.defaults
```diff
@@ -22,6 +22,7 @@ DEBUGSOURCE	?= $(prefix)/src/debug/
 OSLABEL		?= $(EFIDIR)
 DEFAULT_LOADER	?= \\\\grub$(ARCH_SUFFIX).efi
 DASHJ		?= -j$(shell echo $$(($$(grep -c "^model name" /proc/cpuinfo) + 1)))
+SBATPATH	?= sbat.csv
```
vathpela, line 25: Is the intent here for shim.efi, mm.efi, and fb.efi to all always have the "shim" sbat entry?

Makefile
```diff
@@ -185,15 +185,18 @@ else
 	$(INSTALL) -m 0644 $(FBNAME) $(DESTDIR)/$(DATATARGETDIR)/$(FBNAME)
 endif
 
-%.efi: %.so
+%.so.sbat : %.so
```
vathpela, line 188: I would rather see this make a .o, have that be linked into the .so file, and place it explicitly after .vendor_cert in the `elf_ARCH_efi.lds` file.  It's also going to need `_sbat` and `_esbat` symbols defined in the `elf_ARCH_efi.lds` file, so we can find it in the self-check.

vathpela (changes requested):
This is a good start, but it doesn't give us a way to identify the region for the shim self-check, and it gives mm.efi and fb.efi the same `.sbat` as shim has (which may be correct, but I'd like to see it stated in the commit message and maybe one of the README.md / SBAT.md files).

Is there a reason not to also include the actual sbat.csv file in this PR?

Make.defaults
martinezjavier: I think so, this was deliberately because we want a) fallback and mok manager to be allowed when SBAT is enforced and b) these to be revoked as well if a shim is revoked.

The alternative would had been to have different component names and generation number for fb and mm, but I thought that would be adding complexity for no good reason.

martinezjavier:
> Is there a reason not to also include the actual sbat.csv file in this PR?

Not really. Just that thought downstream users would have to add additional entries anyways. But I guess that will be much better to add a reference sbat.csv that has the `sbat` and `shim` entries and downstream users could override / extend if needed.

If downstream users ship an unmodified upstream shim, then that should be enough for them.

Makefile
martinezjavier: Gotcha. I'll do these changes.

vathpela (approved):

PR #260, opened 2021-02-12
efi bins: add an easy way for vendors to add .sbat data

vathpela:
In cases where we accept vendor shim binaries with additional patches,
it may become necessary to identify those builds with additional SBAT
data.  When we consider such patches, we should be proactive in asking
vendors to include that data in the .sbat sections of their trusted EFI
binaries.

This patch adds any data in data/sbat.*.csv (after a quick sanitizing
pass) after data/sbat.csv in the .sbat section, so  that no changes to
the upstream data/sbat.csv are ever required.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,6 +40,7 @@ MOK_OBJS = MokManager.o PasswordCrypt.o crypt_blowfish.o errlog.o sbat.o
 ORIG_MOK_SOURCES = MokManager.c PasswordCrypt.c crypt_blowfish.c shim.h $(wildcard include/*.h)
 FALLBACK_OBJS = fallback.o tpm.o errlog.o sbat.o
 ORIG_FALLBACK_SRCS = fallback.c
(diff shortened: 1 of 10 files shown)


Makefile
```diff
@@ -40,6 +40,7 @@ MOK_OBJS = MokManager.o PasswordCrypt.o crypt_blowfish.o errlog.o sbat.o
 ORIG_MOK_SOURCES = MokManager.c PasswordCrypt.c crypt_blowfish.c shim.h $(wildcard include/*.h)
 FALLBACK_OBJS = fallback.o tpm.o errlog.o sbat.o
 ORIG_FALLBACK_SRCS = fallback.c
+SBATPATH = data/sbat.csv
```
martinezjavier, line 43: Is this needed if is already defined in `Make.defaults` ?

martinezjavier (approved):

PR #261, opened 2021-02-12
Add support to parse the .sbat section from loaded binaries

martinezjavier:
Now that the SBAT metadata is included in a `.sbat` section of the built binaries,
allow shim to parse this information.

This pull-request includes a few preparatory patches and a function to parse the
`.sbat` section. This data is currently still not used but will be in following changes.


vathpela:
I've merged this with a couple of fixes to my wrong SPDX headers.

PR #262, opened 2021-02-13
Add SBAT specification and examples document

martinezjavier:
This pull-request adds a `SBAT.md` document that specifies the Secure Boot Advanced Targeting (SBAT)
revocation mechanism that is meant to replace the DBX Revocation List Files mechanism.

It also adds an `SBAT.examples.md` document that contains an example workflow and better illustrates
what type of content is expected to be present in both the `.sbat` section and the `SBAT` authenticated EFI variable.

### DIFF
--- a/SBAT.example.md
+++ b/SBAT.example.md
@@ -1,0 +1,217 @@
+SBAT: Current proposal
+================
+
+`.sbat` section
+-------------
+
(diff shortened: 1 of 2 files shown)


vathpela:
I've merged https://github.com/rhboot/shim/pull/256 in with this and pushed them both to main.

PR #267, opened 2021-02-14
Parse SBAT variable and perform basic verification

aburmash:
Per @vathpela suggestion, we will be flexible in what data we expect while parsing the variable.
 Three fields are mandatory:
 component_generation, component_name_size, component_name
 However we also support adding comments and additional information to be added after component name, with
 ',' as a separator. Those information will be ignored and not used for verification purposes.
 So:
 1,5,grub\n
 1,5,grub,wow,this,is,my,comment\n
 will provide exactly same set of data for verification.

Reference: https://github.com/rhboot/shim/blob/main/SBAT.md

### DIFF
--- a/include/sbat.h
+++ b/include/sbat.h
@@ -8,6 +8,16 @@
 
 #include "shim.h"
 
+struct sbat_var {
+	const CHAR8 *component_generation;
+	const CHAR8 *component_name_size;
+	const CHAR8 *component_name;
+	list_t list;
+};
+
+#define for_each_sbat_var(sbat, head) list_for_each(sbat, head)
+#define for_each_sbat_var_safe(sbat, n, head) list_for_each_safe(sbat, n, head)
+
 struct sbat_entry {
 	const CHAR8 *component_name;
 	const CHAR8 *component_generation;
@@ -21,6 +31,10 @@ struct sbat {
 	unsigned int size;
 	struct sbat_entry **entries;
 };
+
+EFI_STATUS parse_sbat_var(list_t *entries);
+
+EFI_STATUS verify_sbat(struct sbat *sbat, list_t *entries);
 
 EFI_STATUS parse_sbat(char *sbat_base, size_t sbat_size, char *buffer,
 		      struct sbat *sbat);
--- a/include/str.h
+++ b/include/str.h
@@ -54,7 +54,7 @@ strndupa(const CHAR8 * const src, const UINTN srcmax)
 		return news;
 
 	len = strnlena(src, srcmax);
-	news = AllocateZeroPool(len);
+	news = AllocateZeroPool(len+1);
 	if (news)
 		strncpya(news, src, len);
 	return news;
--- a/pe.c
+++ b/pe.c
@@ -1043,6 +1043,8 @@ handle_image (void *data, unsigned int datasize,
 		unsigned int i;
 		struct sbat sbat = { 0 };
 		struct sbat_entry *entry = NULL;
+		list_t sbat_var_entries;
+		INIT_LIST_HEAD(&sbat_var_entries);
 
 		if (SBATBase && SBATSize) {
 			char *sbat_data;
@@ -1078,13 +1080,24 @@ handle_image (void *data, unsigned int datasize,
 			perror(L"SBAT data not found\n");
 			return EFI_UNSUPPORTED;
 		}
-
+		
+		efi_status = parse_sbat_var(&sbat_var_entries);
+		/* we get EFI_SUCCESS only if variable is fully parsed,
+		   and EFI_NOT FOUND means do not perform verificaiton
+		   continue with EFI_SUCCESS
+		*/
+		if (efi_status == EFI_SUCCESS)
+			efi_status = verify_sbat(&sbat, &sbat_var_entries);
+		else if (efi_status == EFI_NOT_FOUND)
+			efi_status = EFI_SUCCESS;
+		
 		efi_status = verify_buffer(data, datasize,
 					   &context, sha256hash, sha1hash);
 
-		if (sbat.entries)
+		if (sbat.entries) {
 			for (i = 0; i < sbat.size; i++)
 				FreePool(sbat.entries[i]);
+		}
 
 		if (EFI_ERROR(efi_status)) {
 			if (verbose)
--- a/sbat.c
+++ b/sbat.c
@@ -4,6 +4,7 @@
  */
 
 #include "sbat.h"
+#include <string.h>
 
 CHAR8 *
 get_sbat_field(CHAR8 *current, CHAR8 *end, const CHAR8 ** field, char delim)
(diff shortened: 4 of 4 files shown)


include/sbat.h
```diff
@@ -8,6 +8,16 @@
```
martinezjavier, line 8: I wish Github had a way to comment on the commit messages, so I'll comment here.

shim me more or less follow what's described in https://www.kernel.org/doc/html/v4.17/process/submitting-patches.html#the-canonical-patch-format, please flollow that. Specially line wrapping at ~75 columns.

Also I think it's better if you use Peter Jones instead of @vathpela in the commit message.

include/sbat.h
```diff
@@ -22,8 +32,12 @@ struct sbat {
 	struct sbat_entry **entries;
 };
 
+int parse_sbat_var(list_t *entries);
+
+EFI_STATUS verify_sbat(struct sbat *sbat, list_t *entries);
+
 EFI_STATUS parse_sbat(char *sbat_base, size_t sbat_size, char *buffer,
-		      struct sbat *sbat);
+                      struct sbat *sbat);
```
martinezjavier, line 40: This seems to be an unrelated change.

sbat.c
```diff
@@ -120,4 +123,161 @@ EFI_STATUS parse_sbat(char *sbat_base, size_t sbat_size, char *buffer,
...
+			FreePool((CHAR8 *)entry->component_generation);
+		if (entry->component_name_size)
+			FreePool((CHAR8 *)entry->component_name_size);
+		if (entry->component_name)
+			FreePool((CHAR8 *)entry->component_name);
+		FreePool(entry);
+	}
+}
+
+int
```
martinezjavier, line 210: I would use `EFI_STATUS` instead here. That way you could propagate the errors returned by EFI services calls and also would be consistent with other functions in this file.

pe.c
```diff
@@ -1074,18 +1075,23 @@ handle_image (void *data, unsigned int datasize,
 				       entry->vendor_version,
 				       entry->vendor_url);
 			}
+			res = parse_sbat_var(&sbat_var_entries);
+			if (res < 0)
+				console_print(L"could not parse SBAT variable");
```
martinezjavier, line 1080: shouldn't this be an error instead of just an info log? And the error propagated ? This is another reason to make this function return an `EFI_STATUS`.

pe.c
```diff
@@ -1074,18 +1075,23 @@ handle_image (void *data, unsigned int datasize,
...
 		} else {
 			perror(L"SBAT data not found\n");
 			return EFI_UNSUPPORTED;
 		}
-
```
martinezjavier, line 1081: Please drop this unrelated change.

pe.c
```diff
@@ -1074,18 +1075,23 @@ handle_image (void *data, unsigned int datasize,
...
 		efi_status = verify_buffer(data, datasize,
 					   &context, sha256hash, sha1hash);
-
+		if (sbat.entries)
```
martinezjavier, line 1087: We should check for this only one.

pe.c
```diff
@@ -1074,18 +1075,23 @@ handle_image (void *data, unsigned int datasize,
...
+			efi_status = verify_sbat(&sbat, &sbat_var_entries);
+			/* Add EFI_SECURITY_VIOLATION if variable is set,
+			and binary has no SBAT section ?*/
 		if (sbat.entries)
```
martinezjavier, line 1091: Drop this and make it all part of the same block if `sbat.entries` is set. This condition should always be true anyways.

pe.c
```diff
@@ -1074,18 +1075,23 @@ handle_image (void *data, unsigned int datasize,
...
 		} else {
 			perror(L"SBAT data not found\n");
 			return EFI_UNSUPPORTED;
 		}
-
 		efi_status = verify_buffer(data, datasize,
 					   &context, sha256hash, sha1hash);
-
+		if (sbat.entries)
+			efi_status = verify_sbat(&sbat, &sbat_var_entries);
```
martinezjavier, line 1088: you have to propagate the error here.

pe.c
```diff
@@ -1074,18 +1075,23 @@ handle_image (void *data, unsigned int datasize,
...
+			/* Add EFI_SECURITY_VIOLATION if variable is set,
+			and binary has no SBAT section ?*/
 		if (sbat.entries)
 			for (i = 0; i < sbat.size; i++)
 				FreePool(sbat.entries[i]);
-
+
```
martinezjavier, line 1094: Please drop all these unrelated changes.

sbat.c
```diff
@@ -4,27 +4,28 @@
  */
 
 #include "sbat.h"
+#include <string.h>
 
 CHAR8 *
-get_sbat_field(CHAR8 *current, CHAR8 *end, const CHAR8 ** field, char delim)
+get_sbat_field(CHAR8 *current, CHAR8 *end, const CHAR8 **field, char delim)
```
martinezjavier, line 10: I see what happened. You also squashed @vathpela's patch that was cleaning up the formatting issues. That should be left as a separate patch and I think that can already be merged.

sbat.c
```diff
@@ -120,4 +123,161 @@ EFI_STATUS parse_sbat(char *sbat_base, size_t sbat_size, char *buffer,
...
 }
 
+EFI_STATUS
+verify_sbat(struct sbat *sbat, list_t *sbat_entries)
+{
+	unsigned int i;
+	list_t *pos = NULL, *tmp = NULL;
+
+	for (i = 0; i < sbat->size; i++) {
+		struct sbat_entry *entry = entry = sbat->entries[i];
```
martinezjavier, line 133: Maybe add a `verify_sbat_entry()` function or something to make this function smaller and easier to read?

sbat.c
```diff
@@ -120,4 +123,161 @@ EFI_STATUS parse_sbat(char *sbat_base, size_t sbat_size, char *buffer,
...
+		for_each_sbat_var_safe (pos, tmp, sbat_entries) {
```
martinezjavier, line 134: The `_safe` variant is only needed when you will modify the list / remove elements, but just to iterate you can use the `for_each_sbat_var()` variant.

sbat.c
```diff
@@ -120,4 +123,161 @@ EFI_STATUS parse_sbat(char *sbat_base, size_t sbat_size, char *buffer,
...
+			if (strcmp(entry->component_name,
+			           sbat_var_entry->component_name) == 0) {
+				dprint(L"component %a has a matching SBAT variable entry, verifying\n", entry->component_name);
+				/* atoi returns zero for failed conversion, so essentially
+				   badly parsed component_generation will be treated as zero 
+				*/
+				UINT16 sbat_gen = atoi(entry->component_generation);
+				UINT16 sbat_var_gen = atoi(sbat_var_entry->component_generation);
+				if ( sbat_gen < sbat_var_gen ) {
+					dprint(L"component's %a generation: %d. Conflicts with SBAT variable generation %d\n",
```
martinezjavier, line 146: not sure I understand this debug message. What exactly conflicts means in this context ?

Maybe something like "component %a generation %d has been revoked by SBAT variable" or something like that?

sbat.c
```diff
@@ -120,4 +123,161 @@ EFI_STATUS parse_sbat(char *sbat_base, size_t sbat_size, char *buffer,
...
+{
+	unsigned char bom[] = { 0xEF, 0xBB, 0xBF };
+	return !CompareMem(buf, bom, MIN(sizeof(bom), bufsize));
+}
+
+static struct sbat_var *
+new_entry(const CHAR8 *comp_gen, const CHAR8 *comp_name_size,
+          const CHAR8 *comp_name)
+{
+	struct sbat_var *new_entry = AllocatePool(sizeof(*new_entry));
```
martinezjavier, line 169: You need to check for allocation failures.

sbat.c
```diff
@@ -120,4 +123,161 @@ EFI_STATUS parse_sbat(char *sbat_base, size_t sbat_size, char *buffer,
...
+
+	INIT_LIST_HEAD(&new_entry->list); 
+	new_entry->component_generation = comp_gen;
+	new_entry->component_name_size = comp_name_size;
+	new_entry->component_name = comp_name;
+	return new_entry;
+}
+
+static int
```
martinezjavier, line 178: Maybe returning `EFI_STATUS` here as well ?

sbat.c
```diff
@@ -120,4 +123,161 @@ EFI_STATUS parse_sbat(char *sbat_base, size_t sbat_size, char *buffer,
...
+add_entry(list_t *list, const CHAR8 *comp_gen, const CHAR8 *comp_name_size,
+          const CHAR8 *comp_name)
+{
+	struct sbat_var *new;
+	new = new_entry(comp_gen, comp_name_size, comp_name);
+	if (!new)
+		return -1;
```
martinezjavier, line 185: `return EFI_OUT_OF_RESOURCES` ?

sbat.c
```diff
@@ -120,4 +123,161 @@ EFI_STATUS parse_sbat(char *sbat_base, size_t sbat_size, char *buffer,
...
+}
+
+static void
+clean_up_vars(list_t *entries)
+{
+	list_t *pos = NULL, *tmp = NULL;
+	for_each_sbat_var_safe(pos, tmp, entries) {
+		struct sbat_var *entry;
+
+		entry = list_entry(pos, struct sbat_var, list);
```
martinezjavier, line 198: I think `struct sbat_var *entry = list_entry(pos, struct sbat_var, list);` is easier to read.

sbat.c
```diff
@@ -120,4 +123,161 @@ EFI_STATUS parse_sbat(char *sbat_base, size_t sbat_size, char *buffer,
...
+		list_del(&entry->list);
+		if (entry->component_generation)
+			FreePool((CHAR8 *)entry->component_generation);
```
martinezjavier, line 201: `FreePool()` takes `void *` as a parameter, I don't think there's a need to cast there.

sbat.c
```diff
@@ -120,4 +123,161 @@ EFI_STATUS parse_sbat(char *sbat_base, size_t sbat_size, char *buffer,
...
+{
+	UINT8 *data = 0;
+	UINTN datasize, i;
+	EFI_STATUS efi_status;
+	int rc;
+
+	INIT_LIST_HEAD(entries); 
+	efi_status = get_variable(L"SBAT", &data, &datasize, SHIM_LOCK_GUID);
+	if (EFI_ERROR(efi_status)) {
+		return -1;
```
martinezjavier, line 221: `return efi_status`

sbat.c
```diff
@@ -120,4 +123,161 @@ EFI_STATUS parse_sbat(char *sbat_base, size_t sbat_size, char *buffer,
...
+	dprint(L"SBAT variable data:\n");
+
+	rc = 0;
+	char delim;
+	while (start) {
+		const CHAR8 *fields[3] = { NULL, };
+		for (i = 0; i < 4; i++) {
+			const CHAR8 *tmp;
+			/* on fourth iteration we check if we had extra stuff on line while parsing last meaningful extraction of data
+			   if delimeter on 3rd iteration was ',', this means, we have comments after component_name, so let's drop them
```
martinezjavier, line 241: Please keep lines in less than 80 chars.

sbat.c
```diff
@@ -120,4 +123,161 @@ EFI_STATUS parse_sbat(char *sbat_base, size_t sbat_size, char *buffer,
...
+			*/
+			if (i == 3 && start[0] != '\0') {
+				if (delim == ',') {
+					start = get_sbat_field(start, end, &tmp, '\n');
```
martinezjavier, line 245: Please add a comment before this that mentions this is the optional comment field that's being parsed.

sbat.c
```diff
@@ -120,4 +123,161 @@ EFI_STATUS parse_sbat(char *sbat_base, size_t sbat_size, char *buffer,
...
+					break;
+				}
+				else
```
martinezjavier, line 248: Why do you need the else clause if you are breaking in both cases ?

martinezjavier: Also, if we are expecting 3 or 4 fields anyways I'm not sure that implementing it as a for loop is more readable instead of just trying to parse each field individually.

sbat.c
```diff
@@ -120,4 +123,161 @@ EFI_STATUS parse_sbat(char *sbat_base, size_t sbat_size, char *buffer,
...
+			}
+			delim = ',';
+			/* we do not want to jump to next line and grab stuff from that
+			*/
+			if ( (strchrnula(start,'\n') - start + 1) <= (strchrnula(start,',') - start + 1)) {
+				delim = '\n';
+			}
+			if (!start) {
+				rc = -1;
+				break;
```
martinezjavier, line 259: I think is confusing which break is for the for loop and which one is for the while loop, maybe better to add an `err` or `end` label before the `if (rc < 0)` check and a `goto` statement for the error paths ?

sbat.c
vathpela: Yeah, that's probably true - my bad for suggesting int instead of EFI_STATUS.

sbat.c
aburmash: added at 158,  or did you mean to wrap even more code into that verify_single_entry ?

sbat.c
```diff
@@ -120,4 +121,168 @@ EFI_STATUS parse_sbat(char *sbat_base, size_t sbat_size, char *buffer,
...
+verify_single_entry(struct sbat_entry *entry, struct sbat_var *sbat_var_entry)
+{
+        UINT16 sbat_gen, sbat_var_gen;
+
+        if (strcmp(entry->component_name, sbat_var_entry->component_name) == 0) {
+                dprint(L"component %a has a matching SBAT variable entry, verifying\n",
+                       entry->component_name);
+                /* atoi returns zero for failed conversion, so essentially
+                badly parsed component_generation will be treated as zero
+                */
```
martinezjavier, line 134: Right, I remember now reading a comment from @vathpela about this. I wonder if we should add a `strtoul()` to shim or use the one provided by Cryptlib ? The other option is to always start at 1 to make sure that we could discriminate between an error and a 0 component generation number.

@jsetje what do you think ?

sbat.c
martinezjavier: That's OK, I think makes the function easier to read.

sbat.c
```diff
@@ -120,4 +121,168 @@ EFI_STATUS parse_sbat(char *sbat_base, size_t sbat_size, char *buffer,
...
+
+EFI_STATUS
+verify_sbat(struct sbat *sbat, list_t *sbat_entries)
+{
+	unsigned int i;
+	list_t *pos = NULL;
+	EFI_STATUS efi_status;
+	
+	for (i = 0; i < sbat->size; i++) {
+		struct sbat_entry *entry = entry = sbat->entries[i];
```
martinezjavier, line 155: why `struct sbat_entry *entry = entry = sbat->entries[i];` and not just `struct sbat_entry *entry =  sbat->entries[i];` ? 

Probably makes more sense to define the variable outside of the for loop though.

sbat.c
```diff
@@ -120,4 +121,168 @@ EFI_STATUS parse_sbat(char *sbat_base, size_t sbat_size, char *buffer,
...
+		for_each_sbat_var(pos, sbat_entries) {
+			struct sbat_var *sbat_var_entry = list_entry(pos, struct sbat_var, list);
```
martinezjavier, line 157: Same, let's define these at the beginning of the `verify_sbat()` function.

sbat.c
```diff
@@ -120,4 +121,168 @@ EFI_STATUS parse_sbat(char *sbat_base, size_t sbat_size, char *buffer,
...
+	dprint(L"all entries from SBAT section verified\n");
+	
+	return efi_status;
+}
+
+static BOOLEAN
+is_utf8_bom(CHAR8 *buf, size_t bufsize)
+{
+	unsigned char bom[] = { 0xEF, 0xBB, 0xBF };
+	return !CompareMem(buf, bom, MIN(sizeof(bom), bufsize));
```
martinezjavier, line 172: I think @vathpela wanted this to be `return CompareMem(buf, bom, MIN(sizeof(bom), bufsize)) != 0; ?

sbat.c
martinezjavier: You missed this comment.

sbat.c
martinezjavier: This one as well, there's no need to cast when calling `FreePool()`.

sbat.c
```diff
@@ -120,4 +121,168 @@ EFI_STATUS parse_sbat(char *sbat_base, size_t sbat_size, char *buffer,
...
+EFI_STATUS
+parse_sbat_var(list_t *entries)
+{
+	UINT8 *data = 0;
+	UINTN datasize, i;
+	EFI_STATUS efi_status;
+
+	INIT_LIST_HEAD(entries); 
+	efi_status = get_variable(L"SBAT", &data, &datasize, SHIM_LOCK_GUID);
+	if (EFI_ERROR(efi_status)) {
```
martinezjavier, line 227: `if (EFI_ERROR(efi_status) && efi_status != EFI_NOT_FOUND) {`

martinezjavier: for `EFI_NOT_FOUND` you should just print a message and return `EFI_SUCCESS` I believe, since as you mentioned there may not be a variable at the beginning.

sbat.c
```diff
@@ -120,4 +121,168 @@ EFI_STATUS parse_sbat(char *sbat_base, size_t sbat_size, char *buffer,
...
+		}
+		dprint(L"component %a with generation %a\n", fields[2], fields[0]);
+		efi_status = add_entry(entries, fields[0], fields[1], fields[2]);
+		if (EFI_ERROR(efi_status))
+			goto error;
+	}
+	FreePool(data);
+	return EFI_SUCCESS;
+error:
+	dprint(L"failed to parse SBAT variable\n");
```
martinezjavier, line 283: Since this is an error condition probably we don't want a debug message and instead a `perror()`.

include/str.h
```diff
@@ -54,7 +54,7 @@ strndupa(const CHAR8 * const src, const UINTN srcmax)
 		return news;
 
 	len = strnlena(src, srcmax);
-	news = AllocateZeroPool(len);
+	news = AllocateZeroPool(len+1);
```
martinezjavier, line 57: `len + 1`. Also this should be a separate patch if is a fix for `strndupa()`.

pe.c
```diff
@@ -1078,13 +1080,24 @@ handle_image (void *data, unsigned int datasize,
...
+		
+		efi_status = parse_sbat_var(&sbat_var_entries);
+		/* we get EFI_SUCCESS only if variable is fully parsed,
+		   and EFI_NOT FOUND means do not perform verificaiton
+		   continue with EFI_SUCCESS
+		*/
+		if (efi_status == EFI_SUCCESS)
+			efi_status = verify_sbat(&sbat, &sbat_var_entries);
+		else if (efi_status == EFI_NOT_FOUND)
+			efi_status = EFI_SUCCESS;
```
martinezjavier, line 1092: Don't you need to return an error if the `efi_status` isn't `EFI_SUCCESS` or `EFI_NOT_FOUND` ?

pe.c
```diff
@@ -1078,13 +1080,24 @@ handle_image (void *data, unsigned int datasize,
...
+		
 		efi_status = verify_buffer(data, datasize,
 					   &context, sha256hash, sha1hash);
 
-		if (sbat.entries)
+		if (sbat.entries) {
 			for (i = 0; i < sbat.size; i++)
 				FreePool(sbat.entries[i]);
+		}
```
martinezjavier, line 1100: You don't need to add these braces. Although probably makes sense to avoid mistakes later.

Issue #177, opened 2019-05-08
MokManager screen not being drawn properly


Randune:
When running mmx64.efi from the EFI shell, the introductory screen for MokManager doesn't draw entirely, the only thing drawn on the screen is the text which says "Press any key to perform Mok Management" and the countdown timer, the rest of the screen is not drawn (I can see the EFI shell in behind) and when I press any key to continue, the counter stops but the next screen isn't drawn, I know the program is running because when I press Esc it quits back to the EFI shell.  *See the attachment*
![mok_manager](https://user-images.githubusercontent.com/50386677/57341386-3e469180-7108-11e9-9fbe-77223bed6822.jpg)

gibmat:
I have run into this same issue as well while trying to boot into MokManager under a Debian 10 (buster) install; shim-unsigned package version is 15+1533136590.3beb971-7.

There are a few other reports of similar issues online:
* https://askubuntu.com/questions/950395/mok-management-will-not-load-on-boot
* https://bugzilla.redhat.com/show_bug.cgi?id=1520709

The common thread seems to be encountering this issue on ASRock motherboards. The machine I've been working on has a Z270M Pro4 motherboard.

I'd be happy to help test/debug potential solutions to this issue, although I don't have the necessary insight to suggest potential fixes.

Randune:
I opened a ticket with ASRock about this issue, I told them that it is a common problem among ASRock motherboards, and they simply said "we don't support any linux version".

I proceeded to tell them that this isn't a linux issue, as MokManager is an EFI application, not a linux application, but I never received a response after that.  I wish we could get some weight behind this issue and pressure them to do something about it.

gibmat:
Yup, it's disappointing when manufacturers believe Windows is the only thing to ever run on their hardware.

Randune:
ok, I've opened another ticket with ASRock, hopefully we can gain some traction regarding this issue.

jeremyvisser:
I have an ASRock X470 Taichi which also suffers from this exact issue.

tommyvct:
Still nothing?

jjohnson42:
> I got an ASRock X399 Taichi and have encountered the same problem. This MB allows you to clear the keys and installing default keys from the setup, but not importing custom keys. I tried contacting ASRock support but all I got in reply was “_… supports Secure Boot for Win 10 OS only_”.
> 
[...]

PR #271, opened 2021-02-15
Replace Unicode Box Drawing characters with ASCII characters

tonper:
Solves #177. Many ASRock boards will not render MokManager correctly if the Unicode Box Drawing characters are used. This PR replace the `BOXDRAW_` characters with ASCII characters.

I'm not sure how many systems are affected by this problem, so I can't say if ugly ASCII boxes are worth it. A nicer solution would be to detect the issue and use ASCII as a fallback, but I'm not sure if that's possible.

Tested on an ASRock X399 Taichi.


vathpela:
Is it possible to figure out that this is an ASRock board and conditionalize on that?  I'm really hesitant to make this look worse for everyone else because of one vendor of super-low-quality motherboards.

mjg59:
One somewhat horrifying approach - draw a box character, read back the framebuffer, see whether it looks like a box character?

tonper:
Turns out you can use the return value from `co->OutputString()` to detect the problem. I've modified the branch to use ASCII as a fallback instead.

xnox:
@vathpela can this be undrafted? or is this somehow not appropriate to go in?

lib/console.c
```diff
@@ -180,10 +205,16 @@ console_print_box_at(CHAR16 *str_arr[], int highlight,
 		return;
 	}
 
-	SetMem16 (Line, size_cols * 2, BOXDRAW_HORIZONTAL);
+	/* test if boxdraw characters work */
+	co->SetCursorPosition(co, start_col, start_row);
+	Line[0] = boxdraw[0].up_left;
+	Line[1] = L'\0';
+	char_set = co->OutputString(co, Line) == 0 ? 0 : 1;
```
xnox, line 212: UEFI Specification Version 2.8 Errata B does specify the BOXDRAW characters as mandatory. As per the Spec's page 448, the Status Codes that can be returned here are EFI_SUCCESS, EFI_DEVICE_ERROR, EFI_UNSUPPORTED, and EFI_WARN_UNKNOWN_GLYPH.

I would prefer to test for EFI_WARN_UNKNONW_GLYPH (and if that's in fact what is returned on ASRock) and then switch to the other character set based on that.

Also it would be nice to retest that the one one works and doesn't also return EFI_WARN_UNKNOWN_GLYPH in case the firware just returns that always irrespective if it was successful or not.

PR #274, opened 2021-02-16
 Parse SBAT variable and perform basic verification

aburmash:
Some minor fixes and additions + SBAT variable parsing and verification.

### DIFF
--- a/pe.c
+++ b/pe.c
@@ -823,7 +823,7 @@ read_header(void *data, unsigned int datasize,
 	return EFI_SUCCESS;
 }
 
-static EFI_STATUS
+EFI_STATUS
 handle_sbat(char *SBATBase, size_t SBATSize)
 {
 	unsigned int i;
--- a/pe.c
+++ b/pe.c
@@ -1044,6 +1044,8 @@ handle_image (void *data, unsigned int datasize,
 		size_t n;
 		struct sbat_entry **entries;
 		struct sbat_entry *entry = NULL;
+		list_t sbat_var_entries;
+		INIT_LIST_HEAD(&sbat_var_entries);
 
 		if (SBATBase && SBATSize) {
 			char *sbat_data;
@@ -1079,6 +1081,32 @@ handle_image (void *data, unsigned int datasize,
 		} else {
 			perror(L"SBAT data not found\n");
 			return EFI_UNSUPPORTED;
+		}
+
+		efi_status = parse_sbat_var(&sbat_var_entries);
+		/*
+		 * Until a SBAT variable is installed into the systems, it is expected that
+		 * attempting to parse the variable will fail with an EFI_NOT_FOUND error.
+		 *
+		 * Do not consider that error fatal for now.
+		 */
+		if (EFI_ERROR(efi_status) && efi_status != EFI_NOT_FOUND) {
+			perror(L"Parsing SBAT variable failed: %r\n",
+			       efi_status);
+			return efi_status;
+		}
+
+		if (efi_status == EFI_SUCCESS)
+			efi_status = verify_sbat(n, entries, &sbat_var_entries);
+		if (efi_status == EFI_NOT_FOUND)
+			efi_status = EFI_SUCCESS;
+
+		if (EFI_ERROR(efi_status)) {
+			if (verbose)
+				console_print(L"Verification failed: %r\n", efi_status);
+			else
+				console_error(L"Verification failed", efi_status);
+			return efi_status;
 		}
 
 		efi_status = verify_buffer(data, datasize,
--- a/pe.c
+++ b/pe.c
@@ -1041,7 +1041,8 @@ handle_image (void *data, unsigned int datasize,
 	if (secure_mode ()) {
 		unsigned int i;
 		EFI_STATUS efi_status;
-		struct sbat sbat = { 0 };
+		size_t n;
+		struct sbat_entry **entries;
 		struct sbat_entry *entry = NULL;
 
 		if (SBATBase && SBATSize) {
@@ -1057,7 +1058,7 @@ handle_image (void *data, unsigned int datasize,
 			CopyMem(sbat_data, SBATBase, SBATSize);
 			sbat_data[SBATSize] = '\0';
 
-			efi_status = parse_sbat(sbat_data, sbat_size, &sbat);
+			efi_status = parse_sbat(sbat_data, sbat_size, &n, &entries);
 			if (EFI_ERROR(efi_status)) {
 				perror(L"SBAT data not correct: %r\n",
 				       efi_status);
@@ -1065,8 +1066,8 @@ handle_image (void *data, unsigned int datasize,
 			}
 
 			dprint(L"SBAT data\n");
-			for (i = 0; i < sbat.size; i++) {
-				entry = sbat.entries[i];
+			for (i = 0; i < n; i++) {
+				entry = entries[i];
 				dprint(L"%a, %a, %a, %a, %a, %a\n",
 				       entry->component_name,
 				       entry->component_generation,
@@ -1083,9 +1084,9 @@ handle_image (void *data, unsigned int datasize,
 		efi_status = verify_buffer(data, datasize,
 					   &context, sha256hash, sha1hash);
 
-		if (sbat.entries)
-			for (i = 0; i < sbat.size; i++)
-				FreePool(sbat.entries[i]);
+		if (entries)
+			for (i = 0; i < n; i++)
+				FreePool(entries[i]);
 
 		if (EFI_ERROR(efi_status)) {
 			if (verbose)
--- a/pe.c
+++ b/pe.c
@@ -823,6 +823,60 @@ read_header(void *data, unsigned int datasize,
 	return EFI_SUCCESS;
 }
 
+static EFI_STATUS
+handle_sbat(char *SBATBase, size_t SBATSize)
+{
+	unsigned int i;
+	EFI_STATUS efi_status;
+	size_t n;
+	struct sbat_entry **entries = NULL;
+	char *sbat_data;
+	size_t sbat_size;
+
+	if (list_empty(&sbat_var))
+		return EFI_SUCCESS;
+
+	if (SBATBase == NULL || SBATSize == 0) {
+		dprint(L"No .sbat section data\n");
+		return EFI_SECURITY_VIOLATION;
+	}
+
+	sbat_size = SBATSize + 1;
+	sbat_data = AllocatePool(sbat_size);
+	if (!sbat_data) {
+		console_print(L"Failed to allocate .sbat section buffer\n");
+		return EFI_OUT_OF_RESOURCES;
+	}
+	CopyMem(sbat_data, SBATBase, SBATSize);
+	sbat_data[SBATSize] = '\0';
+
+	efi_status = parse_sbat(sbat_data, sbat_size, &n, &entries);
+	if (EFI_ERROR(efi_status)) {
+		perror(L"Could not parse .sbat section data: %r\n", efi_status);
+		goto err;
+	}
+
+	dprint(L"SBAT section data\n");
+        for (i = 0; i < n; i++) {
+		dprint(L"%a, %a, %a, %a, %a, %a\n",
+		       entries[i]->component_name,
+		       entries[i]->component_generation,
+		       entries[i]->vendor_name,
+		       entries[i]->vendor_package_name,
+		       entries[i]->vendor_version,
+		       entries[i]->vendor_url);
+	}
+
+	efi_status = verify_sbat(n, entries);
+
+	cleanup_sbat_entries(n, entries);
+
+err:
+	FreePool(sbat_data);
+
+	return efi_status;
+}
+
 /*
  * Once the image has been loaded it needs to be validated and relocated
  */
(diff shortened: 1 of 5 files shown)


pe.c
```diff
@@ -1079,6 +1081,16 @@ handle_image (void *data, unsigned int datasize,
 			return EFI_UNSUPPORTED;
 		}
 
+		efi_status = parse_sbat_var(&sbat_var_entries);
+		/* we get EFI_SUCCESS only if variable is fully parsed,
+		   and EFI_NOT FOUND means do not perform verificaiton
+		   continue with EFI_SUCCESS
```
martinezjavier, line 1087: There's two typos here: verification and EFI_NOT_FOUND.

But I would maybe change this comment to something like:

```
/*
 * Until a SBAT variable is installed into the systems, it is expected that
 * attempting to parse the variable will fail with a EFI_NOT_FOUND error.
 *
 * Do not consider that error fatal for now.
 */
```

pe.c
```diff
@@ -1079,6 +1081,16 @@ handle_image (void *data, unsigned int datasize,
...
+		*/
+		if (efi_status == EFI_SUCCESS)
+			efi_status = verify_sbat(&sbat, &sbat_var_entries);
+		else if (efi_status == EFI_NOT_FOUND)
+			efi_status = EFI_SUCCESS;
```
martinezjavier, line 1092: you are missing a else clause here to return errors `if (efi_status != EFI_SUCCESS && efi_status != EFI_NOT_FOUND)`.

martinezjavier: Actually, probably makes sense to just do `if (EFI_ERROR(efi_status))` here, to catch both `parse_sbat_var()` and `verify_sbat()` returned errors.

sbat.c
```diff
@@ -120,4 +121,175 @@ EFI_STATUS parse_sbat(char *sbat_base, size_t sbat_size, char *buffer,
...
+verify_single_entry(struct sbat_entry *entry, struct sbat_var *sbat_var_entry)
+{
+        UINT16 sbat_gen, sbat_var_gen;
+
+        if (strcmp(entry->component_name, sbat_var_entry->component_name) == 0) {
+                dprint(L"component %a has a matching SBAT variable entry, verifying\n",
+                       entry->component_name);
+                /* atoi returns zero for failed conversion, so essentially
+                badly parsed component_generation will be treated as zero
+                */
```
martinezjavier, line 134: @jsetje @vathpela I think we need to add a `strtoul()` implementation. Or do you think that we should change the initial generation number to 1, to just use `atoi()` ?

sbat.c
```diff
@@ -120,4 +121,175 @@ EFI_STATUS parse_sbat(char *sbat_base, size_t sbat_size, char *buffer,
...
+}
+
+EFI_STATUS
+verify_sbat(struct sbat *sbat, list_t *sbat_entries)
+{
+	unsigned int i;
+	list_t *pos = NULL;
+	EFI_STATUS efi_status = EFI_SUCCESS;
+	struct sbat_entry *entry;
+	struct sbat_var *sbat_var_entry;
```
martinezjavier, line 171: I would add a blank line here to separate the variables declaration from the logic in the function.

sbat.c
```diff
@@ -120,4 +121,175 @@ EFI_STATUS parse_sbat(char *sbat_base, size_t sbat_size, char *buffer,
...
+	if (sbat_empty(sbat_entries)) {
+		dprint(L"SBAT variable not present or malformed\n");
+		return EFI_INVALID_PARAMETER;
+	}
```
martinezjavier, line 175: same here, please add a new line.

sbat.c
```diff
@@ -120,4 +121,175 @@ EFI_STATUS parse_sbat(char *sbat_base, size_t sbat_size, char *buffer,
...
+	for (i = 0; i < sbat->size; i++) {
+		entry = sbat->entries[i];
+		for_each_sbat_var(pos, sbat_entries) {
+			sbat_var_entry = list_entry(pos, struct sbat_var, list);
+			efi_status = verify_single_entry(entry, sbat_var_entry);
+			if (EFI_ERROR(efi_status))
+				return efi_status;
+		}
+	}
```
martinezjavier, line 184: and here.

sbat.c
```diff
@@ -120,4 +121,175 @@ EFI_STATUS parse_sbat(char *sbat_base, size_t sbat_size, char *buffer,
...
+			return efi_status;
+	}
+
+	CHAR8 *start = (CHAR8 *)data;
+	CHAR8 *end = (CHAR8 *)data + datasize;
+	if (is_utf8_bom(start, datasize))
+		start += 3;
+	while ((*end == '\r' || *end == '\n') && end < start)
+		end--;
+	*end = '\0';
```
martinezjavier, line 246: @christopherco should we drop this while loop as well ?

pe.c
aburmash: I am not actually sure i understand you proposal ( sleep is important stuff )
Currently we do verify_sbat only in case parse_sbat_var returned EFI_SUCCESS.

If parse_sbat_var returned anything else, except for EFI_NOT_FOUND and EFI_SUCCESS this whole loop will not be executed, and error will be propagated to a later stages of code.

verify_sbat can also return only:
EFI_INVALID_PARAMETER
EFI_SECURITY_VIOLATION
EFI_SUCCESS

so we will only be getting EFI_NOT_FOUND only in case parse_sbat_var returned it.

The check for (EFI_ERROR(efi_status) should be at the place, where we halt the failed verification.
Currently even after failed sbat we still do verify_buffer, and binary can be validated by it: whenever we want to fail boot immediately after SBAT failure is what i do not know.

martinezjavier: > f parse_sbat_var returned anything else, except for EFI_NOT_FOUND and EFI_SUCCESS this whole loop will not be executed, and error will be propagated to a later stages of code.

That's not true. If `parse_sbat_var()` returned anything that's not EFI_NOT_FOUND || EFI_SUCCESS, then this error value will not be propagated and `verify_buffer()` wrongly called.

martinezjavier: My proposal is to do something like the following:

```
               efi_status = parse_sbat_var(&sbat_var_entries);
               /*
                * Until a SBAT variable is installed into the systems, it is expected that
                * attempting to parse the variable will fail with a EFI_NOT_FOUND error.
                *
                * Do not consider that error fatal for now.
                */
               if (efi_status == EFI_SUCCESS)
                       efi_status = verify_sbat(&sbat, &sbat_var_entries);
               else if (efi_status == EFI_NOT_FOUND)
                       efi_status = EFI_SUCCESS;

               if (EFI_ERROR(efi_status)) {
                 // print error log and do any cleanup needed
                 return efi_status;
               }

               efi_status = verify_buffer(data, datasize,
                                           &context, sha256hash, sha1hash);
```

martinezjavier: > Currently even after failed sbat we still do verify_buffer, and binary can be validated by it: whenever we want to fail boot immediately after SBAT failure is what i do not know.

If we fail to either parse the SBAT variable (i.e: any other error than EFI_NOT_FOUND) or the `.sbat` section, then I think that we can't boot the binary. Otherwise we risk to boot a binary that has been already revoked by SBAT.

So in that case `verify_buffer()` should be cal it and `handle_image()` return earlier IMO.

aburmash: OK, we are talking about same thing. I was thinking about binaries, that might be enrolled to allowlist by hash. We still would fail them. If yes, this pretty much what i was trying to understand.

sbat.c
```diff
@@ -120,4 +121,191 @@ EFI_STATUS parse_sbat(char *sbat_base, size_t sbat_size, char *buffer,
...
+EFI_STATUS
+verify_sbat(struct sbat *sbat, list_t *sbat_entries)
+{
+	unsigned int i;
+	list_t *pos = NULL;
+	EFI_STATUS efi_status = EFI_SUCCESS;
+	struct sbat_entry *entry;
+	struct sbat_var *sbat_var_entry;
+
+	if (sbat_empty(sbat_entries) || !sbat_entries) {
```
aburmash, line 179: @martinezjavier could you looks at this ? Specifically at || !sbat_entries.
This should not happen, but better check ?

martinezjavier: it depends if you assume the caller will never pass a NULL pointer, agreed that better to check.

But the check is backwards, it has to be `if (!sbat_entries || sbat_empty(sbat_entries)`, otherwise passing a NULL to `sbat_empty()` will cause a NULL pointer dereference when doing `head->next`.

sbat.c
christopherco: Agreed, we should drop the loop here too.

sbat.c
```diff
@@ -117,4 +118,199 @@ parse_sbat(char *sbat_base, size_t sbat_size, struct sbat *sbat)
...
+
+	efi_status = get_variable(L"SBAT", &data, &datasize, SHIM_LOCK_GUID);
+	if (EFI_ERROR(efi_status)) {
+		LogError(L"Failed to read SBAT variable\n",
+				      efi_status);
+		return efi_status;
+	}
+
+	CHAR8 *start = (CHAR8 *)data;
+	CHAR8 *end = (CHAR8 *)data + datasize;
```
christopherco, line 257:
```suggestion
	CHAR8 *end = (CHAR8 *)data + datasize - 1;
```
Currently looks like we are indexing past the end of the data buffer by 1.

aburmash: should be fixed by now

sbat.c
aburmash: should be fixed

aburmash: should be fixed

sbat.c
aburmash: should be fixed

sbat.c
```diff
@@ -117,4 +118,191 @@ parse_sbat(char *sbat_base, size_t sbat_size, struct sbat *sbat)
...
+parse_sbat_var(list_t *entries)
+{
+	UINT8 *data = 0;
+	UINTN datasize, i;
+	EFI_STATUS efi_status;
+	INIT_LIST_HEAD(entries);
+	char delim;
+
+	if (!entries)
+		return EFI_INVALID_PARAMETER;
```
christopherco, line 242:
```suggestion
	char delim;

	if (!entries)
		return EFI_INVALID_PARAMETER;

	INIT_LIST_HEAD(entries);
```
`INIT_LIST_HEAD(entries)` should be after the `!entries` sanity check.

sbat.c
```diff
@@ -117,4 +118,191 @@ parse_sbat(char *sbat_base, size_t sbat_size, struct sbat *sbat)
...
+			 * component_name. If delimeter on 2nd iteration was ',', this means that
+			 * we have comments after component_name. get_sbat_field in this if condition
+			 * parses comments, if they are present and drops them.
+			 */
+			if (i == 2 && start) {
+				if (delim == ',') {
+					start = get_sbat_field(start, end, &tmp,
+					                       '\n');
+				}
+				break;
```
chrisccoulson, line 275: Is the sbat variable payload always terminated with a newline? If the last line in the payload has a comment and isn't terminated with a newline, then I think there is a NULL deref checking the condition of the outer loop after breaking from the inner loop here.

sbat.c
```diff
@@ -117,4 +118,191 @@ parse_sbat(char *sbat_base, size_t sbat_size, struct sbat *sbat)
...
+			}
+			delim = ',';
+			/* we do not want to jump to next line and grab stuff from that
+			 */
+			if ((strchrnula(start, '\n') - start + 1) <=
+			    (strchrnula(start, ',') - start + 1)) {
```
chrisccoulson, line 281: If this condition is true on the first iteration of the inner loop, should the payload be considered malformed?

pe.c
```diff
@@ -1080,6 +1082,22 @@ handle_image (void *data, unsigned int datasize,
...
+		if (EFI_ERROR(efi_status) && efi_status != EFI_NOT_FOUND) {
+			perror(L"Parsing SBAT variable failed: %r\n",
+			       efi_status);
+			return efi_status;
+		}
+
+		if (efi_status == EFI_SUCCESS)
+			efi_status = verify_sbat(&sbat, &sbat_var_entries);
+
 		efi_status = verify_buffer(data, datasize,
```
chrisccoulson, line 1101: Isn't the status returned from verify_sbat() being ignored here? Shouldn't verify_buffer() be skipped if verify_sbat() fails?

sbat.c
```diff
@@ -117,4 +118,191 @@ parse_sbat(char *sbat_base, size_t sbat_size, struct sbat *sbat)
...
+		}
+		dprint(L"component %a with generation %a\n", fields[0], fields[1]);
+		efi_status =
+			add_entry(entries, fields[0], fields[1]);
+		if (EFI_ERROR(efi_status))
+			goto error;
+	}
+	FreePool(data);
+	return EFI_SUCCESS;
+error:
```
chrisccoulson, line 303: data isn't freed on this path.

chrisccoulson (commented):
I've left a few comments on this

sbat.c
aburmash: Yes, per specification

aburmash: https://github.com/rhboot/shim/pull/276/commits/411117a107435506677d5ea4608c9c6af5155827#diff-ffd97917c03f0f6186701da25c73b3bc7107f48cc24182b75b6bae1f1f58c45eR103

pe.c
aburmash: Actually we did had a verification here, this one: https://github.com/martinezjavier/shim/commit/06ee1d9f8b06150a6ea4c7893325fdfbf8b29182#diff-7de403009b80237157eccb017ada7717e5d12c3b24afe8783e7370363191c073R1873 but Javier removed it during his rebase, i guess he had something else in mind. Anyway, i am keeping an eye on this, since most likely we need to readd it.

martinezjavier: Right, that was a mistake during my rebase when resolving the conflicts. I meant to keep the check for `verify_sbat()` there.

sbat.c
aburmash: should be fixed.

sbat.c
aburmash: should be fixed

pe.c
aburmash: should be fixed, though please double check current code. Right now, we will fail validation if there is no SBAT header in binary, even if binary is enrolled by hash.

sbat.c
aburmash: should be fixed

martinezjavier (approved):

jsetje (approved):

sbat.c
christopherco: Was this intentionally reverted?

aburmash: Yes, to make sure \n is always there

christopherco (approved):

jsetje (approved):
I've worked through the logic and am happy with it.

I do agree with the parse/create() verify() cleanup() refactor suggesting, but wouldn't gate on it.

I'd vote for pulling this patch in now, that way more folks can test with it.

martinezjavier (approved):
I believe this is ready to be merged now. Thanks a lot @aburmash for all the respins with this PR!

vathpela:
I've pushed this to main, along with https://github.com/vathpela/mallory/commit/9bef30f52951572a37cd83057aea676721c52380

PR #275, opened 2021-02-16
Allow shim to check it's own .sbat section

martinezjavier:
Shim should check if it has been revoked and avoid being executed if that's the case.

This pull-request depends on #274.

### DIFF
--- a/pe.c
+++ b/pe.c
@@ -823,7 +823,7 @@ read_header(void *data, unsigned int datasize,
 	return EFI_SUCCESS;
 }
 
-static EFI_STATUS
+EFI_STATUS
 handle_sbat(char *SBATBase, size_t SBATSize)
 {
 	unsigned int i;
--- a/pe.c
+++ b/pe.c
@@ -1044,6 +1044,8 @@ handle_image (void *data, unsigned int datasize,
 		size_t n;
 		struct sbat_entry **entries;
 		struct sbat_entry *entry = NULL;
+		list_t sbat_var_entries;
+		INIT_LIST_HEAD(&sbat_var_entries);
 
 		if (SBATBase && SBATSize) {
 			char *sbat_data;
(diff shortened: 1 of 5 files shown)


martinezjavier:
@chrisccoulson @aburmash  could you please take a look to the last two commits of this pull-request ?

shim.c
```diff
@@ -1831,6 +1831,59 @@ debug_hook(void)
...
+	if (efi_status == EFI_SUCCESS)
+		efi_status = verify_sbat(&sbat, &sbat_var_entries);
+	if (efi_status == EFI_NOT_FOUND)
+		efi_status = EFI_SUCCESS;
+
+out:
+	if (sbat.entries)
+		for (i = 0; i < sbat.size; i++)
+			FreePool(sbat.entries[i]);
+	FreePool(sbat_data);
```
chrisccoulson, line 1883: It doesn't look like sbat.entries is freed (allocated here https://github.com/rhboot/shim/blob/06ee1d9f8b06150a6ea4c7893325fdfbf8b29182/sbat.c#L99).

There is similar code (not part of this PR) in handle_image() that has the same issue (and that implementation also leaks sbat_data).

shim.c
```diff
@@ -1831,6 +1831,59 @@ debug_hook(void)
...
+	char *sbat_data = AllocatePool(sbat_size + 1);
+	if (!sbat_data) {
+		console_print(L"Failed to allocate SBAT buffer\n");
+		return EFI_OUT_OF_RESOURCES;
+	}
+
+	CopyMem(sbat_data, sbat_start, sbat_size);
+	sbat_data[sbat_size] = '\0';
+
+	efi_status = parse_sbat(sbat_data, sbat_size + 1, &sbat);
```
chrisccoulson, line 1855: Just an observation - I notice that both callers to parse_sbat() pass in size+1 to include the NULL terminator, but the first thing that parse_sbat() does is subtract 1 from the size.

shim.c
martinezjavier: You are correct about sbat.entries leaking.

I'll also include those memory leak fixes in handle_image() as a part of this series.

shim.c
martinezjavier: Right, I've no idea why parse_sbat() is substracting that 1 from the size...

shim.c
martinezjavier: @chrisccoulson fixed the leaks in this patch and added preparatory fixes in the pull-request for the other issues you pointed out.

shim.c
martinezjavier: @chrisccoulson deleted the subtraction since the `parse_sbat()` function should use the size passed by the caller, which  generally would be a null-terminated string.

shim.c
```diff
@@ -1831,6 +1831,73 @@ debug_hook(void)
...
+	if (EFI_ERROR(efi_status))
+		perror(L"shim SBAT verification failed: %r\n", efi_status);
+out:
+	if (sbat.entries) {
+		for (i = 0; i < sbat.size; i++)
+			FreePool(sbat.entries[i]);
+		FreePool(sbat.entries);
+	}
+	FreePool(sbat_data);
+	cleanup_sbat_var(&sbat_var_entries);
```
aburmash, line 1896: Maybe we could just add cleanup to verify_sbat failures, and you won't need it at all. Though up to you to decide, less stuff for me to change ) I believe we have:
1) cleanup on parse_sbat failure ( internal )
2) cleanup on verify_sbat success

if we add cleanup on verify_sbat failure we should not need it outside, do we ? or we have a condition when we need cleanup in-between parse and verify ?

pe.c
```diff
@@ -1061,12 +1061,12 @@ handle_image (void *data, unsigned int datasize,
...
 			efi_status = parse_sbat(sbat_data, sbat_size, &sbat);
 			if (EFI_ERROR(efi_status)) {
-				perror(L"SBAT data not correct: %r\n",
+				perror(L"second stage SBAT data not correct: %r\n",
 				       efi_status);
 				return efi_status;
 			}
 
-			dprint(L"SBAT data\n");
+			dprint(L"second stage SBAT data\n");
```
aburmash, line 1069: second stage SBAT data**:** ?

aburmash (approved):
Taking in account what we have discussed

martinezjavier:
Closing this since the patches are included now in PR #274

PR #279, opened 2021-02-18
Implement variable mirroring and TPM logging for SBAT 

vathpela:


mok.c
```diff
@@ -78,29 +78,66 @@ typedef vendor_addend_category_t (vendor_addend_categorizer_t)(struct mok_state_
...
+	 * The existing categorizers are for db and dbx; they differ
+	 * because we don't currently support a CERT for dbx.
+	 */
 	UINT8 **addend;
 	UINT32 *addend_size;
 
+	/*
+	 * build_cert is our build-time cert.  Like addend, this is added
+	 * to the input variable, as part of the runtime variable, so that
+	 * they're visible to the kernel.  This is the ephimeral cert used
```
martinezjavier, line 118: ephemeral.

martinezjavier (approved):

include/tpm.h
```diff
@@ -176,6 +175,7 @@ typedef struct efi_tpm2_protocol efi_tpm2_protocol_t;
 
 typedef UINT32                     TCG_EVENTTYPE;
 
+#define EV_IPL                              0x0000000d
```
out0xb2, line 178: **TCG PC Client Platform Firmware Profile, Family “2.0”, Level 00 Revision 1.04, June 3, 2019** says...

"This event is deprecated."

out0xb2: I should say, I like this change to define a magic number... just observing that we're adding a measurement using a deprecated type (should we stick with this for consistency, or choose a different type)?

tpm.c
```diff
@@ -209,7 +209,7 @@ EFI_STATUS tpm_log_event(EFI_PHYSICAL_ADDRESS buf, UINTN size, UINT8 pcr,
 			 const CHAR8 *description)
 {
 	return tpm_log_event_raw(buf, size, pcr, description,
-				 strlen(description) + 1, 0xd, NULL);
+				 strlen(description) + 1, EV_IPL, NULL);
```
out0xb2, line 212: If I had to do it over again, I wonder if I would choose EV_EFI_VARIABLE_DRIVER_CONFIG

include/tpm.h
vathpela: It's actually just /moving/ that definition... and actually using it where we're already using that value.  

That said, I think the point @mjg59 made elsewhere about not changing the event log formats we're already using remains valid here, even though this is a new item being logged.  I think we should stick with this for now: I don't want to come up with a plan on how to transition to different event log entry types ad-hoc in the middle of doing something else.

tpm.c
vathpela: That does seem better in some ways, but I don't see any good way to switch event types without breaking any code that's actually relying on the log.

mjg59 (approved):

PR #284, opened 2021-02-19
build: Import gnu-efi as a submodule and build against it

mjg59:
Shim is rather more friendly with EFI internals than most code, and as a
result can end up making assumptions that are out of step with those made
by gnu-efi. Since both projects are developed independently, and since
distributions are often trying to build versions of shim against whatever
version of gnu-efi they are shipping, this can result in awkward build
failures. The easiest way to handle this is to use a git submodule and
import a known-good version of shim directly into the build tree. Given
static linking, this will also make reproducible builds easier.

Signed-off-by: Matthew Garrett <mjg59@google.com>

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "gnu-efi"]
+	path = gnu-efi
+	url = https://git.code.sf.net/p/gnu-efi/code
--- a/Make.defaults
+++ b/Make.defaults
@@ -31,7 +31,7 @@ OPTIMIZATIONS	?= -Os
 
 SUBDIRS		= $(TOPDIR)/Cryptlib $(TOPDIR)/lib
 
-EFI_INCLUDE	?= /usr/include/efi
+EFI_INCLUDE	?= $(TOPDIR)/gnu-efi/inc
 EFI_INCLUDES	= -nostdinc -I$(TOPDIR)/Cryptlib -I$(TOPDIR)/Cryptlib/Include \
 		  -I$(EFI_INCLUDE) -I$(EFI_INCLUDE)/$(ARCH) -I$(EFI_INCLUDE)/protocol \
 		  -I$(TOPDIR)/include -iquote $(TOPDIR) -iquote $(shell pwd)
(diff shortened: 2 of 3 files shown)


mjg59:
Doesn't currently build because of the CHAR8 issue. Should we mirror gnu-efi with necessary patches into rh-boot and point at that instead?

mjg59:
Oh, and probably needs the CI updated to pull submodules?

dvzrv:
> and since distributions are often trying to build versions of shim against whatever version of gnu-efi they are shipping, this can result in awkward build failures.

IMHO, the solution to this should not be to vendor a patched version of gnu-efi then, but to *specifically require* a version greater than `x`, that is known to work with shim. Required patches should be sent upstream (to the gnu-efi project).

The errors I have encountered so far when building the shim package for Arch Linux were related to `-Werror` being used everywhere (breaking on updates to the toolchain) and not because of gnu-efi. Can you elaborate on what kind of errors users are facing, that led to the made assumption?

Looking at [the list of stable releases of gnu-efi](https://sourceforge.net/projects/gnu-efi/files/), one can see, that the project is rather slow-moving. Depending on distribution, [the provided package versions](https://repology.org/project/gnu-efi/versions) are currently between 3.0.2 and 3.0.12 (some of which are shipped by EOL distribution releases). Depending on the release of shim it should be possible to lock the range of compatible versions with the build system (e.g. *"release 15 is compatible with gnu-efi > 3.0.10"*), which can provide a sane and tested assumption.

> Given static linking, this will also make reproducible builds easier.

Under that assumption it does actually not make a difference, because if a gnu-efi package on a distribution is bit-by-bit reproducible (which it can be and is on Arch Linux and likely on other distributions as well), so will be shim (unless shim has issues with reproducibility itself).

mjg59:
The problem is more newer versions of gnu-efi breaking assumptions that shim is making. From the reproducible build perspective, the goal is to move towards it being easier to compare builds from different distributions - verifying binary equivalence reduces review overhead significantly.

martinezjavier:
I think this PR can be closed now since has moved to #308.

PR #285, opened 2021-02-19
Add --set-section-alignment '.sbat=512' to objcopy command line

jsetje:
Save the next person to cut and paste that objcopy invocation from pain.


vathpela (changes requested):
As we discussed elsewhere, this should change our actual makefile at the same time.

jsetje:
Yes, obviously. Thank you!

vathpela (approved):

Issue #281, opened 2021-02-19
Failed to load MokManager with SBAT


lcp:
When I tried to load mmx64.efi, shim failed with the following message:

SBAT data not found

In handle_image(), it only sets SBATBase and SBATSize only if the .sbat section fulfills the conditon "Section->SizeOfRawData == Section->Misc.VirtualSize". However, the .sbat section from my mmx64.efi looked like this:

00000220  00 00 00 00 40 00 60 c0  2e 73 62 61 74 00 00 00  |....@.`..sbat...|
00000230  83 00 00 00 00 f0 09 00  00 02 00 00 00 72 09 00  |.............r..|
00000240  00 00 00 00 00 00 00 00  00 00 00 00 00 00 10 40  |...............@|
*
00097200  73 62 61 74 2c 31 2c 53  42 41 54 20 56 65 72 73  |sbat,1,SBAT Vers|
00097210  69 6f 6e 2c 73 62 61 74  2c 31 2c 68 74 74 70 73  |ion,sbat,1,https|
00097220  3a 2f 2f 67 69 74 68 75  62 2e 63 6f 6d 2f 72 68  |://github.com/rh|
00097230  62 6f 6f 74 2f 73 68 69  6d 2f 62 6c 6f 62 2f 6d  |boot/shim/blob/m|
00097240  61 69 6e 2f 53 42 41 54  2e 6d 64 0a 73 68 69 6d  |ain/SBAT.md.shim|
00097250  2c 30 2c 55 45 46 49 20  73 68 69 6d 2c 73 68 69  |,0,UEFI shim,shi|
00097260  6d 2c 30 2c 68 74 74 70  73 3a 2f 2f 67 69 74 68  |m,0,https://gith|
00097270  75 62 2e 63 6f 6d 2f 72  68 62 6f 6f 74 2f 73 68  |ub.com/rhboot/sh|
00097280  69 6d 0a 00 00 00 00 00  00 00 00 00 00 00 00 00  |im..............|

Per the section header at 0x228, Misc.VirtualSize is 0x83 and SizeOfRawData is 0x200. We have to tweak the build rule for .sbat.

PR #286, opened 2021-02-20
sbat: Don't assume VirtualSize == SizeOfRawData

chrisccoulson:
The current code rejects the .sbat section if the VirtualSize and
SizeOfRawData fields of the section header aren't the same, but this
is too strict. The VirtualSize corresponds to the size of the SBAT
metadata and isn't necessarily a multiple of the file alignment.
The on-disk size (SizeOfRawData) is a multiple of the file alignment
and is zero padded. Make sure that the on-disk size is at least as
large as virtual size, and ignore the .sbat section if it isn't.
    
This should fix https://github.com/rhboot/shim/issues/281.
   
Signed-off-by: Chris Coulson <chris.coulson@canonical.com>

### DIFF
--- a/pe.c
+++ b/pe.c
@@ -1043,10 +1043,14 @@ handle_image (void *data, unsigned int datasize,
 				return EFI_UNSUPPORTED;
 			}
 
-			/* If it has nonzero size, and our bounds check made
-			 * sense, sizes match, then we believe it's okay. */
+			/* The virtual size corresponds to the size of the SBAT
+			 * metadata and isn't necessarily a multiple of the file
+			 * alignment. The on-disk size is a multiple of the file
+			 * alignment and is zero padded. Make sure that the
+			 * on-disk size is at least as large as virtual size,
+			 * and ignore the section if it isn't. */
 			if (Section->SizeOfRawData &&
-			    Section->SizeOfRawData == Section->Misc.VirtualSize &&
+			    Section->SizeOfRawData >= Section->Misc.VirtualSize &&
 			    base && end) {
 				SBATBase = base;
 				/* +1 because of size vs last byte location */


pe.c
```diff
@@ -1043,10 +1043,14 @@ handle_image (void *data, unsigned int datasize,
...
-			 * sense, sizes match, then we believe it's okay. */
+			/* The virtual size corresponds to the size of the SBAT
+			 * metadata and isn't necessarily a multiple of the file
+			 * alignment. The on-disk size is a multiple of the file
+			 * alignment and is zero padded. Make sure that the
+			 * on-disk size is at least as large as virtual size,
+			 * and ignore the section if it isn't. */
 			if (Section->SizeOfRawData &&
-			    Section->SizeOfRawData == Section->Misc.VirtualSize &&
+			    Section->SizeOfRawData >= Section->Misc.VirtualSize &&
```
vathpela, line 1053: Giant case of nit-picking:
You're right that we have to handle SizeOfRawData > VirtualSize, but I think we may actually need to allow SizeOfRawData <  VirtualSize as well in order to be correct - though probably we should handle that in our "end" computation.  From the PE spec:
|Offset|Size|Field|Description|
|--|--|--|--|
|8|4|VirtualSize|The total size of the section when loaded into memory. If this value is greater than SizeOfRawData, the  section is zero-padded. This field is valid only for executable images and should be set to zero for object files.|
|16|4|SizeOfRawData|The size of the section (for object files) or the size of the initialized data on disk (for image files).  or executable images, this must be a multiple of FileAlignment from the optional header. If this is less than VirtualSize,  the remainder of the section is zero-filled. Because the SizeOfRawData field is rounded but the VirtualSize field is not, it is possible for SizeOfRawData to be greater than VirtualSize as well. When a section contains only uninitialized data, this field should be zero.|

Just so I don't have to have a page of data in my examples, let's pretend FileAlignment is 7 bytes.  As I read it, in the case someone makes the very minor and understandable error of including a .sbat section that is, say: "shim,0\n\000", the compiler is allowed to give us a binary where SizeOfRawData == 7 and VirtualSize == 8.  In this case we should be padding out the section after loading it, though of course for .sbat validation nothing in practice will change, because we're not going to interpret all the "\000" padding anyway.

I don't think this needs to hold back this patch, and I've filed it as https://github.com/rhboot/shim/issues/287

vathpela (approved):

PR #288, opened 2021-02-21
Drop --set-section-alignment from Makefile and clarify in SBAT.md

jsetje:


SBAT.md
```diff
@@ -336,6 +336,11 @@ objcopy --set-section-alignment '.sbat=512' --add-section .sbat=sbat.csv foo.efi
 
 ```
 
+Older versions of objcopy(1) do not support --set-section-alignment
+which is required to force the correct alignment expected from a PE
+file. As long as there is another later, step of the build process,
```
martinezjavier, line 341: This sentence seems a bit off to me. Maybe instead:

"As long as there is another step, later in the build process, such as a linker invocation
that forces alignment.."

jsetje: Fixed, thank you!

martinezjavier (approved):

PR #291, opened 2021-02-22
Fix check for start in parse_sbat_var()

aburmash:
Check data before even trying do to something with it.

### DIFF
--- a/sbat.c
+++ b/sbat.c
@@ -287,6 +287,9 @@ parse_sbat_var(list_t *entries)
 			NULL,
 		};
 		for (i = 0; i < 3; i++) {
+			if (!start) {
+				goto error;
+			}
 			const CHAR8 *tmp;
 			/*
 			 * on third iteration we check if we had extra stuff on line while parsing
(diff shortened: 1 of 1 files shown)


sbat.c
```diff
@@ -287,14 +287,17 @@ parse_sbat_var(list_t *entries)
 			NULL,
 		};
 		for (i = 0; i < 3; i++) {
+			if (!start) {
```
vathpela, line 290: If start is NULL here, we've already dereferenced it wrong at https://github.com/rhboot/shim/blob/125cf985d22b32f561103702381bd73fb864d0ac/sbat.c#L280 and https://github.com/rhboot/shim/blob/125cf985d22b32f561103702381bd73fb864d0ac/sbat.c#L285 , and get_variable() should also have failed and returned an error.

It's probably better to fix get_sbat_field() to return a pointer to the NUL terminator on the string like strchrnul() does. Something like https://github.com/vathpela/mallory/commit/e3a88f0c681331d1ac1d0906a38c68ab696652da maybe?

vathpela (changes requested):

sbat.c
aburmash: Agree. @martinezjavier this should be fine while parsing the binary sbat headers as well?

martinezjavier: @aburmash yes, returning a pointer to the NUL terminator shouldn't affect the parsing of the `.sbat` section.

PR #292, opened 2021-02-22
Fix two errant 'shim,0' outdated sbat cases.

vathpela:
Two places we missed still have 0 for an sbat version - one doc and one
in our data csv.

This fixes those.


steve-mcintyre (changes requested):
Mostly looks good to me. except...

Shouldn't we also change the sbat.csv shim version to be non-zero? Dunno if 16 is correct before a release, but I'd at least say "15" or "15+" maybe?

martinezjavier (approved):

steve-mcintyre:
Good enough for me, go...

jsetje (approved):

PR #293, opened 2021-02-23
sbat: fix the residual "resource section" for SBAT

lcp:
Replace the resource section with SBAT section.


martinezjavier (approved):
Looks good to me.

PR #294, opened 2021-02-24
If the SBAT UEFI variable is not set, initialize it as a bootservicesvariable

jsetje:
I expect this will generate comments on where exactly to do the initialization.

This should probably use get_variable(), but there's no set_variable, and we need the flags.

Making sbat initialization non-fatal for non-secure mode may also be controversial.

### DIFF
--- a/include/sbat.h
+++ b/include/sbat.h
@@ -6,6 +6,16 @@
 #ifndef SBAT_H_
 #define SBAT_H_
 
+#define SBAT_VAR_SIG "sbat,"
+#define SBAT_VAR_VERSION "1,"
+#define SBAT_VAR_DATE "2021030218"
+#define SBAT_VAR SBAT_VAR_SIG SBAT_VAR_VERSION SBAT_VAR_DATE "\n"
+
+#define UEFI_VAR_NV_BS \
+	(EFI_VARIABLE_NON_VOLATILE | EFI_VARIABLE_BOOTSERVICE_ACCESS)
+#define UEFI_VAR_NV_BS_TIMEAUTH \
+	(UEFI_VAR_NV_BS | EFI_VARIABLE_TIME_BASED_AUTHENTICATED_WRITE_ACCESS)
+
 extern UINTN _sbat, _esbat;
 
 struct sbat_var_entry {
@@ -23,6 +33,7 @@ extern list_t sbat_var;
 
 EFI_STATUS parse_sbat_var(list_t *entries);
 void cleanup_sbat_var(list_t *entries);
+EFI_STATUS set_sbat_uefi_variable(void);
 
 struct sbat_section_entry {
 	const CHAR8 *component_name;
--- a/sbat.c
+++ b/sbat.c
@@ -293,4 +293,85 @@ parse_sbat_var(list_t *entries)
 	return parse_sbat_var_data(entries, data, datasize+1);
 }
 
+EFI_STATUS
+set_sbat_uefi_variable(void)
+{
+	EFI_STATUS efi_status = EFI_SUCCESS;
+	UINT32 attributes = 0;
+
+	UINT8 *sbat = NULL;
+	UINTN sbatsize = 0;
+
+	efi_status = get_variable_size(L"SBAT", SHIM_LOCK_GUID, &sbatsize);
+	if (EFI_ERROR(efi_status))
+		dprint(L"SBAT size probe failed %r\n", efi_status);
+
+	if (sbatsize != 0) {
+		sbat = AllocateZeroPool(sbatsize + 1);
+		if (!sbat)
+			return EFI_OUT_OF_RESOURCES;
+
+		efi_status = get_variable_attr(L"SBAT", &sbat, &sbatsize,
+				SHIM_LOCK_GUID, &attributes);
+	}
+
+	/*
+	 * Always set the SBAT UEFI variable if it fails to read.
+	 *
+	 * Don't try to set the SBAT UEFI variable if attributes match, the
+	 * signature matches and it has the same or newer version.
+	 */
+	if (EFI_ERROR(efi_status)) {
+		dprint(L"SBAT read failed %r\n", efi_status);
+	} else if ((attributes == UEFI_VAR_NV_BS ||
+				attributes == UEFI_VAR_NV_BS_TIMEAUTH) &&
+                   strncmp(sbat, SBAT_VAR_SIG, strlen(SBAT_VAR_SIG)) == 0) {
+		FreePool(sbat);
+		return EFI_SUCCESS;
+	}
+
+	/* delete previous variable */
+	efi_status = set_variable(L"SBAT", SHIM_LOCK_GUID, attributes, 0, "");
+	if (EFI_ERROR(efi_status))
+		dprint(L"SBAT variable delete failed %r\n", efi_status);
+
+	/* verify that it's gone */
+	efi_status = get_variable(L"SBAT", &sbat, &sbatsize, SHIM_LOCK_GUID);
+	if (EFI_ERROR(efi_status) || (sbatsize != 0))
+		dprint(L"SBAT variable clearing failed %r\n", efi_status);
+
+	/* set variable */
+	efi_status = set_variable(L"SBAT", SHIM_LOCK_GUID,
+			UEFI_VAR_NV_BS, sizeof (SBAT_VAR), SBAT_VAR);
+	if (EFI_ERROR(efi_status))
+		dprint(L"SBAT variable writing failed %r\n", efi_status);
+
+	FreePool(sbat);
+
+	/* verify that the expected data is there */
+	efi_status = get_variable_size(L"SBAT", SHIM_LOCK_GUID, &sbatsize);
+	if (EFI_ERROR(efi_status) || sbatsize == 0) {
+		dprint(L"SBAT read failed %r\n", efi_status);
+		return EFI_INVALID_PARAMETER;
+	}
+
+	sbat = AllocateZeroPool(sbatsize);
+	if (!sbat)
+		return EFI_OUT_OF_RESOURCES;
+
+	efi_status = get_variable(L"SBAT", &sbat, &sbatsize, SHIM_LOCK_GUID);
+	if (EFI_ERROR(efi_status))
+		dprint(L"SBAT read failed %r\n", efi_status);
+
+	if (strncmp(sbat, SBAT_VAR, strlen(SBAT_VAR)) != 0) {
+		efi_status = EFI_INVALID_PARAMETER;
+	} else {
+		dprint(L"SBAT variable initialization succeeded\n");
+		efi_status = EFI_SUCCESS;
+	}
+
+	FreePool(sbat);
+	return efi_status;
+}
+
 // vim:fenc=utf-8:tw=75:noet
--- a/shim.c
+++ b/shim.c
@@ -1759,7 +1759,8 @@ shim_init(void)
 void
 shim_fini(void)
 {
-	cleanup_sbat_var(&sbat_var);
+	if (secure_mode())
+		cleanup_sbat_var(&sbat_var);
 
 	/*
 	 * Remove our protocols
@@ -1865,6 +1866,7 @@ efi_main (EFI_HANDLE passed_image_handle, EFI_SYSTEM_TABLE *passed_systab)
 		L"shim_init() failed",
 		L"import of SBAT data failed",
 		L"SBAT self-check failed",
+		L"SBAT UEFI variable setting failed",
 		NULL
 	};
 	enum {
@@ -1872,6 +1874,7 @@ efi_main (EFI_HANDLE passed_image_handle, EFI_SYSTEM_TABLE *passed_systab)
 		SHIM_INIT,
 		IMPORT_SBAT,
 		SBAT_SELF_CHECK,
+		SET_SBAT,
 	} msg = IMPORT_MOK_STATE;
 
 	/*
@@ -1901,24 +1904,27 @@ efi_main (EFI_HANDLE passed_image_handle, EFI_SYSTEM_TABLE *passed_systab)
 	 */
 	debug_hook();
 
-	INIT_LIST_HEAD(&sbat_var);
-	efi_status = parse_sbat_var(&sbat_var);
-	/*
-	 * Until a SBAT variable is installed into the systems, it is expected that
-	 * attempting to parse the variable will fail with an EFI_NOT_FOUND error.
-	 *
-	 * Do not consider that error fatal for now.
-	 */
-	if (EFI_ERROR(efi_status) && efi_status != EFI_NOT_FOUND) {
-		perror(L"Parsing SBAT variable failed: %r\n",
-		       efi_status);
-		msg = IMPORT_SBAT;
-		goto die;
-	}
-
-	if (secure_mode ()) {
+	if (secure_mode()) {
 		char *sbat_start = (char *)&_sbat;
 		char *sbat_end = (char *)&_esbat;
+
+		INIT_LIST_HEAD(&sbat_var);
+		efi_status = parse_sbat_var(&sbat_var);
+		if (EFI_ERROR(efi_status)) {
+			efi_status = set_sbat_uefi_variable();
+			if (efi_status == EFI_INVALID_PARAMETER) {
+				perror(L"SBAT variable initialization failed\n");
+				msg = SET_SBAT;
+				goto die;
+			}
+			efi_status = parse_sbat_var(&sbat_var);
+			if (EFI_ERROR(efi_status)) {
+				perror(L"Parsing SBAT variable failed: %r\n",
+					efi_status);
+				msg = IMPORT_SBAT;
+				goto die;
+			}
+		}
 
 		efi_status = handle_sbat(sbat_start, sbat_end - sbat_start);
 		if (EFI_ERROR(efi_status)) {
@@ -1927,6 +1933,14 @@ efi_main (EFI_HANDLE passed_image_handle, EFI_SYSTEM_TABLE *passed_systab)
 			msg = SBAT_SELF_CHECK;;
 			goto die;
 		}
+	}
+
+	efi_status = set_sbat_uefi_variable();
+	if (efi_status == EFI_INVALID_PARAMETER) {
+		perror(L"SBAT variable initialization failed\n");
+		msg = SET_SBAT;
+		if (secure_mode())
+                        goto die;
 	}
 
 	init_openssl();


include/sbat.h
```diff
@@ -6,6 +6,9 @@
 #ifndef SBAT_H_
 #define SBAT_H_
 
+#define SBAT_INIT "sbat,1,2021030218\n"
```
martinezjavier, line 9: What would be the policy to update this? I wonder if we should have a `sbat_version.h.in` or something with a `@@SBAT_DATE@@` that's replaced with the build date ?

include/sbat.h
```diff
@@ -17,6 +20,7 @@ extern list_t sbat_var;
 
 EFI_STATUS parse_sbat_var(list_t *entries);
 void cleanup_sbat_var(list_t *entries);
+EFI_STATUS init_sbat();
```
martinezjavier, line 23: please use `init_sbat(void)` here.

sbat.c
```diff
@@ -335,4 +335,62 @@ parse_sbat_var(list_t *entries)
 	FreePool(data);
 	return EFI_INVALID_PARAMETER;
 }
+
+EFI_STATUS
+init_sbat()
```
martinezjavier, line 340: same here.

sbat.c
```diff
@@ -335,4 +335,62 @@ parse_sbat_var(list_t *entries)
...
+{
+	EFI_STATUS efi_status = EFI_SUCCESS;
+	UINT32 attributes = 0;
+
+	CHAR8 *sbat;
+	UINTN sbatsize = 4096;
```
martinezjavier, line 346: what happens if the SBAT variable size is bigger than 4 KiB ?

sbat.c
```diff
@@ -335,4 +335,62 @@ parse_sbat_var(list_t *entries)
...
+
+	sbat = AllocateZeroPool(sbatsize);
+
+	/* read and get check attributes */
+	efi_status = gRT->GetVariable(L"SBAT", &SHIM_LOCK_GUID, &attributes,
+			&sbatsize, sbat);
+
```
martinezjavier, line 353: I need to check of the `GetVariable()` EFI service will behave here but I wonder if we shouldn't first do a call with `sbat == NULL` to get the size and allocate the buffer if the return value is `EFI_BUFFER_TOO_SMALL` or something like that.

Another option is to use one of the variable API from `gnu-efi`, for example `LibGetVariable()` or `LibGetVariableAndSize()`.

sbat.c
```diff
@@ -335,4 +335,62 @@ parse_sbat_var(list_t *entries)
...
+	 * If attributes and the initial part of the version match, then
+	 * there is no need to re-write it. If it is not set or the read
+	 * failed, then attributes won't match.
+	 */
+	if (((attributes != (EFI_VARIABLE_NON_VOLATILE |
+		EFI_VARIABLE_BOOTSERVICE_ACCESS)) ||
+		(attributes != (EFI_VARIABLE_NON_VOLATILE |
+				EFI_VARIABLE_BOOTSERVICE_ACCESS |
+				EFI_VARIABLE_TIME_BASED_AUTHENTICATED_WRITE_ACCESS)))
+			|| (strncmp(sbat, SBAT_INIT, sizeof(SBAT_SIG) - 1) != 0)) {
```
martinezjavier, line 364: Whey is `SBAT_INIT` used here instead of just `SBAT_SIG` ?

sbat.c
```diff
@@ -335,4 +335,62 @@ parse_sbat_var(list_t *entries)
...
+		efi_status = gRT->SetVariable(L"SBAT", &SHIM_LOCK_GUID,
+				EFI_VARIABLE_NON_VOLATILE |
+				EFI_VARIABLE_BOOTSERVICE_ACCESS,
+				sizeof(SBAT_INIT), SBAT_INIT);
+
+		/* verify that the expected data is there*/
+		ZeroMem(sbat, sbatsize);
+		efi_status = gRT->GetVariable(L"SBAT", &SHIM_LOCK_GUID,
+				0, &sbatsize, sbat);
+		if (strncmp(sbat, SBAT_INIT, sizeof(SBAT_INIT)) != 0) {
```
vathpela, line 384: Pretty sure you want strlen() instead of sizeof() here; see https://gist.github.com/vathpela/cb118cd7d623ce8022107fea784e6556

sbat.c
vathpela: Don't we also want to compare the version here?

shim.h
```diff
@@ -148,7 +148,6 @@
 #include "include/tpm.h"
 #include "include/ucs2.h"
 #include "include/variables.h"
-#include "include/sbat.h"
```
vathpela, line 151: I don't understand this change.

vathpela (changes requested):

sbat.c
vathpela: Can we possibly call this something else?  We now have *three* things we initialize that are effectively called "sbat":
- the `SBAT` variable in EFI
- our `SBAT` variable data: `sbat_var` nearly everywhere now
- the `.sbat` section in a binary we're loading: `sbat_section` nearly everywhere now
Maybe `set_sbat_variable()`?

include/sbat.h
vathpela: Maybe SBAT_VAR_SIG instead of SBAT_INIT?

shim.h
jsetje: It's also included on line 142, I just dropped the dup. You may or may not have already fixed this in your branch.

sbat.c
jsetje: Yes, how about: set_sbat_uefi_variable() ?

sbat.c
jsetje: Good catch!
Adding a manual test case where the next char doesn't match as well.

sbat.c
jsetje: Technically we don't care about anything more than strlen(SBAT_INIT) +1, but I'm just testing the size and allocating dynamically now.

sbat.c
jsetje: I confirmed that calling with GetVariable() 0 and NULL returns the size. Fixing.

include/sbat.h
jsetje: Certainly not an automatically generated date.
and yes to SBAT_VAR_SIG

sbat.c
jsetje: After staring at variously ugly ways to do the date stamp and version comparison, I realized that strcmp() will actually do the right thing.

and yes, I should use SBAT_SIG or rather SBAT_VAR_SIG

sbat.c
```diff
@@ -335,4 +335,75 @@ parse_sbat_var(list_t *entries)
...
+	EFI_STATUS efi_status = EFI_SUCCESS;
+	UINT32 attributes = 0;
+
+	CHAR8 *sbat = NULL;
+	UINTN sbatsize = 0;
+
+	/* get size */
+	efi_status = gRT->GetVariable(L"SBAT", &SHIM_LOCK_GUID, &attributes,
+			&sbatsize, sbat);
+	sbat = AllocateZeroPool(sbatsize);
```
martinezjavier, line 351: Sorry, missed before but you need to check for allocation failures here and return `EFI_OUT_OF_RESOURCES`.

sbat.c
```diff
@@ -335,4 +335,75 @@ parse_sbat_var(list_t *entries)
...
+		|| (attributes != (EFI_VARIABLE_NON_VOLATILE |
+				EFI_VARIABLE_BOOTSERVICE_ACCESS |
+				EFI_VARIABLE_TIME_BASED_AUTHENTICATED_WRITE_ACCESS)))
+		|| (strncmp(sbat, SBAT_VAR_SIG, strlen(SBAT_VAR_SIG)) != 0)
+		|| (strncmp(SBAT_VAR, sbat, strlen(SBAT_VAR)) > 0)
+		) {
+
+		/* delete previous variable */
+		efi_status = gRT->SetVariable(L"SBAT", &SHIM_LOCK_GUID,
+				attributes, 0, "");
```
martinezjavier, line 376: I think we want to check returned EFI status here and report if failed to set.

sbat.c
```diff
@@ -335,4 +335,75 @@ parse_sbat_var(list_t *entries)
...
+		efi_status = gRT->SetVariable(L"SBAT", &SHIM_LOCK_GUID,
+				EFI_VARIABLE_NON_VOLATILE |
+				EFI_VARIABLE_BOOTSERVICE_ACCESS,
+				sizeof (SBAT_VAR), SBAT_VAR);
+
+		/* verify that the expected data is there */
+		FreePool(sbat);
+		efi_status = gRT->GetVariable(L"SBAT", &SHIM_LOCK_GUID,
+				0, &sbatsize, NULL);
+		sbat = AllocateZeroPool(sbatsize);
```
martinezjavier, line 394: Do we need to free and allocate the buffer if we know that the data will have the same size? In any case you can check if the `sbatsize` doesn't match the expected one and report an error.

martinezjavier: I think that you could just `ZeroMem(sbat)` instead of the free + alloc.

sbat.c
jsetje: I'm adding dprintfs only. UEFI failures should not prevent use from ultimately trying to set the variable since that may in fact succeed and is unlikely to do anything worse than not setting it.

sbat.c
```diff
@@ -294,4 +294,96 @@ parse_sbat_var(list_t *entries)
...
+		efi_status = gRT->GetVariable(L"SBAT", &SHIM_LOCK_GUID,
+				0, &sbatsize, sbat);
+		if ((efi_status != EFI_SUCCESS) || (sbatsize != 0))
+			dprint(L"SBAT variable clearing failed %r\n", efi_status);
+
+		/* set variable */
+		efi_status = gRT->SetVariable(L"SBAT", &SHIM_LOCK_GUID,
+				EFI_VARIABLE_NON_VOLATILE |
+				EFI_VARIABLE_BOOTSERVICE_ACCESS,
+				sizeof (SBAT_VAR), SBAT_VAR);
```
martinezjavier, line 356: shouldn't this be `strlen()` instead of `sizeof()` ?

sbat.c
martinezjavier: You are actually correct because the `sbatsize` is the size of the original read variable while the second allocation should match `strlen(SBAT_VAR)`.

Probably I would add a check since there's no point to compare sbat and SBAT_VAR if the sizes don't match.

martinezjavier (approved):

shim.h
jsetje: That's already fixed in your current tip and was dropped by my rebase.

sbat.c
```diff
@@ -294,4 +294,96 @@ parse_sbat_var(list_t *entries)
...
+	 * the version is more significant than the date, strcmp() applies.
+	 */
+	if (((attributes != (EFI_VARIABLE_NON_VOLATILE |
+		EFI_VARIABLE_BOOTSERVICE_ACCESS))
+		|| (attributes != (EFI_VARIABLE_NON_VOLATILE |
+				EFI_VARIABLE_BOOTSERVICE_ACCESS |
+				EFI_VARIABLE_TIME_BASED_AUTHENTICATED_WRITE_ACCESS)))
+		|| (strncmp(sbat, SBAT_VAR_SIG, strlen(SBAT_VAR_SIG)) != 0)
+		|| (strncmp(SBAT_VAR, sbat, strlen(SBAT_VAR)) > 0)
+		) {
```
vathpela, line 338: You probably should consider adding something like the following before this function:
```
#define NV_BS (EFI_VARIABLE_NON_VOLATILE | EFI_VARIABLE_BOOTSERVICE_ACCESS)
#define NV_BS_TIMEAUTH \
        (NV_BS | EFI_VARIABLE_TIME_BASED_AUTHENTICATED_WRITE_ACCESS)
```
And then making this be:
```
        /*
         * lorem ipsum ...
         */
        if (!EFI_ERROR(efi_status) &&
            (attributes != NV_BS || attributes != NV_BS_TIMEAUTH ||
             strncmp(sbat, SBAT_VAR_SIG, strlen(SBAT_VAR_SIG)) == 0 ||
             strncmp(sbat, SBAT_VAR, strlen(SBAT_VAR)) <= 0)) {
                FreePool(sbat);
                return EFI_SUCCESS;
        }
        ...
```

sbat.c
```diff
@@ -294,4 +294,96 @@ parse_sbat_var(list_t *entries)
...
+		/* delete previous variable */
+		efi_status = gRT->SetVariable(L"SBAT", &SHIM_LOCK_GUID,
+				attributes, 0, "");
+		if (EFI_ERROR(efi_status))
+			dprint(L"SBAT variable delete failed %r\n", efi_status);
+
+		/* verify that it's gone */
+		efi_status = gRT->GetVariable(L"SBAT", &SHIM_LOCK_GUID,
+				0, &sbatsize, sbat);
+		if ((efi_status != EFI_SUCCESS) || (sbatsize != 0))
```
vathpela, line 349: This should be `EFI_ERROR(efi_status)` as there are possible return values that aren't EFI_SUCCESS but also aren't errors.

sbat.c
```diff
@@ -294,4 +294,96 @@ parse_sbat_var(list_t *entries)
...
+		if (EFI_ERROR(efi_status))
+			dprint(L"SBAT variable writing failed %r\n", efi_status);
+
+		/* verify that the expected data is there */
+		efi_status = gRT->GetVariable(L"SBAT", &SHIM_LOCK_GUID,
+				0, &sbatsize, NULL);
+		if (EFI_ERROR(efi_status))
+			dprint(L"SBAT read failed %r\n", efi_status);
+
+		/* re-allocate in case of HV intervention  */
```
vathpela, line 366: I have no idea what this comment means.  We need to reallocate anyway in order to fit our new variable, which is most likely larger.  Maybe nix the comment in favor of saying so?

sbat.c
```diff
@@ -294,4 +294,96 @@ parse_sbat_var(list_t *entries)
...
+		FreePool(sbat);
+		sbat = AllocateZeroPool(sbatsize);
+		if (!sbat)
+			return EFI_OUT_OF_RESOURCES;
+
+		efi_status = gRT->GetVariable(L"SBAT", &SHIM_LOCK_GUID,
+				0, &sbatsize, sbat);
```
vathpela, line 373: please use NULL and not 0 (or just use the wrapper function we already have...)

vathpela (changes requested):

vathpela:
> This should probably use get_variable(), but there's no set_variable, and we need the flags.

It really, really should, and so I've added get_variable_size() and set_variable() in https://github.com/rhboot/shim/pull/301

sbat.c
```diff
@@ -294,4 +294,96 @@ parse_sbat_var(list_t *entries)
...
+	EFI_STATUS efi_status = EFI_SUCCESS;
+	UINT32 attributes = 0;
+
+	CHAR8 *sbat = NULL;
+	UINTN sbatsize = 0;
+
+	/* get size */
+	efi_status = gRT->GetVariable(L"SBAT", &SHIM_LOCK_GUID, &attributes,
+			&sbatsize, sbat);
+	if (EFI_ERROR(efi_status))
```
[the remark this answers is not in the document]
jsetje: Adding an exception for EFI_BUFFER_TOO_SMALL both here and on the other size check.

sbat.c
```diff
@@ -294,4 +294,96 @@ parse_sbat_var(list_t *entries)
...
+		dprint(L"SBAT read failed %r\n", efi_status);
+
+	sbat = AllocateZeroPool(sbatsize);
```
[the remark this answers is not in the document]
jsetje: Good catch, that does look like one could game it. Fixing this, and for sanity sake the other alloc as well (even if we don't use that data).

sbat.c
jsetje: Doing the cleanup, but I do think we want to drive on into this if there's an error status on the read. Best case we succeed and set the variable despite it. Worst case we fail, which is the same as giving up.

sbat.c
jsetje: Done and yes, the primary point of that action is clear.

sbat.c
jsetje: thanks, I fixed the other two calls as well.

sbat.c
vathpela: Would something like this be clearer to you:
```
        if (EFI_ERROR(efi_status)) {
                dprint(L"SBAT read failed %r\n", efi_status);
        } else if ((attributes == NV_BS || attributes == NV_BS_TIME_AUTH) &&
                   (strncmp(sbat, SBAT_VAR_SIG, strlen(SBAT_VAR_SIG)) == 0 ||
                    strncmp(sbat, SBAT_VAR, strlen(SBAT_VAR)) <= 0)) {
                FreePool(sbat);
                return EFI_SUCCESS;
        }

```

vathpela: (I think I've got the negation right that time, but certainly double check it.)

jsetje: Yes, that one is right. Thank you.

shim.c
```diff
@@ -1895,25 +1896,20 @@ efi_main (EFI_HANDLE passed_image_handle, EFI_SYSTEM_TABLE *passed_systab)
...
 		char *sbat_start = (char *)&_sbat;
 		char *sbat_end = (char *)&_esbat;
 
+		INIT_LIST_HEAD(&sbat_var);
+		efi_status = parse_sbat_var(&sbat_var);
+		if (EFI_ERROR(efi_status)) {
+			perror(L"Parsing SBAT variable failed: %r\n",
+					efi_status);
+			msg = IMPORT_SBAT;
+			efi_status = set_sbat_uefi_variable();
```
lcp, line 1909: The return status of set_sbat_uefi_variable() wasn't checked and it caused the strange failure during the first boot:

shim.c:1891:efi_main() vendor_deauthorized:0x7DDDB4F9 vendor_deauthorized_size:10684
Parsing SBAT variable failed: Not Found
sbat.c:309:set_sbat_uefi_variable() SBAT read failed Not Found
sbat.c:326:set_sbat_uefi_variable() SBAT read failed Not Found
sbat.c:339:set_sbat_uefi_variable() SBAT variable delete failed Not Found
sbat.c:345:set_sbat_uefi_variable() SBAT variable clearing failed Not Found
sbat.c:372:set_sbat_uefi_variable() SBAT variable initialization succeeded
Something has gone seriously wrong: import of SBAT data failed: Success

shim.c
```diff
@@ -1923,6 +1919,15 @@ efi_main (EFI_HANDLE passed_image_handle, EFI_SYSTEM_TABLE *passed_systab)
...
 
+	efi_status = set_sbat_uefi_variable();
+	if (efi_status == EFI_INVALID_PARAMETER) {
+		perror(L"SBAT variable initialization failed\n");
+		msg = IMPORT_SBAT;
+		if (secure_mode())
+                        goto die;
+	}
+
+
```
lcp, line 1930: Nitpicking: extra newline.

sbat.c
```diff
@@ -294,4 +294,88 @@ parse_sbat_var(list_t *entries)
...
+				attributes == UEFI_VAR_NV_BS_TIMEAUTH) &&
+                   (strncmp(sbat, SBAT_VAR_SIG, strlen(SBAT_VAR_SIG)) == 0 ||
+                    strncmp(sbat, SBAT_VAR, strlen(SBAT_VAR)) <= 0)) {
+		FreePool(sbat);
+		return EFI_SUCCESS;
+	}
+
+	/* delete previous variable */
+	efi_status = gRT->SetVariable(L"SBAT", &SHIM_LOCK_GUID, attributes,
+		       	0, "");
```
lcp, line 338: Nitpicking: extra whitespaces were prepended to 0.

shim.c
jsetje: Thank you! It's not just the missing check, first boot in secure mode isn't right.

This needs to be:
  - try to import / parse sbat
  - try to set variable
  - try to import again

jsetje: @martinezjavier I think I fixed this last night, but can you take another quick look at the ordering?

jsetje: I see why I didn't think I had this quite right last night, adding a check to skip the setting if we succeeded previously.

jsetje: Actually no, this is fine as is now (after fixing the issue pointed out by @lcp). set_sbat_uefi_variable() will run again and see that no work is to be done, or try one more time if the variable magically vanished after previously reading back. (Though that's a super broken system.)

jsetje:
Does anyone that's looked at this have concerns about moving on to the right interfaces:

https://github.com/rhboot/shim/pull/302

If not, I'll push this series here to preserve the comment history and will kill the other branch.

include/sbat.h
```diff
@@ -6,6 +6,16 @@
...
 
+#define SBAT_VAR_SIG "sbat,"
+#define SBAT_VAR_VERSION "1,"
+#define SBAT_VAR_DATE "2021030218"
+#define SBAT_VAR SBAT_VAR_SIG SBAT_VAR_VERSION SBAT_VAR_DATE "\n"
+
+#define UEFI_VAR_NV_BS \
+	(EFI_VARIABLE_NON_VOLATILE | EFI_VARIABLE_BOOTSERVICE_ACCESS)
+#define UEFI_VAR_NV_BS_TIMEAUTH \
+	(UEFI_VAR_NV_BS | EFI_VARIABLE_TIME_BASED_AUTHENTICATED_WRITE_ACCESS)
```
superm1, line 17: so if this is initialized as bootservice access only, how exactly will "new" signed SBAT variables be able to flow down?  I would think you want updates coming from OS userspace.

jsetje: The transition will involve an updated shim,

vathpela:
Merged as https://github.com/rhboot/shim/pull/314/commits/465ecd395343c70946ecb676144b6f9e18a6fb1e

PR #296, opened 2021-02-25
More SBAT parser reworking and lots of test cases

vathpela:
This patch set does a couple of things:
- adds a generic CSV parser that builds a linked list of rows with an array of column pointers from a csv file
- makes parsing .sbat use it
- makes parsing the SBAT variable use it
- adds a test suite for (some of) our string functions, including the ones parsing csv use, with lots of specific tests
- adds a test suite for our csv parser, with lots of specific tests, including several that fuzz it with 8.5kB of /dev/urandom
- adds a test suite for parsing .sbat and the SBAT variable (does not currently test verify_sbat())

Currently it is not possible to /run/ the test suite on this branch; you need two more patches that currently live on https://github.com/rhboot/shim/tree/test-reorg


sbat.c
```diff
@@ -25,51 +25,101 @@ get_sbat_field(CHAR8 *current, CHAR8 *end, const CHAR8 **field, char delim)
...
-	*current = get_sbat_field(*current, end, &entry->component_name, ',');
-	if (!entry->component_name)
-		goto error;
+	INIT_LIST_HEAD(&csv);
 
-	*current = get_sbat_field(*current, end, &entry->component_generation,
-	                          ',');
-	if (!entry->component_generation)
-		goto error;
+	efi_status = parse_csv_data(section_base, end, 6, &csv);
```
martinezjavier, line 27: Should we have a `#define SBAT_SECTION_COLUMNS 6` or something in `sbat.h` instead of having a magic number here? I know is unlikely that this would change but still would make it easier to read.

sbat.c
```diff
@@ -25,51 +25,101 @@ get_sbat_field(CHAR8 *current, CHAR8 *end, const CHAR8 **field, char delim)
...
+	entries = (struct sbat_section_entry **)strtab;
+	strtab += sizeof(struct sbat_section_entry *) * n;
+	entry = (struct sbat_section_entry *)strtab;
+	strtab += sizeof(struct sbat_section_entry) * n;
+	n = 0;
+
+	list_for_each(pos, &csv) {
+		struct csv_row * row;
+		size_t i;
+		const char **ptrs[6] = {
```
martinezjavier: here we could also use such a constant if added.

sbat.c
```diff
@@ -92,7 +74,7 @@ parse_sbat_section(char *section_base, size_t section_size,
 	list_for_each(pos, &csv) {
 		struct csv_row * row;
 		size_t i;
-		const char **ptrs[6] = {
+		const char **ptrs[] = {
```
martinezjavier, line 77: Right, you are removing the array size since is not really needed. Probably should be done in the previous patch.

martinezjavier (approved):
Patches looks good to me. Thanks a lot for this rework!

Since the test requires the `test-reorg` branch, I guess that the last 3 commits should be left for now and added after that ?

aburmash:
This parser even covers some weird cases like double /n/n in the end of variable and section. LGTM.
Smoke tests also show that parsing works as expected.

aburmash (approved):

vathpela:
I've pushed another batch to address javierm's concerns.  It also:
- converts more of test-csv.c and test-sbat.c to use the "newer" helper macros
- makes dprint() work (theoretically) in our tests
- adds a more testable interface for verify_sbat()

PR #301, opened 2021-02-26
Add get_variable_size() and set_variable() wrappers.

vathpela:
This get_variable_size() implementation success in either of two cases:
- EFI_SUCCESS with *lenp == 0 if the variable isn't found
- EFI_SUCCESS with *lenp > 0 on success

In the event of other errors, it returns them to you.

There's nothing particularly interesting about the set_variable()
implementation here.


martinezjavier:
I think this can be closed now since has moved to #294.

martinezjavier:
merging this as agreed with @jsetje.

PR #302, opened 2021-02-26
If the SBAT UEFI variable is not set, initialize it as a bootservices variable (rebased on size and set wrappers)

jsetje:
This is poorly tested, but seems to work.


jsetje:
I added explicit 0 size handling and squashed the commits. I also tested this under a couple more circumstances.

jsetje:
Dropping this PR and pushing this approach into the RP with the previous discussion.

PR #305, opened 2021-02-27
Add tesuite to the github pull request workflow.

xnox:


xnox:
@vathpela make[1]: xxd: No such file or directory

Not sure if you need to rebuild the ci containers, or if dns install will work.

Also are the containers meant to be static? or does one meant to upgrade the toolchain in them?

xnox:
so this now passes workflow, although tests are simply attempted to build and then skip. Cause they don't seem to build anymore.

xnox:
Also it seems to pull from f35 without f35 gpg key in the container. I don't know how to correctly import the f35 key and mark it as trusted =/

PR #308, opened 2021-03-06
Try to do a better job with gnu-efi and all the string problems...

vathpela:
This is basically built on https://github.com/rhboot/shim/pull/284 , but with some fixes to make it build the whole time.


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "gnu-efi"]
+	path = gnu-efi
+	url = https://git.code.sf.net/p/gnu-efi/code
```
martinezjavier: I think we should just use the shim specific `gnu-efi` here instead of doing it as a follow-up separate patch.

Also, this patch has your authorship but mjg's Signed-off-by tag.

jsetje (approved):

martinezjavier (approved):
Looks good to me now.

PR #309, opened 2021-03-09
Set the section flags for .sbat

lcp:
When using "objcopy -O binary", it silently drops the sections without
"alloc" or "load" or the sections with "unload". Since we didn't set any
section flags for .sbat, it just contains the "readonly" flag and
objcopy ignored the section totally when generating EFI images for
ARM32 and AArch64.

This commit sets the common read-only data section flags to .sbat so
that objcopy would always copy the section to the final EFI image.


martinezjavier (approved):
LGTM.

PR #312, opened 2021-03-09
Misc fallback program fixes from Endless OS

jprvita:
These are a few misc fixes and improvements to the fallback program that we have been carrying downstream at Endless OS.

These changes we already reviewed twice via the community review process on https://github.com/rhboot/shim-review/issues/61 and https://github.com/rhboot/shim-review/issues/105.

### DIFF
--- a/fallback.c
+++ b/fallback.c
@@ -17,6 +17,9 @@ EFI_LOADED_IMAGE *this_image = NULL;
 int
 get_fallback_verbose(void)
 {
+#ifdef FALLBACK_VERBOSE
+	return 1;
+#else
 	UINT8 *data = NULL;
 	UINTN dataSize = 0;
 	EFI_STATUS efi_status;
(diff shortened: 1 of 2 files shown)


fallback.c
```diff
@@ -6,6 +6,7 @@
 
 #include <efi.h>
 #include <efilib.h>
+#include <CrtLibSupport.h>
```
vathpela, line 9: This looks incorrect, and it is breaking the CI build - but just removing it works fine.

vathpela (changes requested):
Aside from the one minor issue, it looks good to me.

ManMower:
FWIW, this series LGTM

fallback.c
jprvita: Looks like we need `OpenSslSupport.h` there, for `isxdigit`.

vathpela:
I had meant that I would pull it after the other changeset, which also obviated the need to add the second include, but oh well ;)

Anyway, pushed as 9bfdf2dea93621c9231793e2ade6b56abf7f6b8d.

PR #313, opened 2021-03-10
Fix compilation for older gcc

aburmash:

### DIFF
--- a/shim.c
+++ b/shim.c
@@ -1096,7 +1096,7 @@ EFI_STATUS start_image(EFI_HANDLE image_handle, CHAR16 *ImagePath)
 	void *sourcebuffer = NULL;
 	UINT64 sourcesize = 0;
 	void *data = NULL;
-	int datasize;
+	int datasize = 0;
 
 	/*
 	 * We need to refer to the loaded image protocol on the running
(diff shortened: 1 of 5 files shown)


shim.c
```diff
@@ -1096,7 +1096,7 @@ EFI_STATUS start_image(EFI_HANDLE image_handle, CHAR16 *ImagePath)
 	void *sourcebuffer = NULL;
 	UINT64 sourcesize = 0;
 	void *data = NULL;
-	int datasize;
+	int datasize = 0;
```
jsetje, line 1099: Should this be a separate commit?

aburmash: This still falls under category "fix compilation for older gcc". This is an issue for 4.8.5 and older(?) gcc, new enough will be fine with just int datasize;

jsetje: Actually this is fine, it is always set before reading.

jsetje (approved):

vathpela:
Pushed as 4033d1fd0f25196f9a35e944679676277cd51960, which fixes the one conflict in the OpenSSL Makefile, and adds one workaround for gcc < 9 not having __builtin_strnlen().

PR #315, opened 2021-03-11
Add support to verify fb & mm by hash.

xnox:
This is a bit WIP because currently when i test this new shim it actually fails to boot the fb & mm.

The principle is nice, instead of generating ephemeral shim cert, just verify mm & fb by hash. Useful if one uses vendor-db, as one doesn't need a full on ephemeral shim cert.


xnox:
Before considering to do this, we should make all binaries generate the same authenticode hash when they are signed or unsigned. Thus rejecting this for now.

PR #321, opened 2021-03-15
Disable a specific strncmp test on arm64

steve-mcintyre:
This test demonstrates that gnuefi_signed_strncmp() is broken on
arm64. We knew that, don't break when we're doing a build. We don't
use this implementation anyway...

Signed-off-by: Steve McIntyre <93sam@debian.org>


vathpela:
Closing this because it's a dupe of #320

PR #322, opened 2021-03-15
Stop using EFI_PATH, and error out if people try to use it

steve-mcintyre:
We need to be using our patched version of gnu-efi

Signed-off-by: Steve McIntyre <93sam@debian.org>


xnox:
I always thought of the new vendored gnu-efi as "VENDOR_EFI_PATH" not "LOCAL_" but i guess it is all the same.

steve-mcintyre:
> I always thought of the new vendored gnu-efi as "VENDOR_EFI_PATH" not "LOCAL_" but i guess it is all the same.

Happy to change to that, that's easy... :-)

PR #324, opened 2021-03-15
arm and aarch64: fix some PE headers.

vathpela:
This fixes the SizeOfImage and SizeOfInitializedData headers on arm and
aa64.


chrisccoulson:
This looks ok to me

PR #327, opened 2021-03-17
arm and aarch64: include the aligned part in SizeOfRawData of sbat

lcp:
Similar to x86_64, the .sbat section is aligned to 4096, so we should
include the aligned part in SizeOfRawData as objcopy does for x86_64.
For VirtualSize, _sbat_vsize is used to reflect the actually size of
sbat.

This also fixes a strange hash mismatching in openSUSE build service
when attaching signature to AArch64 EFI images from shim package.

### DIFF
--- a/elf_aarch64_efi.lds
+++ b/elf_aarch64_efi.lds
@@ -76,6 +76,8 @@ SECTIONS
     *(.sbat.*)
   }
   _esbat = .;
+  _sbat_vsize = . - _sbat;
+  . = ALIGN(4096);
   _sbat_size = . - _sbat;
   _alldata_size = . - _data;
 
--- a/elf_arm_efi.lds
+++ b/elf_arm_efi.lds
@@ -74,6 +74,8 @@ SECTIONS
     *(.sbat.*)
   }
   _esbat = .;
+  _sbat_vsize = . - _sbat;
+  . = ALIGN(4096);
   _sbat_size = . - _sbat;
   _alldata_size = . - _data;
 


lcp:
The gnu-efi part is in https://github.com/rhboot/gnu-efi/pull/1

martinezjavier (approved):
Makes sense.

vathpela:
> This also fixes a strange hash mismatching in openSUSE build service
> when attaching signature to AArch64 EFI images from shim package.

This sounds like your signer isn't padding correctly?  Are you using `sbsign` or `pesign`?  If `pesign`, are you using `-h` ?

lcp:
> > This also fixes a strange hash mismatching in openSUSE build service
> > when attaching signature to AArch64 EFI images from shim package.
> 
> This sounds like your signer isn't padding correctly? Are you using `sbsign` or `pesign`? If `pesign`, are you using `-h` ?

We use `pesign -h -P` to verify the hash. What's annoying is that I cannot reproduce the issue locally. Anyway, I'm also checking the pesign code in our build service and trying to find more clues.

lcp:
> > > This also fixes a strange hash mismatching in openSUSE build service
> > > when attaching signature to AArch64 EFI images from shim package.
> > 
> > 
> > This sounds like your signer isn't padding correctly? Are you using `sbsign` or `pesign`? If `pesign`, are you using `-h` ?
> 
> We use `pesign -h -P` to verify the hash. What's annoying is that I cannot reproduce the issue locally. Anyway, I'm also checking the pesign code in our build service and trying to find more clues.

It turned out that I overlooked some part of the signing script which didn't use `-P` so it caused the hash mismatching when building against the older repo with pesign 0.112.

Thanks for merging the patch. I also fixed the signing script to make sure the issue won't happen again.

Issue #337, opened 2021-03-19
pe.c generate_hash generates incorrect hash for fbx64.efi


xnox:
Whilst testing https://github.com/rhboot/shim/pull/315 I realized that it shim fails to verify fbx64.efi by hash.

It appears that the hashes of the fbx64.efi as calculated by `hash-to-efi-sig-list fbx64.efi /dev/zero` or `pesign -P -h -i fbx64.efi` disagree with the hash calculated by shim.

This leads me to believe that allow/deny of unsigned .EFI binaries is incorrect by shim.

https://people.canonical.com/~xnox/efi/

See sample shim, fbx64.efi, screenshot of the generated hashes, and the output of hashes calculated by the userspace tools.

Signing fbx64.efi with any key, results in the correct hash being calculated.

xnox:
It is ill defined how we should hash binaries. Ideally signed & unsigned binaries should produce the same authenticode, we should work on fixing toolchains to produce such binaries. Status quo, is best for now. Of just hashing things as is.

PR #338, opened 2021-03-19
Compute correct PE hashes for unsigned binaries

chrisccoulson:
After hashing section contents, shim hashes extra data up until the
signature for a signed binary, but it ignores this extra data if the
binary is unsigned, calculating an incorrect PE hash. This fixes it
to include the extra data for unsigned images.

Should fix https://github.com/rhboot/shim/issues/337

### DIFF
--- a/pe.c
+++ b/pe.c
@@ -517,13 +517,16 @@ generate_hash(char *data, unsigned int datasize_in,
 		SumOfBytesHashed += Section->SizeOfRawData;
 	}
 
-	/* Hash all remaining data up to SecDir if SecDir->Size is not 0 */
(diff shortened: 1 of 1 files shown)


vathpela (changes requested):
I don't think we can do this, this way, for this release:
- it still needs to handle padding up to the alignment for the signature DD entry to make it match the digest on signed images
- it'll break all existing MokList/MokListX entries for these binaries
- I suspect we need to make it give us a _list_ of hashes to check, so we can cover all our bases in MokListX :/

frozencemetery:
Closing since this has been untouched for a year.

PR #342, opened 2021-03-25
arm/aa64: Swizzle some sections to make old sbsign happier.

vathpela:
xnox reports that with some versions of sbsign/sbcheck, it gets very
unhappy with non-contiguous sections and gaps between sections, which we
currently produce on targets with hand-coded headers.  This is all wrong
behavior from sbsigntools, and has been fixed in newer versions, but
nevertheless it's not hard for us to avoid.

This patch re-arranges the sections so there are no gaps, by padding the
file-size of .data and .sbat up to the full page, moving .sbat to be
before .vendor_cert, and moving .vendor_cert and .rela out of the range
covered by _edata, while still leaving included in the calculation of
SizeOfInitializedData.

### DIFF
--- a/elf_aarch64_efi.lds
+++ b/elf_aarch64_efi.lds
@@ -48,16 +48,38 @@ SECTIONS
    *(.dynbss)
    *(.bss)
    *(COMMON)
-   . = ALIGN(16);
+   . = ALIGN(4096);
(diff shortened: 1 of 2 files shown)


xnox (approved):
The title of the pull request alone deserves instant merge.

PR #346, opened 2021-03-27
tests for sbat variable preservation logic

jsetje:


vathpela:
This appears to have been subsumed into https://github.com/rhboot/shim/pull/348, so I'm gonna close it.  Re-open if I'm wrong.

PR #348, opened 2021-03-27
Correctly preserve the SBAT varaible if the signature matches and change the name to SbatLevel

jsetje:


jsetje:
This includes fixing the preservation logic, adding a test for it, and re-naming the SbatLevel UEFI variable.

vathpela:
I've pushed this with a couple of very minor changes:
- split the move to `preserve_sbat_uefi_variable()` out as its own commit and run clang-format on it
- replaced the places you have:
```
#ifdef ENABLE_SHIM_DEVEL
	UINT32 attributes = UEFI_VAR_NV_BS_RT;
#else
	UINT32 attributes = UEFI_VAR_NV_BS;
#endif
```
 with
```
	UINT32 attributes = SBAT_VAR_ATTRS;
```
- changed `preserve_sbat_uefi_variable()` to use `!` instead of `==0`, just because clang-format wrapping it makes the latter harder to read, and this avoids the wrapping.
- added some more info and your S-o-B to the commit messages

PR #349, opened 2021-03-27
Correctly preserve the SBAT varaible if the signature matches and change the name to SbatLevel

jsetje:
This is the simpler diff without the added testing.


jsetje:
This is the simpler actual fix with no testing

vathpela:
Yeah, as you guessed, I'd rather have the other one :)

PR #358, opened 2021-04-06
fallback: Handle GetNextVariableName returning EFI_BUFFER_TOO_SMALL

jprvita:
Now that fallback lists all EFI variables instead of only `Boot????` variables, a 256 chars buffer may not be large enough for the variable's name. This PR changes it to a dynamic buffer that starts with 256 chars, but gets extended if needed. It also adds extra logs when `GetNextVariableName` returns an error.

This is a follow-up to #312 -- these changes were mixed with other downstream commits and were left behind.

### DIFF
--- a/fallback.c
+++ b/fallback.c
@@ -415,9 +415,12 @@ find_boot_option(EFI_DEVICE_PATH *dp, EFI_DEVICE_PATH *fulldp,
 	cursor += DevicePathSize(dp);
 	StrCpy((CHAR16 *)cursor, arguments);
 
-	CHAR16 varname[256];
 	EFI_STATUS efi_status;
 	EFI_GUID vendor_guid = NullGuid;
+	UINTN buffer_size = 256 * sizeof(CHAR16);
+	CHAR16 *varname = AllocateZeroPool(buffer_size);
+	if (!varname)
+		return EFI_OUT_OF_RESOURCES;
 
 	UINTN max_candidate_size = calc_masked_boot_option_size(size);
 	CHAR8 *candidate = AllocateZeroPool(max_candidate_size);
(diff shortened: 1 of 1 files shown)


jprvita:
For-pushed rebase on latest main.

jprvita:
Force-pushed rebase on latest main.

jprvita:
Force-pushed rebase on latest main.

fallback.c
```diff
@@ -431,8 +431,26 @@ find_boot_option(EFI_DEVICE_PATH *dp, EFI_DEVICE_PATH *fulldp,
 		UINTN varname_size = sizeof(varname);
 		efi_status = gRT->GetNextVariableName(&varname_size, varname,
 						      &vendor_guid);
-		if (EFI_ERROR(efi_status))
+		if (EFI_ERROR(efi_status)) {
+			if (efi_status == EFI_BUFFER_TOO_SMALL)
+				VerbosePrint(L"Buffer too small for next var "
+					     L"name\n");
```
vathpela: Don't break the string across lines here.

fallback.c
```diff
@@ -431,8 +431,26 @@ find_boot_option(EFI_DEVICE_PATH *dp, EFI_DEVICE_PATH *fulldp,
...
+
+			if (efi_status == EFI_DEVICE_ERROR)
+				VerbosePrint(L"The next variable name could "
+					     L"not be retrieved due to a "
+					     L"hardware error\n");
```
vathpela, line 454: Don't break the string across lines here.

fallback.c
```diff
@@ -431,8 +431,26 @@ find_boot_option(EFI_DEVICE_PATH *dp, EFI_DEVICE_PATH *fulldp,
...
+
+			if (efi_status == EFI_INVALID_PARAMETER)
+				VerbosePrint(L"Invalid parameter to "
+					     L"GetNextVariableName: "
+					     L"varname_size=%d, varname=%s\n",
```
vathpela, line 459: Don't break the string across lines here.

fallback.c
```diff
@@ -426,15 +429,24 @@ find_boot_option(EFI_DEVICE_PATH *dp, EFI_DEVICE_PATH *fulldp,
...
+		UINTN varname_size = buffer_size;
 		efi_status = gRT->GetNextVariableName(&varname_size, varname,
 						      &vendor_guid);
 		if (EFI_ERROR(efi_status)) {
-			if (efi_status == EFI_BUFFER_TOO_SMALL)
+			if (efi_status == EFI_BUFFER_TOO_SMALL) {
 				VerbosePrint(L"Buffer too small for next var "
-					     L"name\n");
+					     L"name, re-allocating it to be %d"
+					     L" chars long and retrying\n",
```
vathpela, line 440: Don't break the string across lines here (even though it was incorrectly broken before.)

vathpela (changes requested):
A couple of minor nits, but this basically looks good to me.

vathpela:
I've pushed this as 437caecbab1 and 77648c5c01c .

PR #359, opened 2021-04-06
fallback: add compile option FALLBACK_NONINTERACTIVE

xnox:
In the cloud, all boots are non-interactive with keyboard and console
access either typically not available or prohibited. Also clouds
always do firstboot via fallback. This currently results in an
unacceptable 5s boot delay whilst fallback offers interactive reset
options that cannot be actioned.

In Ubuntu, we'd like to make fallback noninteractive by default
without any boot delays, due to bootspeed impact on firstboot of the
preinstalled images.

Bug-Ubuntu: https://bugs.launchpad.net/ubuntu/+source/shim/+bug/1922581

Signed-off-by: Dimitri John Ledkov <xnox@ubuntu.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -61,6 +61,10 @@ ifneq ($(origin FALLBACK_VERBOSE), undefined)
 	CFLAGS += -DFALLBACK_VERBOSE
 endif
 
+ifneq ($(origin FALLBACK_NONINTERACTIVE), undefined)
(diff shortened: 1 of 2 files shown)


jprvita:
Looks sensible to me.

josephlr:
I'm on the GCE team that's dealt with this, can confirm it saves ~5 seconds on the first boot of a newer GCE instance

PR #361, opened 2021-04-08
mok: allocate MOK config table as BootServicesData

lcp:
Linux kernel is picky when reserving the memory for x86 and it only
expects BootServicesData:

https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/arch/x86/platform/efi/quirks.c?h=v5.11#n254

Otherwise, the following error would show during system boot:

Apr 07 12:31:56.743925 localhost kernel: efi: Failed to lookup EFI memory descriptor for 0x000000003dcf8000

Although BootServicesData would be reclaimed after ExitBootService(),
linux kernel reserves MOK config table when it detects the existence of
the table, so it's fine to allocate the table as BootServicesData.

### DIFF
--- a/mok.c
+++ b/mok.c
@@ -1002,7 +1002,7 @@ EFI_STATUS import_mok_state(EFI_HANDLE image_handle)
 		npages = ALIGN_VALUE(config_sz, PAGE_SIZE) >> EFI_PAGE_SHIFT;
 		config_table = NULL;
 		efi_status = gBS->AllocatePages(AllocateAnyPages,
-						EfiRuntimeServicesData,
+						EfiBootServicesData,
 						npages,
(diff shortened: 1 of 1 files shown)


cmurf:
Downstream bug report
https://bugzilla.redhat.com/show_bug.cgi?id=1948432

lcp:
The kernel warning seems to be a bug in the mokvar driver in Linux kernel:
https://lkml.org/lkml/2021/6/30/179

Per UEFI Spec, InstallConfigurationTable() expects the list allocated in EfiRuntimeServiceData. We can revert the allocation type back to EfiRuntimeServiceData once the patch lands Linux kernel.

PR #362, opened 2021-04-09
Don't set user_insecure_mode and ignore_db in import_one_mok_state

AdamWill:
This seems completely incorrect and unnecessary, unless I'm
missing something. We already set them both to 0 at the start of
`import_mok_state`, which is the only thing that uses
`import_one_mok_state`, so it's unnecessary. It's incorrect
because it means those variables will be set to 0 even when they
should be set to 1 - even if they are momentarily set to 1 when
`import_one_mok_state` is called on the relevant variable, they
immediately get set back to 0 when it's called on the *next*
variable.

Signed-off-by: Adam Williamson <awilliam@redhat.com>

### DIFF
--- a/mok.c
+++ b/mok.c
@@ -888,9 +888,6 @@ EFI_STATUS import_one_mok_state(struct mok_state_variable *v,
 	EFI_STATUS ret = EFI_SUCCESS;
 	EFI_STATUS efi_status;
 
-	user_insecure_mode = 0;
-	ignore_db = 0;
-
 	UINT32 attrs = 0;
 	BOOLEAN delete = FALSE;
 


AdamWill:
Note I am absolutely not a C or SB or EFI expert and this may be wildly and terribly wrong, please check it carefully. But it's my guess as to what causes the problem I hit with 15.4 (my system doesn't boot with SB enabled; with 15 it boots after briefly showing a "Booting in insecure mode" message) and what might solve it. See https://bodhi.fedoraproject.org/updates/FEDORA-2021-cab258a413 . I don't think I can test this without figuring out how to do key enrolment and stuff, because I can't do a properly signed build of shim...

martinezjavier:
@AdamWill yes, I think you are correct and your patch will fix the issue. I was also able to reproduce the issue with Qemu and OVMF, basically did the following:

```
1) Enable SB support in firmware settings
2) Run mokutil --disable-validation and reboot
3) shim 15 boots correctly with SB disabled
4) Update to shim shim 15.4 and reboot
5) shim 15.4 fails to boot and prints this message:
      Bootloader has not verified loaded image.
      System is compromised.  halting.
```

aburmash:
Can confirm, issue is not Fedora specific and is clearly caused by the code piece reported. With validation disabled shim still tries to validate the binaries it boots.

vathpela:
I've pushed this to main as 822d07ad4f07ef66fe447a130e1027c88d02a394.

AdamWill:
thanks for improving the commit message!

steve-mcintyre:
Problem and fix tested and confirmed locally for me too

PR #363, opened 2021-04-12
Allow modifications of overridden ARCH_LDFLAGS

gonzoua:
Add "override" directive so ARCH_LDFLAGS that is passed as
an argument to make(1) can be modified by += operator for
ARM and ARM64 platforms. Without this directive only passed
value is used and --defsym.. part is not passed to ld(1).

Signed-off-by: Oleksandr Tymoshenko <ovt@google.com>

### DIFF
--- a/Make.defaults
+++ b/Make.defaults
@@ -84,7 +84,7 @@ ifeq ($(ARCH),aarch64)
 	ARCH_SUFFIX_UPPER	?= AA64
 	FORMAT			:= -O binary
 	SUBSYSTEM		:= 0xa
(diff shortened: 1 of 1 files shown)


gonzoua:
Hello,

Gentle reminder about this MR. Any additional information I can provide to speed up the review?

vathpela:
This seems weird?  Why not just pass additional options in with CFLAGS, like normal?

frozencemetery (changes requested):
(Marking reviewed from comment above.)

Issue #249, opened 2021-01-18
dmidecode fails when booted with 1f123ac2359cd923e9144f944a4bddf597fddbb5


jyong2:
When booted with shim, dmidecode fails to run:

```
# dmidecode -t baseboard
# dmidecode 3.2
# SMBIOS entry point at 0x748cd000
Found SMBIOS entry point in EFI, reading table from /dev/mem.
# No SMBIOS nor DMI entry point found, sorry.
```

Any ideas about this?

Edit: formatting

jyong2:
Bisect seems to be suggesting fecc2dfb8e408526221091923d9345796b8e294e, tested on qemu 5.1.0.

jsetje:
With secure boot enabled, lockdown prevents reads from /dev/mem. Newer dmidecode should be using sysfs instead.

Can you help us establish if you're seeing this change in behavior with or without secure boot enabled?

If you're seeing it with secure boot enabled, then the /dev/mem failure is expected and prior to that fix your kernel was failing to enforce lockdown.

I you're seeing it with secure boot disabled, then something else may have changed, and we should look into this further.

jyong2:
dmidecode does use sysfs when possible, it works fine with the older shim, but the dmi tables entries are missing in the newer shim, causing it to fallback to /dev/mem. This happens regardless of secureboot or not.

I should also mention I am using ovmf edk2-stable201911 as part of Yocto Project.

jyong2:
As of 018b74d2d69ef35b43b79709f2ea60325f12dde2 commit, I am still getting errors:
```
root@intel-corei7-64:~# dmidecode -t baseboard
# dmidecode 3.2
Getting SMBIOS data from sysfs.
SMBIOS 2.8 present.

Wrong DMI structures length: 383 bytes announced, only 16 bytes available.
Invalid entry length (0). DMI table is broken! Stop.

```

jyong2:
Still same for  39b96c01bfd4547f38c9e573ff5d551057ea272c
```
root@intel-corei7-64:~# dmidecode
# dmidecode 3.2
Getting SMBIOS data from sysfs.
SMBIOS 2.8 present.
9 structures occupying 383 bytes.
Table at 0x1FBCF000.

Wrong DMI structures length: 383 bytes announced, only 16 bytes available.
Invalid entry length (0). DMI table is broken! Stop.

root@intel-corei7-64:~# dmidecode -t baseboard
# dmidecode 3.2
[...]

PR #365, opened 2021-04-19
mok: fix potential buffer overrun in import_mok_state

jyong2:
Fix the case where data_size is 0, so config_template is
not implicitly copied like the size calculation above.

upstream-status: https://github.com/rhboot/shim/issues/249

Signed-off-by: Jonathan Yong <jonathan.yong@intel.com>


pcmoore:
Thanks @jyong2, this fixed a problem I was seeing with my QEMU+OVMF dev/test VM.

PR #367, opened 2021-04-23
lds: remove note.gnu.build-id from most places, move to the end.

### LINKED ISSUE #366: Significant problems with Aarch64 (and Arm?) builds
For x86 platforms, we use support in binutils for making EFI apps (objcopy ... --target efi-app-x86_64 or --target efi-app-ia32) and that seems to work OK. For Arm-based platforms, we don't have such support.

Instead, we have some locally-crafted CRT and linker script hacks to try and make things work. Pre-SBAT this seemed to work, but since then I've been seeing significant issues in this area.

1. Older versions of binutils will not create an aarch64 shim that works with Secure Boot enabled. (I've bisected binutils to find that the specific commit that makes things "work" is https://sourceware.org/git/?p=binutils-gdb.git;a=commit;h=8c803a2dd7d3d742a3d0071914f557ef465afe71 . Older builds (before release 2.35, basically) will cause a Synchronous Exception during early boot of shim.

2. Even using a later version of binutils, I've seen another worrying problem. We normally expect shim builds to be reproducible, and this is important. As part of the release process for Debian's shim, I build and test shim locally. I then compare checksums to ensure that exactly the same binaries are produced by Debian's build machines. Building the same source (Debian's 15.4-1 release) on my aarch64 development machine (Cortex-A72 Macchiatobin) consistently gives the same output. However, building that exact same release using the same build setup (gcc, binutils, etc.) on other machines gives me different results. On that 15.4-1 release, I tested found that 5 different aarch64 machines repeatably gave me 4 different binaries.

As a result of these issues, I have decided to abandon Secure Boot support on Debian's arm64 port. Until we have proper reliable toolchain support for what we're building, I don't think work on this platform is sustainable.

xnox:
.note.gnu.build-id was included in rodata sections on some
architectures, which get added to the elf .so & debug .so, but are not
copied into pe binary. Unless of course this section was embeeded
inside the .rodata.

Remove embedded copies of build-id, move build-id to the end of the
elf binaries. Note that stability and reproducibility of build-id note
is not needed, as most distributions strip & replace build-id with a
rebuilt one based on package name & version.

Fixes #366 (point 2)

Signed-off-by: Dimitri John Ledkov <xnox@ubuntu.com>

### DIFF
--- a/elf_aarch64_efi.lds
+++ b/elf_aarch64_efi.lds
@@ -80,8 +80,6 @@ SECTIONS
     *(.srodata)
     *(.dynsym)
     *(.dynstr)
-    . = ALIGN(16);
-    *(.note.gnu.build-id)
     . = ALIGN(4096);
     *(.vendor_cert)
     *(.data.ident)
@@ -100,4 +98,5 @@ SECTIONS
     *(.note.GNU-stack)
   }
   .comment 0 : { *(.comment) }
+  .note.gnu.build-id : { *(.note.gnu.build-id) }
 }
--- a/elf_arm_efi.lds
+++ b/elf_arm_efi.lds
@@ -80,8 +80,6 @@ SECTIONS
     *(.srodata)
     *(.dynsym)
     *(.dynstr)
-    . = ALIGN(16);
-    *(.note.gnu.build-id)
     . = ALIGN(4096);
     *(.vendor_cert)
     *(.data.ident)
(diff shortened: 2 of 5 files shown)


xnox:
Settings this to draft, as that's just a concept. Did not compiler or test boot this yet on all the arches.

In theory this change should not result in change of the pe binaries, or make them reproducible.

frozencemetery:
Closing this since it's been untouched for a year.

xnox:
This is still needed, because without this change the builds of shim end up being non-reproducible which harms shim review.

frozencemetery:
That's fine, but it sat in draft for a year without changes.  Per your comment: "Settings this to draft, as that's just a concept. Did not compiler or test boot this yet on all the arches.", so feel free to reopen if that changes or you need to run against CI.

PR #369, opened 2021-05-06
mok: relax the maximum variable size check

lcp:
Some UEFI environment such as u-boot doesn't implement
QueryVariableInfo(), so we couldn't rely on the function to estimate the
available space for RT variables. All we can do is to call SetVariable()
directly and check the return value of SetVariable().

### DIFF
--- a/mok.c
+++ b/mok.c
@@ -351,13 +351,18 @@ mirror_mok_db(CHAR16 *name, CHAR8 *name8, EFI_GUID *guid, UINT32 attrs,
 	SIZE_T max_var_sz;
 
 	efi_status = get_max_var_sz(attrs, &max_var_sz);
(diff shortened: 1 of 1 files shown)


julian-klode:
It's worth pointing out that some firmware implementations (Apple 2.40) even lie and tell you they have space but then die/hang when you try to write :D

lcp:
We had a similar case that the firmware lied about the space, but the firmware _kindly_ returned EFI_OUT_OF_RESOURCES :-p

https://github.com/rhboot/shim/pull/372

julian-klode:
+1 on merging

We got hit by this too https://bugs.launchpad.net/ubuntu/+source/shim/+bug/1934780 and verified this fixes it. Thanks!

julian-klode (approved):

PR #374, opened 2021-05-19
BUG/RFE: more set_second_stage() fixes and improvements

pcmoore:
*NOTE: This PR is against `shim-15.4` since that is what we are concerned with, but it should also apply to the default branch.*

This PR has two patches: the first fixes a regression in the shim argument parsing, the second adds the ability to handle a single UCS-2 string passed as an argument.  Both patches have been tested on modern QEMU+OVMF and Cisco UCS systems (physical hardware).  Please see each patch description for more information.

### DIFF
--- a/shim.c
+++ b/shim.c
@@ -1520,7 +1520,7 @@ EFI_STATUS set_second_stage (EFI_HANDLE image_handle)
 	 * case.
 	 */
 	if (strings == 1) {
-		UINT16 *cur = start = li->LoadOptions;
+		UINT16 *cur = li->LoadOptions;
 
 		/* replace L' ' with L'\0' if we find any */
 		for (i = 0; i < li->LoadOptionsSize / 2; i++) {
@@ -1534,7 +1534,8 @@ EFI_STATUS set_second_stage (EFI_HANDLE image_handle)
 	}
 
 	/*
-	 * If it's not string data, try it as an EFI_LOAD_OPTION.
+	 * Try to parse the LoadOptions as either a EFI_LOAD_OPTION or as
+	 * a variable number of UCS-2 strings.  Wish us luck.
 	 */
 	if (strings == 0) {
 		/*
@@ -1548,6 +1549,24 @@ EFI_STATUS set_second_stage (EFI_HANDLE image_handle)
 		if (EFI_ERROR(efi_status))
 			return EFI_SUCCESS;
 
+		remaining_size = 0;
+	} else if (strings == 1) {
+		start = li->LoadOptions;
+		if (is_our_path(li, start)) {
+			/*
+			 * I found a version of BDS that gives us our own path
+			 * in LoadOptions:
+77162C58                           5c 00 45 00 46 00 49 00          |\.E.F.I.|
+77162C60  5c 00 42 00 4f 00 4f 00  54 00 5c 00 42 00 4f 00  |\.B.O.O.T.\.B.O.|
+77162C70  4f 00 54 00 58 00 36 00  34 00 2e 00 45 00 46 00  |O.T.X.6.4...E.F.|
+77162C80  49 00 00 00                                       |I...|
+			 * which is just cruel... So yeah, just don't use it.
+			 */
+			return EFI_SUCCESS;
+		}
+
+		/* assume this string is the second stage loader */
+		loader_len = li->LoadOptionsSize;
 		remaining_size = 0;
 	} else if (strings >= 2) {
 		/*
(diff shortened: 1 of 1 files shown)


pcmoore:
I just did a force push to fix a typo in the example code provided in one of the commit descriptions.  There were no code changes in the force push.

julian-klode:
Is `loader_len` in `else` off by one? If our load options are `A\0B\0C...`, start will be +2, and the else part is taken at C (+4), so `(cur[i]-start)==2` long characters, but is it correct to include the \0 in the count?

It also reads a bit confunsingly, it's not clear the else part always happens after the if part and splits arguments off the loader. would be great to add some comments there, or splitting the loop into two again such that the search for the loader end is clearly identifiable.

julian-klode:
Kind of wanting an echo.efi to test this, but not sure if shim argument parsing from EFI shell even works these days.

julian-klode:
I applied the first patch and launched shim from EFI shell:

- `shimx64.efi fwupdx64.efi` works
- `shimx64.efi shimx64.efi fwupdx64.efi` tries to load grubx64.efi
- `shimx64.efi shimx64.efi` is unhappy:
```
FS0:\> shimx64.efi shimx64.efi
Failed to allocate loader string
set_second_stage() failed: Out of Resources
Something has gone seriously wrong: shim_init() failed: Out of Resources
```

julian-klode:
FWIW, 15.4 has been released already, so you want to target the main branch.

julian-klode:
Changing the cur[i] to cur[i-1] did not help my issues. Another interesting bit is that if I pass two spaces between shimx64.efi and argument, it tries to load grub :| EFI shell weird.

pcmoore:
Yes, something is a bit odd ... I double promise - :) - it was working when I was doing the development, but as I started preparing a production build last week with all the necessary certs, etc. it started failing for me with similar problems.  I'm debugging this now and will report back when I figure out the root cause and a solution.

Thanks for the review, I apologize for the noise!

julian-klode:
@pcmoore See #379 for another approach, it doesn't just fix bugs, but actually becomes more readable :)

julian-klode:
I've also checked if this helps with loading fwupdx64.efi and it does not.

```
is_our_path() dppath: \EFI\ubuntu\shimx64.efi
is_our_path() path: \fwupdx64.efi
load_image() attempting to load \EFI\ubuntu\grubx64.efi
```

So, it fixed the regression in the shell to some extend, but clearly not the one that matters to most users.

pcmoore:
> I've also checked if this helps with loading fwupdx64.efi and it does not.

When you say "this", are you talking about PR #379 ?

FWIW, I found the problem with this PR yesterday ... long story short, there was some debug code that affected the `strings >= 2` case` and when the debug code was removed it broke that case (the most likely EFI shell case).  This branch has the fix:

* https://github.com/pcmoore/misc-rhboot_shim/tree/gh-pr-12

For those who are curious, here is the fixed patch's code snippet:
```patch
diff --git a/shim.c b/shim.c
index c5cfbb83..c0d9241e 100644
--- a/shim.c
+++ b/shim.c
@@ -1555,16 +1555,24 @@ EFI_STATUS set_second_stage (EFI_HANDLE image_handle)
                 * LoadOptions.  We ignore the string before the first L'\0',
                 * i.e. the name of this program.
                 */
-               UINT16 *cur = li->LoadOptions;
-               for (i = 1; i < li->LoadOptionsSize / 2; i++) {
+               UINT16 *cur;
+               start = cur = li->LoadOptions;
+               for (i = 1; i < (li->LoadOptionsSize / 2) - 1; i++) {
                        if (cur[i - 1] == L'\0') {
-                               start = &cur[i];
-                               remaining_size = li->LoadOptionsSize - (i * 2);
-                               break;
+                               if (start == li->LoadOptions) {
+                                       start = &cur[i];
+                                       loader_len = li->LoadOptionsSize - (i *>
+                                       remaining_size = 0;
+                               } else {
+                                       loader_len = (&cur[i] - start) * 2;
+                                       remaining_size = li->LoadOptionsSize - >
+                                       break;
+                               }
                        }
                }
-
-               remaining_size -= i * 2 + 2;
+               /* if we didn't find at least one NULL, something is wrong */
+               if (start == li->LoadOptions)
+                       return EFI_SUCCESS;
        } else if (strings == 1 && is_our_path(li, start)) {
                /*
                 * And then I found a version of BDS that gives us our own path
```

... the important change is that we properly initialize the `start` to `li->LoadOptions` along with `cur`.  I also changed the limit of the for loop to `(li->LoadOptions / 2) - 1` as we can assume an option string has to be *at* *least* one `CHAR16` character long; this is more of a minor optimization and shouldn't have any real effect.

I currently don't have my QEMU+OVMF+SWTPM test setup working with SBAT so I need to finish setting that up before I can fully test this in UEFI SB mode and reissue the PR, but if someone wants to give the branch above a quick test that would be helpful.

julian-klode:
> When you say "this", are you talking about PR #379 

No, I'd have written that (literally that, or the PR link again) then. I did mean this PR (#374); well the first commit only, the second one did not mention anything about regressions. Though, arguably it is possible that's a regression fix too, fixing commit 937afbe9e63fa88d80b10874d682bf30776f4e71.

I also submitted #380 as a simple revert and fix of the NULL condition - it really was just missing the initialization of start there, but oh well, did not help me either, so the commit fixed here did not actually break fwupd loading, which has this booth entry.

```
HD(1,GPT,a155bb6c-2d31-4a90-8678-f0f921e11e9b,0x800,0x100000)/File(\EFI\ubuntu\shimx64.efi)\.f.w.u.p.d.x.6.4...e.f.i...
```

It's all funny. I believe #379 is the best approach to load option handling, it gets rid of all the hackery and makes things work correctly in a lot more places (e.g. replacing all spaces with \0 is invalid if you like do `shimx64.efi linux.efi option1 option2` - the kernel actually needs spaces). It's fairly radical though.

pcmoore:
> > When you say "this", are you talking about PR #379
> 
> No, I'd have written that (literally that, or the PR link again) then. I did mean this PR (#374); well the first commit only, the second one did not mention anything about regressions.

It wasn't clear from the context of this discussion and I just wanted to be sure.

> I also submitted #380 ...

I really don't care what approach the maintainers want to take, I just want shim to work for my kernel-as-a-second-stage-loader use case and I'm happy to do whatever I can to help make that happen :)

pcmoore:
FWIW, I updated my patch to preserve the spaces in the LoadOption string but lost some time this afternoon messing around with some of the SBAT stuff.  It looks like everything is working now but I want to make one more pass through the patch before posting and that is going to need to wait until tomorrow.

pcmoore:
I just submitted PR #381 as discussed above, closing this PR.

PR #375, opened 2021-05-21
Post-process our PE to be sure.

vathpela:
On some versions of binutils, including binutils-2.23.52.0.1-55.el7, do not correctly initialize the data when computing the PE optional header checksum.  Unfortunately, this means that any time you get a build that reproduces correctly using the version of objcopy from those versions, it's just a matter of luck.

This patch series introduces a new utility program, post-process-pe, which does some basic validation of the resulting binaries, and if necessary, performs some minor repairs:

- sets the timestamp to 0 (this was previously done with dd using constant offsets that aren't really safe.)
- re-computes the checksum.

In the near future, I intend to refactor the duplicated code this adds, as well as making this a general purpose efi-pe-lint program (which we've clearly needed for some time now), but we need this subset of the functionality sooner rather than later.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,6 +29,7 @@ Make.local
 /build*/
 /certdb/
 /cov-int/
+/post-process-pe
 /random.bin
 /sbat.*.csv
 /scan-results/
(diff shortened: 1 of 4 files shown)


jsetje (approved):

vathpela:
I don't know why github won't believe it pushed this to main, but it really did, so I'm closing it.

PR #377, opened 2021-06-04
shim: don't fail on the odd LoadOptions length

lcp:
Some firmware feeds the LoadOptions with an odd length when booting from
an USB device(*). We should only skip this kind of LoadOptions, not fail
it, or the user won't be able to boot the system from USB or CD-ROM.

(*) https://bugzilla.suse.com/show_bug.cgi?id=1185232#c62


vathpela:
I've pushed this as ada7ff69bd8a95614aa63dd618764daf428f235d.

PR #378, opened 2021-06-05
Don't unhook ExitBootServices when EBS protection is disabled

sforshee:
When EBS protection is disabled unhook_system_services() still
"restores" EBS in the boot services table. However, EBS was not
replaced and the original value not saved, so NULL is written. Then
when the OS tries to call EBS it ends up jumping to NULL.

Fix this by also compiling out the code to unhook EBS when
DISABLE_EBS_PROTECTION is defined.

Fixes: 4b0a61dc9a95 ("shim: compile time option to bypass the ExitBootServices() check")
Signed-off-by: Seth Forshee <seth.forshee@canonical.com>

### DIFF
--- a/replacements.c
+++ b/replacements.c
@@ -33,7 +33,9 @@ get_active_systab(void)
 static typeof(systab->BootServices->LoadImage) system_load_image;
 static typeof(systab->BootServices->StartImage) system_start_image;
 static typeof(systab->BootServices->Exit) system_exit;
+#if !defined(DISABLE_EBS_PROTECTION)
 static typeof(systab->BootServices->ExitBootServices) system_exit_boot_services;
+#endif /* !defined(DISABLE_EBS_PROTECTION) */
 
 static EFI_HANDLE last_loaded_image;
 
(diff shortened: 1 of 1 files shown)


julian-klode (approved):
LGTM. Love how it adds a safety check by ifdefing the variable.

xnox (commented):
Is there one more place similar to this that is not taken care of in hook_exit() and unhook_exit()?

It looks like hook_exit replaces system_exit, but unhook_exit never restores the original one when `DISABLE_EBS_PROTENCTION` is on. It feels like unhook_exit should be unhooking things even with DISABLE_EBS_PROTECTION on.

sforshee:
> It looks like hook_exit replaces system_exit, but unhook_exit never restores the original one when `DISABLE_EBS_PROTENCTION` is on. It feels like unhook_exit should be unhooking things even with DISABLE_EBS_PROTECTION on.

Yeah, looking again you are probably right that the mistake was putting the ifdef around the code around unhooking Exit() instead of ExitBootServices(). I'll update the patch.

xnox (approved):

julian-klode (approved):
Launching `shimx64.efi` from EFI shell, and then exiting the grub it chainloads. With the change applied, it correctly exits back to the shell; before it was resetting the system.

PR #384, opened 2021-06-21
mok: skip the empty variables when copying the data to MOK config table

lcp:
When calculating the size of the MOK config table, we skip the empty
variables. However, when copying the data, we copied the zeroed config
templates for those empty variables, and this could cause crash since we
may write more data than the allocated pages. This commit skips the
empty variables when copying the data so that the size of copied data
matches config_sz.


lcp:
Just found that my patch is exactly the same as https://github.com/rhboot/shim/pull/365

Issue #386, opened 2021-06-30
Mirrored MOK variables could be accidentally deleted


lcp:
Per the comment of MOK_MIRROR_DELETE_FIRST, any MOK variable with the flag would delete the existing mirrored variable first.
https://github.com/rhboot/shim/blob/9f973e4e95b1136b8c98051dbbdb1773072cc998/mok.c#L127

The code to mirror a MOK variable is in mirror_mok_db():
https://github.com/rhboot/shim/blob/9f973e4e95b1136b8c98051dbbdb1773072cc998/mok.c#L373-L379
It writes the variable only when `only_first` is true.

However, in maybe_mirror_one_mok_variable(), 
https://github.com/rhboot/shim/blob/9f973e4e95b1136b8c98051dbbdb1773072cc998/mok.c#L866-L870
LibDeleteVariable() is called to delete the mirrored MOK variable when `only_first` is false.

The logic looks strange to me since it tries to delete the mirrored MOK variable we just created. The problem doesn't show in QEMU/OVMF VMs because LibDeleteVariable() deletes the variables with BS+RT+NV but we create the mirrored variable with BS+RT, so OVMF just returns "EFI_INVALID_PARAMETER". However, for the VMWare VMs, the firmware doesn't check the attributes and just deletes the variable matching the given name+guid, so those RT variables never show in OS.

PR #387, opened 2021-07-01
mok: delete the existing RT variables only when only_first=TRUE

lcp:
For the firmware without the variable writing issues, MOK variables are
mirrored when only_first=TRUE. However, LibDeleteVariable() was called
in maybe_mirror_one_mok_variable() when only_first=FALSE, and this
could delete MOK variables that were just mirrored in the first round.

This bug was hidden since LibDeleteVariable() deletes BS+RT+NV variables
while we mirror MOK variables as BS+RT, and the firmware refused to
delete the mirrored MOK variable due to mismatching attributes. However,
some firmwares, such as VMWare, didn't enforce the attribute check and
just deleted the variables with matched name and GUID. In such system,
MokListRT was always removed before it reached OS.

Fixes: https://github.com/rhboot/shim/issues/386

### DIFF
--- a/mok.c
+++ b/mok.c
@@ -863,7 +863,7 @@ maybe_mirror_one_mok_variable(struct mok_state_variable *v,
 	BOOLEAN present = FALSE;
 
 	if (v->rtname) {
-		if (!only_first && (v->flags & MOK_MIRROR_DELETE_FIRST)) {
+		if (only_first && (v->flags & MOK_MIRROR_DELETE_FIRST)) {
 			dprint(L"deleting \"%s\"\n", v->rtname);
 			efi_status = LibDeleteVariable(v->rtname, v->guid);
 			dprint(L"LibDeleteVariable(\"%s\",...) => %r\n", v->rtname, efi_status);


julian-klode:
I have verified this causes no regressions on OVMF and Lenovo ThinkPad T480s. No VMWare keys yet :(

vathpela:
> This bug was hidden since LibDeleteVariable() deletes BS+RT+NV variables while we mirror MOK variables as BS+RT, and the firmware refused to delete the mirrored MOK variable due to mismatching attributes. However, some firmwares, such as VMWare, didn't enforce the attribute check and just deleted the variables with matched name and GUID. In such system, MokListRT was always removed before it reached OS.

I think this patch is correct, but I also think that means our deletion path could be better.  Do you know if these systems (either/both types) will delete variables with `SetVariable(L"foo", &guid, 0, 0, NULL)`, as opposed to with the attributes the variable was set with?  As I read the spec this should always work (Italics mine):

> The Attributes have the following usage rules:
> - If a preexisting variable is rewritten with different attributes, **SetVariable()** shall not modify
>   the variable and shall return **EFI_INVALID_PARAMETER**. The only exception to this is when
>   the only attribute differing is **EFI_VARIABLE_APPEND_WRITE**. In such cases the call's
>   successful outcome or not is determined by the actual value being written. There are two
>   exceptions to this rule:
>    - If a preexisting variable is rewritten with no access attributes specified, the variable will be
>      deleted.
>    - **EFI_VARIABLE_APPEND_WRITE** attribute presents a special case. It is acceptable to
>      rewrite the variable with or without **EFI_VARIABLE_APPEND_WRITE** attribute.
> - _Setting a data variable with no access attributes causes it to be deleted._
> - **EFI_VARIABLE_AUTHENTICATED_WRITE_ACCESS** is deprecated and should not be used.
>   Platforms should return **EFI_UNSUPPORTED** if a caller to **SetVariable()** specifies this
>   attribute.
> - Unless the **EFI_VARIABLE_APPEND_WRITE**,
>   **EFI_VARIABLE_TIME_BASED_AUTHENTICATED_WRITE_ACCESS**, or
>   **EFI_VARIABLE_ENHANCED_AUTHENTICATED_WRITE_ACCESS** attribute is set, setting a data
>  variable with zero DataSize specified, causes it to be deleted.
> _..._

If that's actually commonly implemented, that may be a better choice.

At any rate, https://github.com/rhboot/shim/pull/394 has test cases that show your fix does what it claims to.

lcp:
> > This bug was hidden since LibDeleteVariable() deletes BS+RT+NV variables while we mirror MOK variables as BS+RT, and the firmware refused to delete the mirrored MOK variable due to mismatching attributes. However, some firmwares, such as VMWare, didn't enforce the attribute check and just deleted the variables with matched name and GUID. In such system, MokListRT was always removed before it reached OS.
> 
> I think this patch is correct, but I also think that means our deletion path could be better. Do you know if these systems (either/both types) will delete variables with `SetVariable(L"foo", &guid, 0, 0, NULL)`, as opposed to with the attributes the variable was set with? As I read the spec this should always work (Italics mine):
> 
I totally agree that the deletion path should be amended. My patch is more like a quick fix due to the time pressure from customers.

> > The Attributes have the following usage rules:
> > 
> > * If a preexisting variable is rewritten with different attributes, **SetVariable()** shall not modify
> >   the variable and shall return **EFI_INVALID_PARAMETER**. The only exception to this is when
> >   the only attribute differing is **EFI_VARIABLE_APPEND_WRITE**. In such cases the call's
> >   successful outcome or not is determined by the actual value being written. There are two
> >   exceptions to this rule:
> >   
> >   * If a preexisting variable is rewritten with no access attributes specified, the variable will be
> >     deleted.
> >   * **EFI_VARIABLE_APPEND_WRITE** attribute presents a special case. It is acceptable to
> >     rewrite the variable with or without **EFI_VARIABLE_APPEND_WRITE** attribute.
> > * _Setting a data variable with no access attributes causes it to be deleted._
> > * **EFI_VARIABLE_AUTHENTICATED_WRITE_ACCESS** is deprecated and should not be used.
> >   Platforms should return **EFI_UNSUPPORTED** if a caller to **SetVariable()** specifies this
> >   attribute.
> > * Unless the **EFI_VARIABLE_APPEND_WRITE**,
> >   **EFI_VARIABLE_TIME_BASED_AUTHENTICATED_WRITE_ACCESS**, or
> >   **EFI_VARIABLE_ENHANCED_AUTHENTICATED_WRITE_ACCESS** attribute is set, setting a data
> >   variable with zero DataSize specified, causes it to be deleted.
> >   _..._
> 
> If that's actually commonly implemented, that may be a better choice.
> 
Yes, it'd be nice if shim detects the attributes before deleting the variables. BTW, I remember that we already removed all usage of EFI_VARIABLE_APPEND_WRITE to avoid problems in some firmwares, so we can ignore such cases.

> At any rate, #394 has test cases that show your fix does what it claims to.

Thanks!

vathpela:
I've pushed this as 34460fe0ad6463fd8a694b70c045f34df3b325a6.

PR #393, opened 2021-07-26
Fallback to default loader if parsed one does not exist

julian-klode:
If the specified second stage loader does not exist (invalid
parameter), fall back to the DEFAULT_LOADER. This avoids failing
the boot on any garbage that made it through the load option parser
as a second stage loader name.

Bug-Ubuntu: https://bugs.launchpad.net/bugs/1937115


julian-klode:
Arguably I have not tested this. I'm confused by the code a bit because it seems any remaining load_options from set_second_stage() remain unused (the variable is assigned at the end, but never read outside that function).

julian-klode:
Updated with changes for `load_options` and `EFI_NOT_FOUND`. I believe in the bug in question, `EFI_INVALID_PARAMETER` was returned because the given bytes were simply not valid in filenames, but can't be sure.

julian-klode:
Moved it out of the wrong if, added the msleep and a message, and tested in qemu.

julian-klode:
Also added code to dump the load options in debug mode

haobinnan:
Exception occurred after using patches 393-1 / 393-2 / 399-1 / 399-2. And I got following messages after replacing \EFI\Microsoft\Boot\bootmgfw.efi with shimx64.efi: 
Falled to open \EFI\Microsoft\Boot\S - Invalid Parameter
Failed to load image \EFI\Microsoft\Boot\S: Invalid Parameter
start_image() returned Invalid Parameter, falling back to default loader

@julian-klode

vathpela:
> Exception occurred after using patches 393-1 / 393-2 / 399-1 / 399-2. And I got following messages after replacing \EFI\Microsoft\Boot\bootmgfw.efi with shimx64.efi:
> Falled to open \EFI\Microsoft\Boot\S - Invalid Parameter
> Failed to load image \EFI\Microsoft\Boot\S: Invalid Parameter
> start_image() returned Invalid Parameter, falling back to default loader

This is pretty clearly not caused by this change, so I'm going to merge this and let you all discuss that issue in the thread for #399.

PR #394, opened 2021-07-26
Test mok mirror

vathpela:
This adds test cases for #387, which is also the second commit in the series.

Additionally, a memory leak from retrieving the same variable multiple times is fixed and a test for that problem is also added.


frozencemetery (commented):
I didn't see anything obviously wrong here.  One comment inline, and some feedback on commit messages:

- A few commit messages are missing DCO.
- Several commit messages have periods at the end of their subjects (unsure if this is important to you; feel free to ignore if it's not).
- "make gcc ignore its mistake even harder" captures the frustration but not what the issue is :)

PR #396, opened 2021-07-27
fallback: find_boot_option() needs to return the index for the boot entry in optnum

jsetje:
The CopyMem() calls in add_to_boot_list() expect that
find_boot_option() returned an index to the matching entry in the
BootOrder array. The previous code returned the numerical portion of
the boot entry label, which in some cases resulted in -1 *
sizeof(CHAR16) being passed to CopyMem() which would in turn corrupt
the running firmware resulting in an exception and a failure to boot
or reset.

### DIFF
--- a/fallback.c
+++ b/fallback.c
@@ -462,10 +462,15 @@ find_boot_option(EFI_DEVICE_PATH *dp, EFI_DEVICE_PATH *fulldp,
 			first_new_option_size = StrLen(arguments) * sizeof (CHAR16);
 		}
 
-		*optnum = xtoi(varname + 4);
-		FreePool(candidate);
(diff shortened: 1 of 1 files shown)


julian-klode (approved):
This all seems to make sense to me.

julian-klode:
@jsetje still marked draft, intentional?

jsetje:
This appears to do the right thing on the systems I currently have access to. I'm continuing to test.

jsetje:
> @jsetje still marked draft, intentional?

At the time yes, I hadn't tested adding a new device to a previously booted system yet, which seemed like a key case I should cover.

PR #398, opened 2021-07-30
Modify sbat.md to help with readability.

eshiman:
Signed-off-by: Esther Shimanovich eshimanovich@google.com

I've had a bit of trouble understanding SBAT.md, so I thought I might open up a PR with a few additions that might help with readability, and some questions. Hopefully this might be helpful to other people as well.

1) Can you define disclosure (from the chart included in the file)?
2) Where do the numbers 5 and 6 in the chart come from? I expected the fields under "after Vendor C's first Update" to be 2 and 3 instead of 5 and 6. Why the big jump from 1 to 5?
3) Why do you need both upstream's version of grub and downstream's version of grub in the sbat.csv file? Isn't only one used?
4) Will the product specific generation number continually increase for each global or do they reset? I was a little confused by this sentence: "The product-specific generation number does not reset and continues to monotonically increase over the course of these events." ----What events are you referencing to here?
5) Where can we look up global generation numbers? Or is that in flux still? This info would be helpful in documentation, even if there is one entry so far.

I highlighted some of the words that were defined in this doc, and I also added an example earlier on, as that would have been helpful for me to visualize some of the new concepts as I read it. I didn't want to make any structural changes so I avoided those. If you respond to my questions above, I am happy to add them to the doc in this PR!

Thanks a bunch!
Esther

### DIFF
--- a/SBAT.md
+++ b/SBAT.md
@@ -140,13 +140,21 @@ value associated with the data currently stored in that variable.
 
 #### Generation-Based Revocation Scenarios
 
-Products (**not** vendors, a vendor can have multiple products or even pass a
+**Products** (**not** vendors, a vendor can have multiple products or even pass a
 product from one vendor to another over time) are assigned a name. Product
 names can specify a specific version or refer to the entire product family. For
 example mydistro and mydistro,12.
 
-Components that are used as a link in the UEFI Secure Boot chain of trust are
+**Components** that are used as a link in the UEFI Secure Boot chain of trust are
 assigned names. Examples of components are shim, GRUB, kernel, hypervisors, etc.
+
+Below is an example of a product and component, both in the same `sbat.csv` file. `sbat`
+defines the version of the format of the revocation metadata itself.
+`grub.acme` is an example of a product name.
+```
+sbat,1,SBAT Version,sbat,1,https://github.com/rhboot/shim/blob/main/SBAT.md
+grub.acme,1,Acme Corporation,grub,1.96-8191,https://acme.arpa/packages/grub
+```
 
 We could conceivably support sub-components, but it's hard to conceive of a
 scenario that would trigger a UEFI variable update that wouldn't justify a
@@ -154,7 +162,7 @@ hypervisor or kernel re-release to enforce that sub-component level from there.
 Something like a "level 1.5 hypervisor" that can exist between different kernel
 generations can be considered its own component.
 
-Each component is assigned a minimum global generation number. Vendors signing
+Each component is assigned a **minimum global generation number**. Vendors signing
 component binary artifacts with a specific global generation number are
 required to include fixes for any public or pre-disclosed issue required for
 that generation. Additionally, in the event that a bypass only manifests in a
@@ -162,7 +170,9 @@ specific product's component, vendors may ask for a product-specific generation
 number to be published for one of their product's components. This avoids
 triggering an industry wide re-publishing of otherwise safe components.
 
-A product-specific minimum generation number only applies to the instance of
+In the example above, 1 is sbat's minimum global generation number.
+
+A **product-specific minimum generation number** only applies to the instance of
 that component that is signed with that product name. Another product's
 instance of the same component may be installed on the same system and would
 not be subject to the other product's product-specific minimum generation
@@ -173,6 +183,8 @@ generation number, but is labeled with that number. Rather than having the
 entire industry that uses that component re-release, just that product's
 minimum generation number would be incremented and that product's component
 re-released along with a UEFI variable update specifying that requirement.
+
+In the example above, 1 is grub.acme's product-specific minimum generation number.
 
 The global and product-specific generation number name spaces are not tied to
 each other. The global number is managed externally, and the vast majority of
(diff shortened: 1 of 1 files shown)


julian-klode:
1. I don't understand the question
2. That seems to be a bug IMO. Presumably an earlier spec version required product-specific >= global (which is 4), I don't know for sure. In practice, as long it increases, it won't matter anyway.
3. Need might be a strong word, but the file format is more consistent this way; arguably you only need those for humans to read.
4. The table shows that the number resets after a global event. You can essentially consider that a global event will fold in all product-specific events. So if you had shim.foo 1,2,3 (shim 1), than your revocation list looks like {shim.foo < 3}. After a global event `shim 2`, any shim.foo entries are removed and the list becomes {shim < 2}.
5. In the repositories of those components? Currently they are all 1, presumably at some point revoked sbat levels will be published on the uefi website with dbx updates

SBAT.md
```diff
@@ -140,29 +140,39 @@ value associated with the data currently stored in that variable.
...
 product from one vendor to another over time) are assigned a name. Product
 names can specify a specific version or refer to the entire product family. For
 example mydistro and mydistro,12.
 
-Components that are used as a link in the UEFI Secure Boot chain of trust are
+**Components** that are used as a link in the UEFI Secure Boot chain of trust are
 assigned names. Examples of components are shim, GRUB, kernel, hypervisors, etc.
 
+Below is an example of a product and vendor, both in the same `sbat.csv` file. SBAT
+is a vendor, as it is a link in the UEFI Secure Boot chain of trust.
```
jsetje, line 152: I think you meant to say SBAT acts like a component rather than a vendor. And yes, the sbat version basically does.

SBAT.md
```diff
@@ -140,29 +140,39 @@ value associated with the data currently stored in that variable.
...
+Each component is assigned a **minimum global generation number**. Vendors signing
 component binary artifacts with a specific global generation number are
 required to include fixes for any public or pre-disclosed issue required for
 that generation. Additionally, in the event that a bypass only manifests in a
 specific product's component, vendors may ask for a product-specific generation
 number to be published for one of their product's components. This avoids
 triggering an industry wide re-publishing of otherwise safe components.
 
-A product-specific minimum generation number only applies to the instance of
+In the example above, 1 is sbat's minimum global generation number.
```
jsetje, line 173: I think we've been referring to the sbat "generation" as the version, but I'm not sure that distinction doesn't cause more confusion than it helps. The rest of the doc should be scanned and either fixed up globally of this should use the term version as well.

jsetje (changes requested):
Thanks for taking a look, here are some quick comments.

jsetje:
> Can you define disclosure (from the chart included in the file)?

A disclosure in this context is the event/date where a CVE and fixes for it are made public. It could be worth describing that in a bit more detail, the document was generally written with an audience of system security folks in mind, but the group of folks that have to deal with shim and how it works is actually broader than that. I can try to draft something.

SBAT.md
eshiman: I'm happy to do that, but I'd like you to make the call on the term. In other places, I've been seeing "version-number" and "global number". (I'm assuming global number is a subcategory of version-number)

eshiman:
Hey! I put up some changes in response to your feedback 20 days ago. Is there anything else I need to do to push this forward?

jsetje (approved):

SBAT.md
jsetje: I guess this is slightly confusing:

The number on the sbat line is the version of the format of the sbat metadata. We may or may not ever need to change that. We'll certainly try hard to extend it in a backwards compatible way. It doesn't really revoke anything.

There is a global generation number for a component which needs to be satisfied by any instance of that component. Non-global generation numbers are the product specific ones that allow a component to be revoked at some generation for that product only.

This can be separated out from this PR though.

jsetje:
Thanks for the ping, this had fallen off my radar.

Also, thank you for the edits, other than the one item I flagged this looks good.

SBAT.md
```diff
@@ -140,29 +140,39 @@ value associated with the data currently stored in that variable.
...
 product from one vendor to another over time) are assigned a name. Product
 names can specify a specific version or refer to the entire product family. For
 example mydistro and mydistro,12.
 
-Components that are used as a link in the UEFI Secure Boot chain of trust are
+**Components** that are used as a link in the UEFI Secure Boot chain of trust are
 assigned names. Examples of components are shim, GRUB, kernel, hypervisors, etc.
 
+Below is an example of a product and component, both in the same `sbat.csv` file. `sbat`
+is a component, as it is a link in the UEFI Secure Boot chain of trust.
```
eshiman, line 152: I see in my email you left a comment requesting that I switch this to `'sbat' defines the version of the format of the revocation metadata itself`however I don't see that comment here. (Was it deleted? I also see in your lgtm comment you wanted me to address something you flagged.)
Does that comment still apply?
I had my specific wording originally because the word "component" seems to be used in a specific way in this document, as you mentioned, and I wanted it to be clear what the word "component" referred to. But I do get that components is an overloaded word. It's entirely possible I didn't understand what you meant by "component" in this document. Is `sbat` a component as defined above? Or is it just "the version of the format of the revocation metadata itself"? If that is true, should I use a different example to illustrate what a "component" is?
Here's a guess as to what I could switch it to, to clarify the concern you had with the word being overloaded:
```
**Component** as defined here refers to a component that is committed to a UEFI boot services variable, as opposed to a component loaded from a file in a filesystem. They are used as a link in the UEFI Secure Boot chain of trust and are assigned names...
```

I'm more than happy to rewrite it in the way you prefer, I just want to make sure that is still what you want, as that comment seems to be gone.

jsetje:
Yes, I tried to turn it back into a review and deleted the comment, same content though.

PR #403, opened 2021-09-01
httpboot: Ignore case when checking HTTP headers

frozencemetery:
Some servers (HAProxy) yield Content-Length in lowercase, and shim would
fail to find it.


julian-klode (approved):
:+1:

PR #404, opened 2021-09-02
csv: detect end of csv file correctly

xypron:
.sbat sections contain user supplied data. We should not assume that it is
well formed. The last line feed might be missing or it might not be at the 
end of the file. Instead one or more \0 might follow.

In parse_csv_data() variable 'line' is a pointer with a value between
the values of 'data' and 'data_end'.

There is no reason to check that it is non-zero after assigning it
from 'data' as we already check 'data'.

Instead at the beginning of the file and after each line we must check that
we have not reached the end of the file marked by a '\0' character.

### DIFF
--- a/csv.c
+++ b/csv.c
@@ -15,6 +15,7 @@ parse_csv_line(char * line, size_t max, size_t *n_columns, const char *columns[]
 	char *token = NULL;
 
 	bool valid = true;
+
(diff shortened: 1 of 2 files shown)


xypron:
By mistake in the original pull request I included a superfluous '#include <stdio.h>'. It is removed now.

PR #405, opened 2021-09-03
fallback: incorrect check after AllocateZeroPool()

xypron:
After calling AllocateZeroPool() we must check the returned pointer.

Fixes: 3ce517fdbb4e ("Add a fallback loader for when shim is invoked as BOOTX64.EFI")


frozencemetery:
Thank you for the pull request.  However, I believe this is already part of #402 so let's keep discussion there.

PR #407, opened 2021-09-03
Correct signedness when calling string functions.

xypron:
The signedness of gnu-efi's CHAR8 is independent of the signedness of char.

Our string functions expect (char *) or (const char *) while some of our
variables are CHAR8.

Provide the necessary conversions.

Signed-off-by: Heinrich Schuchardt <xypron.glpk@gmx.de>

### DIFF
--- a/elf_riscv64_efi.lds
+++ b/elf_riscv64_efi.lds
@@ -1,0 +1,77 @@
+/* SPDX-License-Identifier: GPL-2.0+ */
+
+OUTPUT_FORMAT("elf64-littleriscv", "elf64-littleriscv", "elf64-littleriscv")
+OUTPUT_ARCH(riscv)
+ENTRY(_start)
+SECTIONS {
+.text 0x0 :
+	{
+		_text = .;
+		*(.text.head)
+		*(.text)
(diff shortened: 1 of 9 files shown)


elf_riscv64_efi.lds
```diff
@@ -0,0 +1,77 @@
+/* SPDX-License-Identifier: GPL-2.0+ */
```
vathpela, line 1: I think you probably didn't mean to add a riscv64 linker script here?

pe.c
```diff
@@ -651,6 +651,8 @@ static const UINT16 machine_type =
 	IMAGE_FILE_MACHINE_I386;
 #elif defined(__ia64__)
 	IMAGE_FILE_MACHINE_IA64;
+#elif defined (__riscv) && __riscv_xlen == 64
+	IMAGE_FILE_MACHINE_RISCV64;
```
vathpela, line 655: This also should be on a different PR.

vathpela (changes requested):
Most of this looks good, but you've got a couple of changes for RISCV64 here that need to be on a different PR.

PR #409, opened 2021-09-03
MokManager: remove redundant logical constraints

xypron:
CertList->SignatureSize is of type UINT32 which is always positive.

If CertList->SignatureListSize == 0, then
CertList->SignatureListSize <= CertList->SignatureSize
is also true.

Remove the redundant CertList->SignatureListSize == 0 checks.

A message "Corrupted signature list" is better suited then
"Invalid MOK detected! Ignoring MOK List." in this case.

### DIFF
--- a/MokManager.c
+++ b/MokManager.c
@@ -136,17 +136,14 @@ static UINT32 count_keys(void *Data, UINTN DataSize)
 	void *end = Data + DataSize;
 
 	while ((dbsize > 0) && (dbsize >= CertList->SignatureListSize)) {
-		/* Use ptr arithmetics to ensure bounded access. Do not allow 0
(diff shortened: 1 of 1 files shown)


frozencemetery (approved):

julian-klode:
While we're at it,  `(void *)(CertList + 1) > end` is undefined behavior if it is true, it should be `end - (void*)CertList < 1`.

xypron:
@julian-klode:

`(void *)(CertList + 1) > end` checks that a complete entry with the same type as CertList is inside the memory area ending at pointer end and can be safely processed. The check that you propose would only require a single first byte to be available.

If you don't see anything wrong with the current patch, please, add your approval.

PR #410, opened 2021-09-03
Correct signedness when calling string functions.

xypron:
The signedness of gnu-efi's CHAR8 is independent of the signedness of char.

Our string functions expect (char *) or (const char *) while some of our
variables are CHAR8.

Provide the necessary conversions.

### DIFF
--- a/Cryptlib/SysCall/BaseStrings.c
+++ b/Cryptlib/SysCall/BaseStrings.c
@@ -3,7 +3,7 @@
 CHAR8 *
 AsciiStrCat(CHAR8 *Destination, const CHAR8 *Source)
 {
-	UINTN dest_len = strlen((CHAR8 *)Destination);
+	UINTN dest_len = strlen((char *)Destination);
 	UINTN i;
 
 	for (i = 0; Source[i] != '\0'; i++)
@@ -61,7 +61,7 @@ WriteUnaligned32(UINT32 *Buffer, UINT32 Value)
 UINTN
 AsciiStrSize(const CHAR8 *string)
 {
-	return strlen(string) + 1;
+	return strlen((char *)string) + 1;
 }
 
 /* Based on AsciiStrDecimalToUintnS() in edk2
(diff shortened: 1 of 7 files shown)


xypron:
Fixed issues mentioned in https://github.com/rhboot/shim/pull/407

xnox:
But our forked gnu-efi defines CHAR8 as char, and we do not support building against any other version of gnu-efi, and it was my understanding that doing casts like these => or building against other versions of gnu-efi leads to incorrect verification of sbat sections.

It would be nice if we did use char8 or CHAR8 throughout, without mixing or casting.

I think I am confused a little.

Are there any casting issues in the current code, when built against vendored gnu-efi?

vathpela:
> But our forked gnu-efi defines CHAR8 as char, and we do not support building against any other version of gnu-efi, and it was my understanding that doing casts like these => or building against other versions of gnu-efi leads to incorrect verification of sbat sections.
> 
> It would be nice if we did use char8 or CHAR8 throughout, without mixing or casting.
> 
> I think I am confused a little.
> 
> Are there any casting issues in the current code, when built against vendored gnu-efi?

I'm not seeing what issue this fixes either - additionally, if I compare the built objects, all of the differences are in the debuginfo.

xypron:
I was having trouble compiling for RISC-V without this patch. But may that is due to the definitions for RISC-V that I copied from upstream gnu-efi in https://github.com/rhboot/gnu-efi/pull/3.

What does not make any sense for me in the shim code is converting to CHAR8 * when calling string functions. Converting to char * may be a nop currently. But we should strive to get our code into upstream gnu-efi. Therefore converting to char * here makes most sense to me.

julian-klode:
Part of the story here is that we don't want to use gnu-efi long term, AFAIUI, so we should patch gnu-efi to work best for shim and not shim to work better with upstream gnu-efi.

xypron:
@julian-klode 
Thanks for providing the direction. I will update https://github.com/rhboot/gnu-efi/pull/3 accordingly.

PR #411, opened 2021-09-03
pe: simplify generate_hash()

xypron:
Copying the value of datasize_in to two further variables and then using
all three randomly in the code makes it hard to read.

datasize_in is never changed in generate_hash() so we can do with this
parameter alone. Let's declare it as const to verify that its value
is invariant.

### DIFF
--- a/include/pe.h
+++ b/include/pe.h
@@ -25,7 +25,7 @@ handle_image (void *data, unsigned int datasize,
 	      UINTN *alloc_pages);
 
 EFI_STATUS
-generate_hash (char *data, unsigned int datasize_in,
+generate_hash (char *data, unsigned int datasize,
 	       PE_COFF_LOADER_IMAGE_CONTEXT *context,
 	       UINT8 *sha256hash, UINT8 *sha1hash);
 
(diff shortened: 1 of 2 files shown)


xnox:
Given that there is no "datasize_out" and that most of the function code uses "datasize" maybe it's better to rename the function argument to "datasize" and then drop all references to "size" and "datasize_in"? this should make the diff smaller, no?

xypron:
@xnox: Thank you for reviewing. I will update the patch and force push it.

xypron:
@xnox: I have update the patch according to your recommendation. Please, review again.

xypron:
Rebased on current origin/main

frozencemetery (changes requested):
While the rest of the code is fine, I don't think making the function parameter const increases readability, or really gains anything.  If anything, it's unusual enough that it makes the function harder to read in my opinion.

Code seems fine otherwise.

xypron:
@frozencemetery: thank you for reviewing. I removed const and force pushed to corrected patch.

frozencemetery (approved):
Thanks, I'm happy with this.

PR #412, opened 2021-09-06
sbat: append NUL byte to .sbat section

xypron:
When the UEFI firmware loads the .sbat section into memory it only
intializes VirtSize bytes. Directly behind the .sbat section will be
unintialized memory with unknown content.

Shim's image file header is not available at runtime. Up to now the
.sbat section ended with LF. So it was not possible to determine the
end of the .bat section in memory.

By appending a NUL byte to the .sbat section we get a well defined
file end.


xypron:
The check for the NUL byte was added in #404 "csv: detect end of csv file correctly"

vathpela:
I don't understand why we would need this?  We're already padding it before parsing the section, at https://github.com/rhboot/shim/blob/main/pe.c#L843 :
```
	sbat_size = SBATSize + 1;
	sbat_data = AllocatePool(sbat_size);
	if (!sbat_data) {
		console_print(L"Failed to allocate .sbat section buffer\n");
		return EFI_OUT_OF_RESOURCES;
	}
	CopyMem(sbat_data, SBATBase, SBATSize);
	sbat_data[SBATSize] = '\0';

	efi_status = parse_sbat_section(sbat_data, sbat_size, &n, &entries);
```

xypron:
@vathpela:
Thanks for pointing me to the code section.

It seems I misplaced _esbat in my elf_riscv64_efi.lds.

PR #415, opened 2021-09-09
tests: add "include-fixed" GCC directory to include directories

diabonas:
On Arch Linux, the GCC version of `limits.h` is in the `include-fixed` instead of the `include` directory. It needs to be included in the include directories list, otherwise attempting to compile the test suite fails with the following error:
```
In file included from /usr/include/efivar/efivar-dp.h:22,
                 from /usr/include/efivar/efivar.h:238,
                 from include/test.h:51,
                 from shim.h:68,
                 from csv.c:6:
/usr/include/limits.h:124:16: fatal error: limits.h: No such file or directory
  124 | # include_next <limits.h>
      |                ^~~~~~~~~~
compilation terminated.
```

### DIFF
--- a/include/test.mk
+++ b/include/test.mk
@@ -46,6 +46,10 @@ CFLAGS = $(OPTIMIZATIONS) -std=gnu11 \
 	 -DSHIM_UNIT_TEST \
 	 "-DDEFAULT_DEBUG_PRINT_STATE=$(DEBUG_PRINTS)"
 
+# On some systems (e.g. Arch Linux), limits.h is in the "include-fixed" instead
+# of the "include" directory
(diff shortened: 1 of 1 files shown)


frozencemetery (commented):
A comment in test.mk indicating that this is for Arch would be nice.  Otherwise, while have to add this isn't ideal, the change looks fine to me.

diabonas:
> A comment in test.mk indicating that this is for Arch would be nice.

Sure, done :) I don't think adding "inline" comments in a variable definition continued over multiple lines using backslashes is possible in a Makefile, so used a separate addtion to `CFLAGS` at the very end instead (which builds fine as well).

frozencemetery (approved):
Yeah, I think like this is fine.  (If order became tricky, it could be another variable that gets added where you had it originally.)

PR #416, opened 2021-09-16
Better console checks

vathpela:
There's a potential for future (or even existing) platforms to without console output to not populate ST->ConIn and ST->ConOut.  This patch set adds checks to all of our usage of those protocol shortcuts.


frozencemetery (approved):
Looks fine as an improvement.

PR #417, opened 2021-09-20
Test import_mok_state() when MokListRT would be bigger than available size

vathpela:
This patch set adds `test_mok_mirror_1()`, which is basically like `test_mok_mirror_0()` except with a variable size limit smaller than `MokListRT` would be, and with the `.must_be_absent` flag set to `true` in the test data, so the test will fail if the variable is successfully created *or* if the data is not in the mokvars config variable.

### DIFF
--- a/include/mock-variables.h
+++ b/include/mock-variables.h
@@ -115,6 +115,8 @@ void mock_uninstall_query_variable_info(void);
 void mock_reset_variables(void);
 void mock_reset_config_table(void);
 void mock_finalize_vars_and_configs(void);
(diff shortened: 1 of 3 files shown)


julian-klode:
Should the tests maybe have more descriptive names than 0 and 1? And/Or comments would be nice, so you won't need to dig through commits later to find out what the test tests.

frozencemetery (changes requested):
This needs rebased on main - I'm not sure where github is getting "no conflicts" from.

jsetje (approved):

PR #422, opened 2021-10-05
fallback: Fix for BootOrder crash when index returned by find_boot_option() is not in current BootOrder list

rmetrich:
Seen on Dell PowerEdge R740 when booting with BOOTX64 constantly.

This patch keeps the behaviour previous to commit #1b30c2b by returning the index of the "Linux" entry.
Then a check is made to find the entry in the current BootOrder:
- if it isn't there, prepend the entry and copy the rest (this enlarges the BootOrder array by 1)
- if it's there, prepend the entry and copy all remaining entries

---

Example of outputs on a Dell PowerEdge R740:
- 0000 is BOOTX64 entry
- other entries are Dell's default ones (internal, no "Linux" there)

1. Entry not already existing in BootOrder

  ~~~
  set_boot_order:486: Original nbootorder: 3
  Original BootOrder: 0000 0003 0004
   :
  add_to_boot_list:578: device path: "HD(1,GPT,99D47E76-590F-48FD-8FD6-0A0CE790D635)/\EFI\redhat\shimx64.efi"
  find_boot_option:454: Found boot entry "Boot0005" with label "Red Hat Enterprise Linux" for file "\EFI\redhat\shimx64.efi"
  add_to_boot_list:623: New nbootorder: 4
  BootOrder: 0005 0000 0003 0004
  find_boot_options:937: Found directory named "Dell"
  update_boot_order:509: nbootorder: 4
  BootOrder: 0005 0000 0003 0004
  ~~~

2. Entry not existing at all

  ~~~
  set_boot_order:486: Original nbootorder: 3
  Original BootOrder: 0000 0001 0002
   :
  add_to_boot_list:578: device path: "HD(1,GPT,99D47E76-590F-48FD-8FD6-0A0CE790D635)/\EFI\redhat\shimx64.efi"
  add_boot_option:245: Creating boot entry "Boot0005" with label "Red Hat Enterprise Linux" for file "\EFI\redhat\shimx64.efi"
  add_boot_option:282: nbootorder: 4
  BootOrder: 0005 0000 0001 0002
  find_boot_options:937: Found directory named "Dell"
  update_boot_order:509: nbootorder: 4
  BootOrder: 0005 0000 0001 0002
  ~~~

3. Entry already existing in BootOrder

  ~~~
  set_boot_order:486: Original nbootorder: 4
  Original BootOrder: 0000 0005 0001 0002
   :
  add_to_boot_list:578: device path: "HD(1,GPT,99D47E76-590F-48FD-8FD6-0A0CE790D635)/\EFI\redhat\shimx64.efi"
  find_boot_option:454: Found boot entry "Boot0005" with label "Red Hat Enterprise Linux" for file "\EFI\redhat\shimx64.efi"
  add_to_boot_list:623: New nbootorder: 4
  BootOrder: 0005 0000 0001 0002
  find_boot_options:937: Found directory named "Dell"
  update_boot_order:509: nbootorder: 4
  BootOrder: 0005 0000 0001 0002
  ~~~


fallback.c
```diff
@@ -612,18 +612,45 @@ add_to_boot_list(CHAR16 *dirname, CHAR16 *filename, CHAR16 *label, CHAR16 *argum
 	if (EFI_ERROR(efi_status)) {
 		add_boot_option(dp, full_device_path, fullpath, label,
 				arguments);
-	} else if (option != 0) {
+	} else {
```
frozencemetery, line 615: We can save a level of indentation here by renaming `err:` to `done:` (or so) and then `goto done;` on the previous line (i.e., after add_boot_option()).

fallback.c
```diff
@@ -612,18 +612,45 @@ add_to_boot_list(CHAR16 *dirname, CHAR16 *filename, CHAR16 *label, CHAR16 *argum
...
-			sizeof (CHAR16) * (nbootorder - option - 1));
-		FreePool(bootorder);
-		bootorder = newbootorder;
+		/* Search for the option in the current bootorder */
+		for (bootnum = 0; bootnum < nbootorder; bootnum++)
+			if (bootorder[bootnum] == option)
+				break;
+		if (bootnum == nbootorder) {
+			/* Option not found, prepend option and copy the rest */
+			newbootorder = AllocateZeroPool(sizeof (CHAR16)
```
frozencemetery, line 624: No space between `sizeof` and the open parenthesis (here and in the cases below).

frozencemetery (commented):
Thank you for the patch.  I have some code style stuff inline (and it would be great if you could reduce the length of the commit message subject), but otherwise I don't see anything wrong (and we can fix those on merge if needed).  Probably should wait for a second review, however.

fallback.c
rmetrich: Right, renamed to "done" and fixed the broken "return" stanza on error which was leaking memory in the mean time.

fallback.c
rmetrich: OK, I personally prefer a space between `sizeof` and type because it's a compiler directive, not a function call, but well, depends on the style I guess ...

PR #423, opened 2021-10-05
Introduce a new MOK variable called MokListTrustedRT

esnowberg:
Introduce a new MOK variable called MokListTrustedRT.  It allows an end-user
to decide if they want to trust MOKList keys within the soon to be booted
Linux kernel.  This variable does not change any functionality within shim
itself.  When Linux boots, if MokListTrustedRT is set and
EFI_VARIABLE_NON_VOLATILE is not set, keys in MokListRT are loaded into the
.machine keyring instead of the .platform keyring.

Signed-off-by: Eric Snowberg <eric.snowberg@oracle.com>

### DIFF
--- a/MokManager.c
+++ b/MokManager.c
@@ -41,6 +41,12 @@ typedef struct {
 } __attribute__ ((packed)) MokDBvar;
 
 typedef struct {
+	UINT32 MokTMLState;
+	UINT32 PWLen;
+	CHAR16 Password[SB_PASSWORD_LEN];
+} __attribute__ ((packed)) MokTMLvar;
+
+typedef struct {
 	INT32 Timeout;
 } __attribute__ ((packed)) MokTimeoutvar;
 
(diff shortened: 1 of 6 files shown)


esnowberg:
Upstream Linux seems to be receptive to this approach.   However they will not move forward with the Linux review until code enters shim first.  Please see the patch above for review.

References to this topic:

https://lwn.net/ml/linux-kernel/20210914211416.34096-1-eric.snowberg@oracle.com/
https://lwn.net/ml/linux-kernel/270f47e1b152a1fb8fd909ec188b5573176980fc.camel@kernel.org/
https://github.com/esnowberg/mokutil/tree/0.3.0-mokvars-v2

MokManager.c
```diff
@@ -1678,6 +1684,121 @@ static EFI_STATUS mok_db_prompt(void *MokDB, UINTN MokDBSize)
...
+		return EFI_ACCESS_DENIED;
+	}
+
+	if (var->MokTMLState == 0)
+		ret = console_yes_no(trust_tml);
+	else
+		ret = console_yes_no(untrust_tml);
+
+	if (ret == 0) {
+		LibDeleteVariable(L"MokTML", &SHIM_LOCK_GUID);
```
vathpela, line 1775: Here we're using "MokTML"....

MokManager.c
```diff
@@ -1678,6 +1684,121 @@ static EFI_STATUS mok_db_prompt(void *MokDB, UINTN MokDBSize)
...
+		return EFI_ABORTED;
+	}
+
+	if (var->MokTMLState == 0) {
+		efi_status = gRT->SetVariable(L"MokTMLState", &SHIM_LOCK_GUID,
```
vathpela, line 1780: But here we've got "MokTMLState".  Also this should use RT-> instead of gRT->.

MokManager.c
```diff
@@ -2436,6 +2580,16 @@ static EFI_STATUS check_mok_request(EFI_HANDLE image_handle)
 		console_error(L"Could not retrieve MokDB", efi_status);
 	}
 
+	efi_status = get_variable(L"MokTML", (UINT8 **) & MokTML, &MokTMLSize,
+				  SHIM_LOCK_GUID);
+	if (!EFI_ERROR(efi_status)) {
+		efi_status = LibDeleteVariable(L"MokTML", &SHIM_LOCK_GUID);
```
vathpela, line 2586: Here we're using "MokTML" again...

mok.c
```diff
@@ -166,6 +166,20 @@ struct mok_state_variable mok_state_variable_data[] = {
 		  MOK_VARIABLE_MEASURE,
 	 .pcr = 7,
 	},
+	{.name = L"MokTMLState",
+	 .name8 = "MokTMLState",
+	 .rtname = L"MokListTrustedRT",
+	 .rtname8 = "MokListTrustedRT",
```
vathpela, line 172: And here we've got "MokTMLState", but the RT variable is MokListTrustedRT".  Unless I've missed something major, "MokTML" vs "MokTMLState" is a bug.  In any case, I think the BS|NV variable needs to reflect the name of the RT variable; please make it "MokListTrusted".

MokManager.c
```diff
@@ -1678,6 +1684,121 @@ static EFI_STATUS mok_db_prompt(void *MokDB, UINTN MokDBSize)
...
+	int ret;
+	CHAR16 *untrust_tml[] = { L"Do not trust the MOK list", NULL };
+	CHAR16 *trust_tml[] = { L"Trust the MOK list", NULL };
+
+	if (MokTMLSize != sizeof(MokTMLvar)) {
+		console_notify(L"Invalid MokTML variable contents");
+		return EFI_INVALID_PARAMETER;
+	}
+
+	ST->ConOut->ClearScreen(ST->ConOut);
```
vathpela, line 1707: This should use clear_screen().

MokManager.c
```diff
@@ -1678,6 +1684,121 @@ static EFI_STATUS mok_db_prompt(void *MokDB, UINTN MokDBSize)
...
+		efi_status = gRT->SetVariable(L"MokTMLState", &SHIM_LOCK_GUID,
+					      EFI_VARIABLE_NON_VOLATILE |
+					      EFI_VARIABLE_BOOTSERVICE_ACCESS,
+					      1, &dbval);
+		if (EFI_ERROR(efi_status)) {
+			console_notify(L"Failed to set TML state");
+			return efi_status;
+		}
+	} else {
+		efi_status = gRT->SetVariable(L"MokTMLState", &SHIM_LOCK_GUID,
```
vathpela, line 1789: Likewise.

MokManager.c
esnowberg: The MokTML variable is set through mokmanager to indicate the user wants to change the contents of MokTMLState.  Would it be less confusing if I used a different variable name?

MokManager.c
esnowberg: I'll change this to RT

mok.c
esnowberg: I'l change this to MokListTrusted

esnowberg:
Thanks for your review Peter.  I'll also add the new var to the MokVars.txt.

MokManager.c
vathpela: Ah, now I understand. For consistency with other variables, even though it's kind of a mouthful, how about "MokListTrustedNew" for the variable mokutil sets?

vathpela:
Pushed as 4e513405b4f164

mok.c
```diff
@@ -166,6 +166,20 @@ struct mok_state_variable mok_state_variable_data[] = {
...
+	{.name = L"MokListTrusted",
+	 .name8 = "MokListTrusted",
+	 .rtname = L"MokListTrustedRT",
+	 .rtname8 = "MokListTrustedRT",
+	 .guid = &SHIM_LOCK_GUID,
+	 .yes_attr = EFI_VARIABLE_BOOTSERVICE_ACCESS |
+		     EFI_VARIABLE_NON_VOLATILE,
+	 .no_attr = EFI_VARIABLE_RUNTIME_ACCESS,
+	 .flags = MOK_MIRROR_DELETE_FIRST |
+		  MOK_VARIABLE_MEASURE |
```
baloo, line 178: What was the intention for this flag to be set?
This pushes the content of the MokListTrusted in the PCR7, which doesn't make sense to me. This should only change the behavior of linux which is after the `ExitBootServices` that this variable is set to zero or one doesn't change the Secure Boot Policy  (https://trustedcomputinggroup.org/wp-content/uploads/TCG_PCClient_PFP_r1p05_v23_pub.pdf#%5B%7B%22num%22%3A140%2C%22gen%22%3A0%7D%2C%7B%22name%22%3A%22XYZ%22%7D%2C33%2C552%2C0%5D) of the machine before that and should not make it to PCR7

See also:
 https://github.com/rhboot/shim/issues/492
 https://github.com/rhboot/shim/issues/484

PR #426, opened 2021-10-20
Fix a component version in SBAT.example.md

nicholasbishop:
In the bug2 section, the first Debian `.sbat` has a `grub,1`
component. But at this point in the story, `grub,1` has been revoked by
the update in the bug1 section. Updated it to `grub,2` so that it passes
that check.


vathpela:
Pushed as 63f48e0ce2b826138fd27ff592ff471ba9cb0d39

PR #427, opened 2021-10-26
Add code of conduct

frozencemetery:
This is the standard Contributor Covenant.

See-also: https://www.contributor-covenant.org/

[I added my email address as the contact on the theory that if someone else wanted to do it they already would have, but if that's incorrect I'm happy to change it.]


steve-mcintyre:
Looks good to me!

PR #433, opened 2021-11-09
fallback: fix fallback not passing arguments of the first boot option

martinezjavier:
The buffer used to read the data in the CSV is declared as a stack variable
in the try_boot_csv() function, but a pointer to the arguments field of the
first boot option is stored in the global first_new_option_args variable.

Later, when is used set the arguments to boot the first entry, the variable
points to memory that no longer exists. This leads to booting an entry with
garbage as arguments instead of the correct value.

Reported-by: Alexander Larsson <alexl@redhat.com>
Signed-off-by: Javier Martinez Canillas <javierm@redhat.com>

### DIFF
--- a/fallback.c
+++ b/fallback.c
@@ -248,7 +248,7 @@ add_boot_option(EFI_DEVICE_PATH *hddp, EFI_DEVICE_PATH *fulldp,
 
 			if (!first_new_option) {
 				first_new_option = DuplicateDevicePath(fulldp);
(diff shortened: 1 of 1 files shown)


frozencemetery (approved):

gicmo (approved):
FWIW this looks good to me!

alexlarsson:
I built this and verified that it fixed my problem

PR #438, opened 2021-12-03
shim: Don't stop forever at "Secure Boot not enabled" notification

rmetrich:
Requesting a keystroke when Secure Boot is not enabled and verbosity is enabled is really annoying.

### DIFF
--- a/shim.c
+++ b/shim.c
@@ -429,8 +429,11 @@ BOOLEAN secure_mode (void)
 		return FALSE;
 
 	if (variable_is_secureboot() != 1) {
-		if (verbose && !in_protocol && first)
-			console_notify(L"Secure boot not enabled");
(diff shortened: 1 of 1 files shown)


frozencemetery (commented):
The commit message mentions that "Requesting a keystroke... is really annoying", but it doesn't look like this patch fixes that.  Am I missing something?

rmetrich:
> The commit message mentions that "Requesting a keystroke... is really annoying", but it doesn't look like this patch fixes that - just adds mention that a keystroke is needed. Am I missing something?

Well yes,`console_notify()` requires hitting Enter to continue, so as soon as Verbose mode is set, the system will stop at the prompt and waiting for the user to hit Enter which prevents unattended boot.

PR #443, opened 2022-01-18
pe: missing perror argument

xypron:
perror(L"%d sections contain entry point\n") lacks an argument
corresponding to %d.


xypron:
Should we add __attribute__ ((format (printf, 4, 5))) to the LogError_() definition to find such errors more easily?

frozencemetery (approved):
Thanks!

Issue #418, opened 2021-09-21
fallback presents a scary message to users


jprvita:
In the past, before fallback cared about TPMs, it would always chain-load the entry it had created, but to better support TPMs it now reboots the system if a TPM is present. Due to some firmware implementations overriding any changes made outside of the firmware, the system may end-up in a reboot loop of BOOTX64.EFI, as already reported by @lcp on https://github.com/rhboot/shim/issues/128.

This problem is currently addressed by commit a5db51a52e, which presents a screen warning the user the system is about to reboot, with a countdown, allowing them to tell the system to keep booting instead. While this solution works (and thanks @lcp for implementing it!), it has a few shortcomings:
 1. It makes an otherwise glitch-free boot process not smooth anymore;
 2. The message presented is not accessible / potentially scary for non-technical users: if they press a key to interrupt the boot process, the meaning of each option is not really clear for users not familiar with how shim / fallback work;
 3. The whole experience is made a bit worse by the fact that after selecting "Continue boot" / "Always continue boot", the screen will remain frozen until something else draw on the framebuffer. If GRUB is configured to be quiet for a glitch-free boot, this may last several seconds until the kernel has started and loaded the manufacturer logo from BGRT, which gives the impression that the whole boot process froze.
 4. This Boot Option Restoration screen overwrites all the debug information printed before it is displayed, essentially neutering FALLBACK_VERBOSE or SHIM_VERBOSE and making it impossible to enable debug without rebuilding fallback.

Some of these may be seen as non-issues for distributions that use a traditional installer, like Fedora and OpenSuse, where the correct boot entry is created before first boot by the installer and the fallback path is only taken if the boot entry got invalidated somehow (ESP UUID has changed, firmware was factory-reset etc). But for distributions that are distributed as "ready-to-dd" disk image, like Endless OS, fallback is responsible for creating the boot entry on first boot, and then update it on the second boot after the ESP UUID changes during the first-boot re-partitioning process (since we don't want the partitions on every installation to have the same UUIDs). This means users will see this message on the first and second boots in such scenarios.

jprvita:
To avoid the problems mentioned above, we have been shipping a downstream change on Endless OS to always try to chain-load the new boot entry, despite the presence of a TPM. But I am now trying to improve things upstream through https://github.com/rhboot/shim/pull/419.

dsd:
[...]

Issue #128, opened 2018-05-09
shim/fallback 14 may fail to boot in some machines


lcp:
We recently got a bug report(*) that shim/fallback was stuck in an infinite reset loop. It's because that the firmware seems to ignore the restored boot option and also provides TPM protocol.

One possible solution is to add a count down menu before resetting the system so that the user has the chance to stop system reset and choose to load the boot option directly. We could also set a NV variable to notify fallback.efi not to reset afterward.

(*) https://bugzilla.opensuse.org/show_bug.cgi?id=1092000

PR #445, opened 2022-02-01
fallback: Add build flag to always try to chain-load the new boot entry

jprvita:
The firmware on some Acer machines (and maybe others) always resets the
boot entries and BootOrder variable to what was defined in the firmware
setup program, overriding any external changes (including the changes
made by fallback).

Before shim cared about TPMs this was not a problem in practice, as
fallback would create and chain-load a boot entry for the OS on every
boot. However, since commit 431b8a2e75 the system is restarted if a TPM
is detected on the system, triggering an infinite reboot loop in systems
with such firmware. This is a known problem which has been previously
reported on https://github.com/rhboot/shim/issues/128

More recently, the problem has been addressed by commit a5db51a52e,
which presents a screen with a countdown to the user, where they can
interrupt boot and choose to have fallback always chain-load the new
entry instead of restarting the system, to break out of the reboot loop.
While this solution works, it has a few shortcomings:

 1. It makes an otherwise glitch-free boot process not smooth anymore.
 2. The message presented is not accessible / potentially scary for
    non-technical users: if they press a key to interrupt the boot
    process, the meaning of each option is not really clear for users
    not familiar with how shim and fallback work.
 3. The whole experience is made a bit worse by the fact that after
    selecting "Continue boot" / "Always continue boot", the screen will
    remain frozen until something else draw on the framebuffer. If GRUB
    is configured to be quiet, for a glitch-free boot, this may last
    several seconds until the kernel has started and loaded the
    manufacturer logo from BGRT, which gives the impression that the
    whole boot process froze.
 4. This Boot Option Restoration screen overwrites all the debug
    information printed before it is displayed, essentially neutering
    FALLBACK_VERBOSE or SHIM_VERBOSE and making it impossible to enable
    debug without rebuilding fallback.

This commit adds a build-time flag that forces fallback to always try to
chain-load the newly created boot entry, in the same way it did before
TPM support was added.

Fixes: https://github.com/rhboot/shim/issues/418

Signed-off-by: João Paulo Rechi Vita <jprvita@endlessos.org>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,6 +65,10 @@ ifneq ($(origin FALLBACK_NONINTERACTIVE), undefined)
 	CFLAGS += -DFALLBACK_NONINTERACTIVE
 endif
 
+ifneq ($(origin FALLBACK_NEVER_REBOOT), undefined)
+	CFLAGS += -DFALLBACK_NEVER_REBOOT
+endif
+
 ifneq ($(origin FALLBACK_VERBOSE_WAIT), undefined)
 	CFLAGS += -DFALLBACK_VERBOSE_WAIT=$(FALLBACK_VERBOSE_WAIT)
 endif
--- a/fallback.c
+++ b/fallback.c
@@ -1059,6 +1059,9 @@ try_start_first_option(EFI_HANDLE parent_image_handle)
 static UINT32
 get_fallback_no_reboot(void)
 {
+#ifdef FALLBACK_NEVER_REBOOT
+	return 1;
+#else
 	EFI_STATUS efi_status;
 	UINT32 no_reboot;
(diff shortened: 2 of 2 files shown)


frozencemetery (changes requested):
So far, I do not think a build flag for this is a good idea.  This doesn't need to be configurable; it needs to be *working*.  Adding a configuration handle is asking folks building shims which of two partly-broken behaviors they'd like to have.

Your criticisms of the countdown approach seem extraneous: none of them are addressed by this proposed change that I can see.

Based on what you've stated so far, it seems best to detect the reboot loop and handle it - and inform the machine vendor(s).

PR #448, opened 2022-02-03
Remove post-process-pe on 'make clean'

vathpela:


frozencemetery (approved):
Does what it says on the tin.

PR #455, opened 2022-02-18
Allow MokListTrusted to be enabled by default

esnowberg:
Within previous versions of shim the MokListTrusted var did not
exist.  Add the ability through a compile time option to set
the MokListTrustedRT when the BS var does not exist.  When
the BS var exists, MokListTrustedRT is not set.  This inverse
logic allows a distro to give the end-user the ability to opt
out of this feature instead of opting in.

Many Linux distros carry out of tree patches to trust the mok
keys by default.  These out of tree patches can be dropped
when compiling shim this way and using a Linux kernel that
supports MokListTrustedRT.

Signed-off-by: Eric Snowberg <eric.snowberg@oracle.com>

### DIFF
--- a/MokManager.c
+++ b/MokManager.c
@@ -1776,17 +1776,7 @@ static EFI_STATUS mok_tml_prompt(void *MokTML, UINTN MokTMLSize)
 		LibDeleteVariable(L"MokListTrustedNew", &SHIM_LOCK_GUID);
 		return EFI_ABORTED;
 	}
-
 	if (var->MokTMLState == 0) {
-		efi_status = RT->SetVariable(L"MokListTrusted", &SHIM_LOCK_GUID,
-					      EFI_VARIABLE_NON_VOLATILE |
-					      EFI_VARIABLE_BOOTSERVICE_ACCESS,
-					      1, &dbval);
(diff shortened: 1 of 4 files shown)


frozencemetery (commented):
I'm generally wary of compile-time options to shim.  Is there a reason we wouldn't just make this behavior the default?

esnowberg:
I did it this way incase someone built a shim the other way and might see this as being a regression.  If it would be acceptable to just change the default behavior, let me know and I’ll update the pull request with the change.

frozencemetery:
I'll defer to vathpela on that.

jsetje (approved):

jsetje:
I'm fine either way, I guess I have a slight preference for this being the default behavior.

jsetje:
This was briefly discussed and based on the assumption that this will end up enabled in nearly every distro anyway, we might as well have the upstream default enable this as well.

vathpela:
> This was briefly discussed and based on the assumption that this will end up enabled in nearly every distro anyway, we might as well have the upstream default enable this as well.

Yeah, I'd rather just make it the default.

frozencemetery (approved):

PR #459, opened 2022-04-12
Add support for NX (W^X) mitigations.

### LINKED ISSUE #432: Implement NX (w^x) support
A couple of things need to be done to correctly support NX:
- we need to be able to modify page mapping flags (https://mantis.uefi.org/mantis/view.php?id=2262)
- `load_image()` needs to know how to map binary sections with the correct flags
- Currently have no idea how the firmware loader (and shim's `load_image()` as well) can detect whether a binary is safe (or not) to load with NX, i.e. if it needs all mappings to be executable OR UEFI_MEMORY_ATTRIBUTE present in order to function correctly.

vathpela:
Resolves #432 .


frozencemetery (commented):
Nothing jumps out as obviously incorrect.

vathpela:
I've got an issue with x86_64 kernel loading in grub that I want to understand better before I push this.

PR #471, opened 2022-05-04
CI: drop Fedora ia32 builders

frozencemetery:
We don't build ia32 in Fedora anymore.

[Last commit of #469 split out as requested]


Conan-Kudo:
We don't? Then how does 32-bit UEFI to 64-bit Linux work? That's still a supported target.

PR #473, opened 2022-05-13
Fix github PR hook built test failures on Fedora 36

jsetje:
This may or may not be appropriate.

If the variable types can be resolved in the more recent efivar-devel, then the first patch should not be merged. The ull literals
should likely gain new names.

### DIFF
--- a/include/test.h
+++ b/include/test.h
@@ -50,6 +50,10 @@
 
 #include <efivar/efivar.h>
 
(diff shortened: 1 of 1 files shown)


vathpela:
I think the issues here have been fixed by 6aac5959bbb6adbe5f061a7f95139ab794bcfda7 and by the newer f36 container image?  If I'm wrong, please re-open this.

PR #474, opened 2022-05-13
shim: use SHIM_DEVEL_VERBOSE when built in devel mode

vathpela:
This makes SHIM_VERBOSE / SHIM_DEVEL_VERBOSE work the same way as
SHIM_DEBUG / SHIM_DEVEL_DEBUG when shim is built with ENABLE_SHIM_DEVEL
set.

### DIFF
--- a/lib/console.c
+++ b/lib/console.c
@@ -732,7 +732,7 @@ setup_verbosity(VOID)
 	UINTN verbose_check_size;
 
 	verbose_check_size = sizeof(verbose);
-	efi_status = get_variable(L"SHIM_VERBOSE", &verbose_check_ptr,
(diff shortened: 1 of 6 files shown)


lib/console.c
```diff
@@ -699,9 +699,15 @@ setup_verbosity(VOID)
 	EFI_STATUS efi_status;
 	UINT8 *verbose_check_ptr = NULL;
 	UINTN verbose_check_size;
+	const CHAR16 * const verbose_var_name =
+#ifdef ENABLE_SHIM_DEVEL
```
frozencemetery, line 703: I would personally do this as a #define rather than littering the variable initialization with ifdefs, but it's not wrong or anything.

frozencemetery (approved):

PR #476, opened 2022-05-17
Break out of the inner sbat loop if we find the entry.

vathpela:


sbat.c
```diff
@@ -171,11 +172,14 @@ verify_sbat_helper(list_t *local_sbat_var, size_t n, struct sbat_section_entry *
 	}
 
 	for (i = 0; i < n; i++) {
+		bool found = false;
```
julian-klode, line 175: Mmm this can go one block further inside?

vathpela: Sure.

PR #477, opened 2022-05-18
Misc fixups from scan-build.

vathpela:


frozencemetery (approved):
Coverity sure is a program.

PR #478, opened 2022-05-19
Fix preserve_sbat_uefi_variable() logic

jsetje:
preserve_sbat_uefi_variable() shoud really deal with the sbat
metadata version as a numerical value that could gain more digits.

It also needs to only compare the datestamp since the actual
metadata can grow and shrink

Signed-off-by: Jan Setje-Eilers <Jan.SetjeEilers@oracle.com>

### DIFF
--- a/sbat.c
+++ b/sbat.c
@@ -319,16 +319,53 @@ check_sbat_var_attributes(UINT32 attributes)
 #endif
 }
 
+static char *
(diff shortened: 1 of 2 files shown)


jsetje:
And now I got my test system into a strange state. Likely just bad test metadata, but need to confirm.

PR #388, opened 2021-07-12
Enable TDX measurement to RTMR register

kenplusplus:
Intel Trust Domain Extensions(Intel TDX) refers to an Intel technology
that extends Virtual Machine Extensions(VMX) and Multi-Key Total Memory
Encryption(MK-TME) with a new kind of virtual machine guest called a
Trust Domain(TD)[1]. A TD runs in a CPU mode that protects the confidentiality
of its memory contents and its CPU state from any other software, including
the hosting Virtual Machine Monitor (VMM).

Trust Domain Virtual Firmware (TDVF) is required to provide TD services to
the TD guest OS.[2] Its reference code is available at https://github.com/tianocore/edk2-staging/tree/TDVF.

To support TD measurement/attestation, TDs provide 4 RTMR registers like
TPM/TPM2 PCR as below:
- RTMR[0] is for TDVF configuration
- RTMR[1] is for the TD OS loader and kernel
- RTMR[2] is for the OS application
- RTMR[3] is reserved for special usage only

This patch adds TD Measurement protocol support along with TPM/TPM2 protocol.

References:
[1] https://software.intel.com/content/dam/develop/external/us/en/documents/tdx-whitepaper-v4.pdf
[2] https://software.intel.com/content/dam/develop/external/us/en/documents/tdx-virtual-firmware-design-guide-rev-1.pdf

Signed-off-by: Lu Ken <ken.lu@intel.com>

PR #479, opened 2022-05-22
[V2] Enable TDX measurement to RTMR register

kenplusplus:
Intel Trust Domain Extensions(Intel TDX) refers to an Intel technology
that extends Virtual Machine Extensions(VMX) and Multi-Key Total Memory
Encryption(MK-TME) with a new kind of virtual machine guest called a
Trust Domain(TD)[1]. A TD runs in a CPU mode that protects the confidentiality
of its memory contents and its CPU state from any other software, including
the hosting Virtual Machine Monitor (VMM).

Trust Domain Virtual Firmware (TDVF) is required to provide Intel TDX
implementation and service for EFI_CC_MEASUREMENT_PROTOCOL[2]. The bugzilla
for TDVF is at https://bugzilla.tianocore.org/show_bug.cgi?id=3625.

To support CC measurement/attestation with Intel TDX technology, following 4
RTMR registers will be extended by TDX service like TPM/TPM2 PCR:
- RTMR[0] is for TDVF configuration
- RTMR[1] is for the TD OS loader and kernel
- RTMR[2] is for the OS application
- RTMR[3] is reserved for special usage only

This patch adds TDX Implementation for CC Measurement protocol along with
TPM/TPM2 protocol.

References:
[1] https://software.intel.com/content/dam/develop/external/us/en/documents/tdx-whitepaper-v4.pdf
[2] https://software.intel.com/content/dam/develop/external/us/en/documents/tdx-virtual-firmware-design-guide-rev-1.pdf
[3] https://software.intel.com/content/dam/develop/external/us/en/documents/intel-tdx-guest-hypervisor-communication-interface-1.0-344426-002.pdf

Signed-off-by: Lu Ken <ken.lu@intel.com>


kenplusplus:
Sorry for later response, since wait for the review of similar grub's patch at https://www.mail-archive.com/grub-devel@gnu.org/msg34244.html. 

I resubmitted this V2 patch and close the previous one at https://github.com/rhboot/shim/pull/388:

1. Rebase from shim-15.4 to shim-15.5
2. Clean up unncessary comments for the CC structure to align with tpm1 & 2's structure.
3. Clean up the indentations to use tab

tpm.c
```diff
@@ -212,6 +257,8 @@ static EFI_STATUS tpm_log_event_raw(EFI_PHYSICAL_ADDRESS buf, UINTN size,
 EFI_STATUS tpm_log_event(EFI_PHYSICAL_ADDRESS buf, UINTN size, UINT8 pcr,
 			 const CHAR8 *description)
 {
+	cc_log_event_raw(buf, size, pcr, description, strlen(description) + 1,
```
kraxel, line 260: Error check missing here.

tpm.c
```diff
@@ -108,6 +108,51 @@ static EFI_STATUS tpm_locate_protocol(efi_tpm_protocol_t **tpm,
...
+	CopyMem(event->Event, (VOID *)log, logsize);
+	if (hash) {
+		efi_status = cc->hash_log_extend_event(cc,
+			EFI_CC_FLAG_PE_COFF_IMAGE, buf, (UINT64)size, event);
+	}
+
+	if (!hash || EFI_ERROR(efi_status)) {
+		efi_status = cc->hash_log_extend_event(cc,
+			0, buf, (UINT64)size, event);
+	}
```
kraxel, line 151: I think it would be better to explicitly pass down the information whenever the data we are hashing is supposed to be a PE binary or not.  This trial and error game here doesn't look like a good idea from a security point of view.

kraxel (changes requested):

tpm.c
kenplusplus: Thanks for recommend, yes, different with TPM1/TPM2. CC will handle https://github.com/tianocore/edk2-staging/blob/8126186bb1bd484a4c8d731cd9702ad22547797f/OvmfPkg/Tcg/Tcg2Dxe/TdTcg2Dxe.c#L1326. Will update according to your comment

tpm.c
kenplusplus: Fixed at https://github.com/rhboot/shim/compare/d00d575db075e4efd6e1ca431db939aa5cfcae1e..772ab3c708276eaa0873d2f3b787f3ac1439aa4f please help review, thanks!

tpm.c
kraxel: When (1) fails the code will re-try (2) though.  I think this should be something along the lines of ...
```
flags = hash ? EFI_CC_FLAG_PE_COFF_IMAGE : 0;
efi_status = cc->hash_log_extend_event(cc, flags, buf, (UINT64)size, event);
```
... instead.

Also the hash parameter could be named better, could be `BOOLEAN is_pe_image` for example.

kraxel (changes requested):

tpm.c
kenplusplus: Thanks for the suggestions on clean code. I have updated at https://github.com/rhboot/shim/compare/772ab3c708276eaa0873d2f3b787f3ac1439aa4f..065153bc4a143fa7f8c22a8682c5c36b6a33d094 please help review again.

I still keep the parameter "hash" since it is still requied for the implementation with TPM 1.2, please check https://github.com/rhboot/shim/blob/75449bcab1aad6f32046531a5b47ed3dc14cea6b/tpm.c#L191 , or do you have better suggestion to keep parameter hash and is_pe?

include/cc.h
```diff
@@ -0,0 +1,81 @@
...
+  uint32_t HeaderSize;
+  uint16_t HeaderVersion;
+  EFI_CC_MR_INDEX MrIndex;
+  uint32_t EventType;
+} __attribute__ ((packed)) EFI_CC_EVENT_HEADER;
+
+typedef struct tdEFI_CC_EVENT {
+  uint32_t Size;
+  EFI_CC_EVENT_HEADER Header;
+  uint8_t Event[1];
```
kraxel, line 45: If you make this `Event[0]` ...

tpm.c
```diff
@@ -108,6 +108,45 @@ static EFI_STATUS tpm_locate_protocol(efi_tpm_protocol_t **tpm,
...
+				       (VOID **)&cc);
+	if (EFI_ERROR(efi_status) || cc == NULL)
+		return EFI_SUCCESS;
+
+	efi_status = cc->map_pcr_to_mr_index(cc, pcr, &mr);
+	if (EFI_ERROR(efi_status))
+		return EFI_NOT_FOUND;
+
+	UINTN event_size = sizeof(*event) - sizeof(event->Event) + logsize;
+
```
kraxel, line 130: ... the math here becomes simpler here.

(didn't we have the same discussion on the grub patch already?)

tpm.c
```diff
@@ -212,6 +251,12 @@ static EFI_STATUS tpm_log_event_raw(EFI_PHYSICAL_ADDRESS buf, UINTN size,
 EFI_STATUS tpm_log_event(EFI_PHYSICAL_ADDRESS buf, UINTN size, UINT8 pcr,
 			 const CHAR8 *description)
 {
+	EFI_STATUS efi_status;
+	efi_status = cc_log_event_raw(buf, size, pcr, description,
+				      strlen(description) + 1, EV_IPL, NULL);
+	if (EFI_ERROR(efi_status))
```
kraxel, line 257: Hmm, so `cc_log_event_raw()` is never called with hash != NULL.

Looks like we are missing something here, I guess the `tpm_log_event() ` callers should pass down the information whenever they log a PE binary or config data?

kraxel (changes requested):

tpm.c
kenplusplus: I am dilemma since TPM structure has similiar issue at https://github.com/rhboot/shim/blob/75449bcab1aad6f32046531a5b47ed3dc14cea6b/include/tpm.h#L39 and https://github.com/rhboot/shim/blob/75449bcab1aad6f32046531a5b47ed3dc14cea6b/include/tpm.h#L138 Do you want I fixed TPM and CC protocol at same time? At least, they should be consistent I think. What about your thoughts?

tpm.c
kraxel: Oh, *this* is where the hash argument comes from.  I wasn't aware of the tpm 1.2 vs. tpm 2.0 difference.  In any case the hash variable is not properly passed through so the code surely doesn't work for the tpm 1.2 case.

I'm wondering whenever the tpm 1.2 case matter in the first place?  Or can we ignore that, support tpm 2.0 only and simply skip the `cc_log_event_raw()` call in case `buf == NULL && size == 0`?  And remove some dead code?

kraxel: (if ignoring tpm 1.2 isn't an option the code surely deserves some comments explaining this and indicating which code path is taken for the 1.2 case and which for the 2.0 case).

tpm.c
kraxel: (after looking at the complete tpm.c file).  There are three places calling `tpm_log_event_raw()`.  So, yes, either call `cc_log_event_raw()` from `tpm_log_event_raw()`, which indeed looks easiest, or add a `cc_log_event_raw()` call to all places calling `tpm_log_event_raw()`.

kraxel: Oh, I see where the "try again without PE_COFF flag" logic comes from, `tpm_log_event_raw()` does that too.
Would be interesting to know why the shim tpm code does that, it still looks wrong to me.

tpm.c
kenplusplus: Sorry, I am not very clear what your recommendation:
1. Is this OK? https://github.com/rhboot/shim/compare/6825122515190866c120cd296cc92ad04418a548..56621679cd862213211d794ba6c67df0681c5b20 
2. tpm 1.2 is still important and popular, so I think we could not skip it
3. buf, size, log are required for CC, TPM1, TPM2 cases.
4. I added some comments to clarify in above commit, could you please help review again.

Thanks!

tpm.c
kenplusplus: > Oh, I see where the "try again without PE_COFF flag" logic comes from, `tpm_log_event_raw()` does that too. Would be interesting to know why the shim tpm code does that, it still looks wrong to me.

Unlike CC measurement via Intel TDX or AMD SEV, TPM measurement is done by hardware device, so I think some of hardwares may exist bug to support PE_COFF. that is why fallback code.

kraxel:
Latest version looks fine to me.

tuanggolt:
ok

frozencemetery:
>  Do you need me rebase this patch to 15.6?

Yes - your patch is targeted for main, so it should be based on main.

kenplusplus:
@frozencemetery Thanks! I resubmitted V3 PR https://github.com/rhboot/shim/pull/485 , targeting to main branch. Could we close this PR and move to new V3 PR?

frozencemetery:
Yes.  Couple notes though:

- I believe you as the person who opened the PR are able to close your own PRs.
- You don't need to open a new PR - in the future you can just `git push --force` the branch.

PR #480, opened 2022-05-23
Minor fixups from cov-scan and similar.

vathpela:


frozencemetery (approved):
Don't see anything wrong here.

jsetje:
FWIW, this all looks good to me as well.

PR #481, opened 2022-05-24
SBAT Policy latest should be a one-shot

jsetje:
Since booting from removable media can be hard to detect,
setting a persistent latest SBAT policy is risky in a typical
client system. This changes latest to be a one-shot operation
that could be set at the time of an OS update if desired.

### DIFF
--- a/sbat.c
+++ b/sbat.c
@@ -369,6 +369,16 @@ preserve_sbat_uefi_variable(UINT8 *sbat, UINTN sbatsize, UINT32 attributes,
 	return false;
 }
 
+static void
+clear_sbat_policy()
+{
+	EFI_STATUS efi_status = EFI_SUCCESS;
+
+	efi_status = del_variable(SBAT_POLICY, SHIM_LOCK_GUID);
+	if (EFI_ERROR(efi_status))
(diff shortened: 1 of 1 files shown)


sbat.c
```diff
@@ -394,33 +404,29 @@ set_sbat_uefi_variable(void)
...
+				clear_sbat_policy();
 				break;
 			case SBAT_POLICY_PREVIOUS:
 				dprint("Custom sbat policy: previous\n");
 				sbat_var = SBAT_VAR_PREVIOUS;
 				break;
 			case SBAT_POLICY_RESET:
 				if (secure_mode()) {
-					console_print(L"Cannot reset SBAT policy: Secure Boot is enabled.\n");
+					console_print(L"Cannot reset SBAT policy: "
```
vathpela, line 415: Don't break strings across lines for formatting, leave them intact so they're easy to grep for.

vathpela (changes requested):
Just one minor change, otherwise lgtm.

vathpela:
I've pushed the fix as https://github.com/rhboot/shim/commit/77144e5a404df89b45941bfc54fd2f59e0ee607b

PR #483, opened 2022-05-31
Make SBAT variable payload introspectable

chrisccoulson:
This is for after 15.6, but we'll probably add something like this to
our next shim in Ubuntu, so I'd at least like to agree on the correct
way to do this.

Given a set of EFI variables and boot assets, it should be possible
to compute what the value of PCR 7 will be on the next boot.

As shim manages the contents of the SbatLevel variable and this is
measured to PCR 7, export the payloads that shim contains in a new
COFF section (.sbatlevel) so that it can be introspected by code
outside of shim.

The new section works a bit like .vendor_cert - it contains a header
and then the payload. In this case, the header contains no size fields
because the strings are NULL terminated. Shim uses this new section
internally in set_sbat_uefi_variable.

The .sbatlevel section starts with a 4 byte version field which is
not used by shim but may be useful for external auditors if the
format of the section contents change in the future.


julian-klode:
:+1:

sbat_var.S
```diff
@@ -0,0 +1,46 @@
+// SPDX-License-Identifier: BSD-2-Clause-Patent
+
+# include "include/sbat_var_defs.h"
+
+#if defined(ENABLE_SHIM_DEVEL)
```
frozencemetery, line 5: Wouldn't it be better to have this stuff in sbat_ver_defs.h?

sbat_var.S
```diff
@@ -0,0 +1,46 @@
+// SPDX-License-Identifier: BSD-2-Clause-Patent
+
+# include "include/sbat_var_defs.h"
```
frozencemetery, line 3: Extra whitespace between the '#' and "include".

sbat_var.S
```diff
@@ -0,0 +1,46 @@
...
+	SBAT_VAR_PREVIOUS_REVOCATIONS
+
+#define SBAT_VAR_LATEST_DATE SBAT_VAR_ORIGINAL_DATE
+#define SBAT_VAR_LATEST_REVOCATIONS
+#define SBAT_VAR_LATEST \
+	SBAT_VAR_SIG SBAT_VAR_VERSION SBAT_VAR_LATEST_DATE "\n" \
+	SBAT_VAR_LATEST_REVOCATIONS
+#endif /* ENABLE_SHIM_DEVEL */
+
+        .section .sbatlevel, "a", %progbits
```
frozencemetery, line 31: Indentation from here down seems wonky.  So far for asm files (i.e., cert.S) we've used hard tabs only.  Some of these lines are using tabs, some spaces.

frozencemetery (changes requested):
Some minor stuff inline.

frozencemetery:
(Also, this needs rebased at some point - conflicts in sbat.h.)

sbat_var.S
chrisccoulson: I've fixed this to just use tabs now.

frozencemetery (approved):
Thanks, no further objections.

Issue #376, opened 2021-06-03
objcopy for SBAT doesn't work?


TBOpen:
I'm now building the GRUB2 versions and when I use:

_objcopy --set-section-alignment '.sbat=512' --add-section .sbat=sbat.csv foo.efi_

the resulting GRUB cannot be booted by shim because it says:

**Section 0 is inside image headers
Malformed section header
Failed to load image: Unsupported
start_image() returned Unsupported**

Is there a particular version of objcopy needed?
I noticed it puts the section in the beginning instead of end of the PE like shim itself.

Is there something else needed?

Update:

It appears because the .sbat virtual address is created as 0.

TBOpen:
I see shim used a different command line which I tried same issue either way.   I was able to hack around it via:

> objcopy --add-section .sbat=mycsvfile.csv --set-section-flags .sbat=contents,alloc,load,readonly,data grubx64.efi
> objcopy --adjust-vma 4096 grubx64.efi

However, it won't load, it just hangs the system when attempting to boot.  If I don't touch grubx64.efi with the objcopy then it works fine.

This one works fine:

```
Dump of file: grubx64.efi, type: COFF64
Dump of PE/COFF file grubx64.efi
-----------------------------------------------
File size: 1062664
File header:
Machine: x86-64
TimeDate: 0x54A48E00 - Wed Dec 31 16:00:00 2014

Number of sections:  4
Number of symbols:   0
Optional header size: 240
Flags: 0x020E
Symbol table offset: 0x0
String table offset: 0x0
Section headers offset: 0x104

Optional header:
Magic number: 0x20B
Size of code: 0xB000
Size of uninitialized data: 0xF6000
Address of entry point: 0x1000
Base of code: 0x1000
Image base: 0x0000000000000000
Section alignment: 0x1000
File alignment: 0x1000
Size of image: 0x103000
Size of headers: 0x1000
Dll characteristics: 0x0
Size of stack reserve: 0x0000000000010000
Size of stack commit: 0x0000000000010000
Size of heap reserve: 0x0000000000010000
Size of heap commit: 0x0000000000010000
Directory  5, Base_relocation_table:
  Address 0x102000, Size 0x1000, Section 4, Offset 0x0000

 1 Section .text
Virtual size: 0xB000
Virtual address: 0x1000
Size of raw data: 0xB000
Raw data pointer: 0x1000
Characteristics: Text, Executable, Readable

 2 Section .data
Virtual size: 0x10000
[...]

PR #491, opened 2022-07-18
Improve adding sbat section example in SBAT.md

bernhardreiter:
* Add `--adjust-section-vma .sbat+10000000` to the `objcopy` example command to make sure
   to add the .sbat section to the end of the binary and avoid problems like #376


vathpela:
Please don't push merges into the PR, it makes it impossible to figure out what's going on.

bernhardreiter:
@vathpela I usually look at the diff (which is very short for this pull-request) and had hoped that merging changes in from main would make acceptance easier. (Won't do it again in this repo.)

jsetje:
@bernhardreiter I've looked at this twice now, and it seems like this is something you had to do due to something specific to your binary. I'm not sure this makes sense as a general recommendation.

bernhardreiter:
> @bernhardreiter I've looked at this twice now, 

@jsetje thanks for having taken the time! 

> and it seems like this is something you had to do due to something specific to your binary. I'm not sure this makes sense as a general recommendation.

Being no expert, I do not know enough to decide this. From the descriptions in #376, it seems that I wasn't the only one with the issue and some `objcopy` binaries may need this option. I have suggesting it, because it seems harmless to give the option just in case someone has an `objcopy` binary that needs it.

vathpela:
I don't think this is an ongoing issue, so I'm closing this.

PR #499, opened 2022-08-08
a shim tool running in UEFI shell

dennis-tseng99:

To help configuring for shim.efi running in UEFI shell, a shim utility is created so called shimutil.efi. This tool now can:

1. get secure boot state (enable or disable)
2. get/set sbat mode (lastest/previous/delete)
3. get key's content briefly

Due to more flexible software architecture, it is easy to add functions into shimutil.c


frozencemetery (changes requested):
Thanks for your interest.

I'm not really seeing the use case for this in general.  For inclusion, we'd need to see not only a use case, but justification for why it needs to be in shim itself and can't be a standalone binary.  We do generally want shim to be reasonably simple for security.  But perhaps I've missed something and you can clarify.

Also, for this and future PRs, please make an effort to match code to existing style.  For commit messages, you may also find [this post](https://cbea.ms/git-commit/) a helpful reference for this and many other projects.

dennis-tseng99:
Thanks for your comments. I will revise my commit messages, and put more reasons why we need this. Many thanks.
________________________________

Issue #498, opened 2022-08-05
GRUB crashing due to apparently stale instruction cache


dannf:
We're seeing intermittent (~1/100) Synchronous Exceptions in early GRUB code on an ARM Cortex-A72-based platform. GRUB is being loaded by `shim` (Secure Boot is disabled fwiw). Curiously, the  instruction $IP points to never seems like something that should cause an exception. One theory was that the instruction cache contained stale content, which seemed to be confirmed by manually flushing the cache within a debugger after hitting the exception, and re-executing the same code. This worked. Our working theory is that `shim` should be flushing the I-cache after loading the GRUB binary. (This [story](https://www.mail-archive.com/osv-dev@googlegroups.com/msg06203.html) sounds pretty similar, btw).

We added the following code to a test shim build, and the test is no longer failing. Multiple machines have survived over 1000 reboots:

```
--- pe.c~       2022-07-28 23:31:51.399506464 +0000
+++ pe.c        2022-07-28 23:51:25.398746245 +0000
@@ -963,6 +963,7 @@
 
        CopyMem(buffer, data, context.SizeOfHeaders);
 
+       __builtin___clear_cache((char *)buffer, (char *)(buffer + context.ImageSize));
        *entry_point = ImageAddress(buffer, context.ImageSize, context.EntryPoint);
        if (!*entry_point) {
                perror(L"Entry point is invalid\n");
```

dannf:
> Thank you for debugging into this. Could you please open this as a PR so we can review it for inclusion?

Thanks for the reply @frozencemetery. What I have above is only a POC, not something I want to propose for merging. I'm working on a better implementation for which I'll submit a PR when ready.

PR #504, opened 2022-08-19
shim: Flush the memory region from i-cache before execution

dannf:
…ing it

We've seen crashes in early GRUB code on an ARM Cortex-A72-based platform
that point at seemingly harmless instructions. Flushing the i-cache of
those instructions prior to executing seems to work fine, which seems to
have parallels with this story:
  https://www.mail-archive.com/osv-dev@googlegroups.com/msg06203.html

Add a cache flushing utility function that is a no-op for !arm64, and
and provide an implementation using a GCC intrinsic.

This fixes issue #498.

### DIFF
--- a/pe.c
+++ b/pe.c
@@ -5,6 +5,7 @@
  */
 
 #include "shim.h"
+#include "cache.h"
 
 #include <openssl/err.h>
 #include <openssl/bn.h>
@@ -1195,6 +1196,9 @@ handle_image (void *data, unsigned int datasize,
 			 MEM_ATTR_X);
 
 	CopyMem(buffer, data, context.SizeOfHeaders);
+
+	/* Flush the instruction cache for the region holding the image */
+	cache_invalidate(buffer, buffer + context.ImageSize);
 
 	*entry_point = ImageAddress(buffer, context.ImageSize, context.EntryPoint);
 	if (!*entry_point) {
--- a/Makefile
+++ b/Makefile
@@ -40,7 +40,7 @@ TARGETS += $(MMNAME) $(FBNAME)
 endif
 OBJS	= shim.o globals.o mok.o netboot.o cert.o replacements.o tpm.o version.o errlog.o sbat.o sbat_data.o sbat_var.o pe.o httpboot.o csv.o load-options.o
 KEYS	= shim_cert.h ocsp.* ca.* shim.crt shim.csr shim.p12 shim.pem shim.key shim.cer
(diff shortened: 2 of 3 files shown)


frozencemetery (changes requested):
Thanks for the patch.  I've a comment inline... since I'm asking for changes already, please also fixup the line lengths in your commit messages (I know we're not always great about this, but it's nicer to have 72 in body and not much over 50 in subject).

pe.c
```diff
@@ -1196,6 +1197,9 @@ handle_image (void *data, unsigned int datasize,
 
 	CopyMem(buffer, data, context.SizeOfHeaders);
 
+	/* Flush the instruction cache for the region holding the image */
+	cache_invalidate(buffer, buffer + context.ImageSize);
```
frozencemetery, line 1201: This line and the line above suggest to me that the invalidation always occurs, while that's not what the function does.

frozencemetery (changes requested):

pe.c
dannf: Thanks for the review @frozencemetery. Along with correcting the comment, what do you think about renaming the function "maybe_cache_invalidate()" or "perhaps "arch_cache_invalidate()"?

dannf: I've pushed a version w/ the `arch_cache_invalidate()` name, but open to other suggestions.

dannf:
> Thanks for the patch. I've a comment inline... since I'm asking for changes already, please also fixup the line lengths in your commit messages (I know we're not always great about this, but it's nicer to have 72 in body and not much over 50 in subject).

Also done, thanks for the feedback!

frozencemetery (approved):

dannf:
I've pushed a new version due to feedback at https://github.com/rhboot/grub2/pull/107#issuecomment-1232205900

- It no longer has an architecture guard; we rely on the compiler to decide if `__builtin___clear_cache()` should be a no-op on that architecture (GCC does so on x86)
- I renamed `arch_cache_invalidate` back to `cache_invalidate` because we now compile it everywhere - the compiler just may decide it doesn't need to generate code for it.

fyi, I tested with clang as well, and found that it does define `__GNUC__`, and does apparently implement `__builtin___clear_cache()`, though shim fails to build with clang for other reasons.

vathpela:
I've pushed this as 5c537b3d0cf8c393dad2e61d49aade68f3af1401, which is just organized a bit differently.

Issue #95, opened 2017-09-08
What happens if LoadOption is L"\0" ?


edjee:
https://github.com/rhboot/shim/blob/e06765ae0dbbc874ff9e04dbd0e7d82989048261/shim.c#L2531

Hi @vathpela , may I ask about the function "set_second_stage" in shim.c ?

If the LoadOption data happens to be L"\0" , what's the expected results?
I found that the function "count_ucs2_strings" returns 1, when the input is L"\0".

In such a case, do we want to fall back to DEFAULT_LOADER? Or halt?

edjee:
To me it's strange that count_ucs2_strings() returns 1 for the input L"\0".
Is that an expected behavior? Shouldn't it return 0?

vathpela:
L"\0" is one string of length 0.  That looks right to me.
I think we want DEFAULT_LOADER in this case.

PR #505, opened 2022-08-23
Discard load-options that start with a NUL

frozencemetery:
In 6c8d08c0af4768c715b79c8ec25141d56e34f8b4 ("shim: Ignore UEFI
LoadOptions that are just NUL characters."), a check was added to
discard load options that are entirely NUL.  We now see some firmwares
that start LoadOptions with a NUL, and then follow it with garbage (path
to directory containing loaders).  Widen the check to just discard
anything that starts with a NUL.

Resolves: #490
Related: #95
See-also: https://bugzilla.redhat.com/show_bug.cgi?id=2113005

### DIFF
--- a/include/ucs2.h
+++ b/include/ucs2.h
@@ -63,22 +63,4 @@ StrCSpn(const CHAR16 *s, const CHAR16 *reject)
 	return ret;
 }
 
-/*
- * Test if an entire buffer is nothing but NUL characters.  This
- * implementation "gracefully" ignores the difference between the
- * UTF-8/ASCII 1-byte NUL and the UCS-2 2-byte NUL.
- */
-static inline bool
-__attribute__((__unused__))
-is_all_nuls(UINT8 *data, UINTN data_size)
-{
-	UINTN i;
-
-	for (i = 0; i < data_size; i++) {
-		if (data[i] != 0)
-			return false;
-	}
-	return true;
-}
-
 #endif /* SHIM_UCS2_H */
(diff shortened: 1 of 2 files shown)


julian-klode:
I thought my fix went in in 15.5 or so to make it ignore load options when running from removable media path so I'm confused how that happens; or confused re having applied the patch

frozencemetery:
While it's always possible that my ability to read m4 is failing, your changes were gated behind DISABLE_REMOVABLE_LOAD_OPTIONS being defined, and by default it's not.

frozencemetery:
More generally, that kind of divergence is why I don't like having things gated by options in shim.  I think there's a case to be made that the behavior of DISABLE_REMOVABLE_LOAD_OPTIONS should be the default, or at least that DISABLE_REMOVABLE_LOAD_OPTIONS should be default-on, if you'd prefer that.

julian-klode:
Ah right, sorry. We enabled that in Ubuntu and I feel like that's sensible as it avoids these issues entirely, but @vathpela had some concerns IIRC.

There were a whole bunch more of these issues but we stopped debugging that after disabling it.

dennis-tseng99:
It could be straightforward if changing UINT16 to CHAR16 although both are uint16_t :
```if (((UINT16 *)li->LoadOptions)[0] == 0)``` -->
```if (((CHAR16 *)li->LoadOptions)[0] == 0)```

frozencemetery:
That's fine, updated.

frozencemetery:
Thanks for the testing.  For those hitting this issue before their distros update the officially signed shim: in my opinion, it is better from a security perspective to run the older, signed shim rather than disabling secureboot to run the patched shim above from #505.  

(If you're able to enroll your own keys, of course you can have the best of both worlds; and if you're just testing fixes, none of this applies anyway.)

PR #511, opened 2022-09-05
patch openssl CVE-2022-0778

j30031148:
This CVE is from OpenSSL， shim involves this function.
Signed-off-by:JinLun jinlun@huawei.com


frozencemetery:
When we investigated this CVE, we determined that shim was unaffected due to not using EC crypto.  Do you have evidence that this does affect shim?

Note that it is not our policy to update the bundled openssl every time openssl has a CVE - we only update when there's one that affects shim.

Issue #429, opened 2021-11-03
Obsolete fallback.efi


vathpela:
`fallback.efi` as a separate binary creates a pile of issues:
- with tpm enabled, the fallback path has to reboot, so that the new boot option doesn't have fallback in its history of PCR[7] values
- that means we need to detect boot loops
- if the reason fallback has been run is that variables are defective, as is fairly often the case, detecting boot loops won't work

If `fallback` is a function rather than a separate executable, things are better:
- we can go back to just directly launching whatever boot entry we select
- we no longer have to deal with fallback in the path mangling in `set_second_stage()`
- fallback still succeeds when boot variables are horribly broken
- fallback no longer needs to exist in the make files

There are a things we need to ensure when doing this as well:
- must have a well defined and documented way of deploying shim on removable media/recovery media/cloud that *doesn't* invoke fallback.  I'm thinking the right answer here is probably the presence of a second stage bootloader in the current working directory.

(Previously discussed here: https://github.com/rhboot/shim/issues/418#issuecomment-942635768 )

PR #512, opened 2022-09-15
Add way to avoid reboots with a tpm.

GZibrat:
On GCE we have the TPM enabled by default so we hit this reboot on every VM start. Some VM types have a very expensive reboot so we'd like to avoid this reboot if possible.

The mechanism here does that by not having the fallback change the boot order so that it always remains in the boot path. This should keep TPM measurements stable to make the reboot not needed.

author: Gary Zibrat <gzibrat@google.com>

### DIFF
--- a/fallback.c
+++ b/fallback.c
@@ -6,8 +6,24 @@
 #include "shim.h"
 
 #define NO_REBOOT L"FB_NO_REBOOT"
+#define NO_WRITE_BOOT_ORDER L"FB_NO_WRITE_BOOT_ORDER"
 
 EFI_LOADED_IMAGE *this_image = NULL;
+
+
+static UINT32 get_fallback_no_write_boot_order() {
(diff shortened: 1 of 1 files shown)


julian-klode:
We probably should avoid this workaround, options are bad (who tests them, why do systems behave differently, etc), and the right approach to this is integrating the fallback into shim directly which is already planned.

GZibrat:
https://github.com/rhboot/shim/issues/429

Perhaps this is the issue to track?

I am happy to work on it if it doesn't seem terribly complicated or if I can get some pointers on the general idea.

GZibrat:
@vathpela 

Any answers or thoughts on this?

vathpela:
Hopefully this branch https://github.com/vathpela/mallory/tree/merge-mm-and-fb-into-shim (still very much a work in progress) should address the issue for you?  But yes, this is issue #429

PR #514, opened 2022-10-01
Disable exporting vendor-dbx to MokListXRT

dennis-tseng99:
As the vendor-dbx grows, it caused some problems when writing such a large variable. Some firmwares lie the avaiable space(*1) , and some even crash(*2) for no good reason after the writing of MokListXRT. Both shim and kernel don't rely on MokListXRT to block anything, so we just stop exporting vendor-dbx to MokListXRT to avoid the potential hassles.

(*1) https://bugzilla.suse.com/show_bug.cgi?id=1185261 
(*2) https://github.com/rhboot/shim/pull/369#issuecomment-855275115

### DIFF
--- a/mok.c
+++ b/mok.c
@@ -118,8 +118,10 @@ struct mok_state_variable mok_state_variable_data[] = {
 		     EFI_VARIABLE_NON_VOLATILE,
 	 .no_attr = EFI_VARIABLE_RUNTIME_ACCESS,
 	 .categorize_addend = categorize_deauthorized,
+#ifndef DISABLE_EXPORT_DBX
 	 .addend = &vendor_deauthorized,
 	 .addend_size = &vendor_deauthorized_size,
+#endif
 	 .flags = MOK_MIRROR_KEYDB |
 		  MOK_MIRROR_DELETE_FIRST |
 		  MOK_VARIABLE_LOG,


frozencemetery (changes requested):
This PR introduces a new configuration option.  I dislike this in general: where possible, we should figure out whether "defined" or "undefined" is correct, and not have it configurable.  (In this specific case, it's worse because the define is secret - there's no documentation or mention of it other than reading the source.)

I think there's a conversation to be had here on whether handling the write error or disabling exporting it altogether are correct and others beyond me will I hope weigh in on that.

dennis-tseng99:
A configuration option is added in mok.c to stop exporting. For those vendors who do not have such issue will feel nothing changed. On the other hand, if vendors want to stop exporting, then the DISABLE_EXPORT_DBX must be defined in Makefile and shim.spec command-line respectively. For example:
In Makefile:
CFLAGS += -DDISABLE_EXPORT_DBX
In shim.spec:
make RELEASE=0 ... DISABLE_EXPORT_DBX=1 ... shim.efi

A README.compile will be added soon.

julian-klode:
Basically anyone with a larger vendor dbx needs this. We've been shipping this without a config option in Ubuntu for a while and I think we gotta continue despite us switching from 20 KB of hashes to like 9KB of certs - it's still to big.

The Linux kernel can revoke via certificate hashes, which would significantly reduce the size requirement, but I'm not sure how one would represent that.

vathpela:
I don't really like the implementation here - I'd rather see a mok_state_variable flag to say to mirror it using config table space but not variable space, rather than simply not adding the data we have.

PR #520, opened 2022-10-24
Make the names of EFI binaries arch-independent

dennis-tseng99:
Since we only build the 64-bit binaries, we don't have the issue of the mixed architecture binaries in the same directory. Besides, we will use the same install script for x86_64 and AArch64. It's easier to maintain the script with the same names.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -69,6 +69,10 @@ ifneq ($(origin FALLBACK_VERBOSE_WAIT), undefined)
 	CFLAGS += -DFALLBACK_VERBOSE_WAIT=$(FALLBACK_VERBOSE_WAIT)
 endif
 
+ifneq ($(origin ARCH_INDEPEND),undefined)
+	CFLAGS += -DARCH_INDEPEND
+endif
+
 all: confcheck $(TARGETS)
 
 confcheck:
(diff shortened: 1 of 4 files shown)


julian-klode:
There's plenty of folks building 32-bit ones.

dennis-tseng99:
@julian-klode 
Thanks for your comment. I added a compiled flag(ARCH_INDEPEND) in Makefile, shim.h, shim.c and fallback.c to let user make his/her final decision.

frozencemetery (changes requested):
Adds a configuration option, -1.

julian-klode:
I think what we could do is try arch specific first and then the independent ones. But also I'd prefer not to have arch-independent shim binary named out there.

The moment you do it, product managers will appear and request an installer  image that runs on both amd64 and arm64.😜

steve-mcintyre:
> I think what we could do is try arch specific first and then the independent ones. But also I'd prefer not to have arch-independent shim binary named out there.
> 
> The moment you do it, product managers will appear and request an installer image that runs on both amd64 and arm64.stuck_out_tongue_winking_eye

Agreed - this is a really bad mistake IMHO.

julian-klode:
I think we're going to close this as we don't want this as there's lots of opposition here.

PR #527, opened 2022-11-14
make-archive: Build reproducible tarball

julian-klode:
Remove timestamps, user names, etc. from the tarball so that it can be built reproducibly by multiple people, on different machines.

The outer bzip2 layer might still be different, no reproducible bzip2 known.

This sets the time to 0, it would also be possible to set it to the git tag time.

### DIFF
--- a/make-archive
+++ b/make-archive
@@ -86,14 +86,16 @@ main() {
 	cd ..
 	if [ "x" = "x${SHIM_GIT_TAG}" ] ; then
 		git archive --format=tar "$(git log -1 --pretty=format:%h)" | ( cd "${ARCHIVE_DIR}/shim-${VERSION}" ; tar x )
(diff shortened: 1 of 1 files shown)


steve-mcintyre:
LGTM, but as you suggest let's take the mtime from tag time

julian-klode:
Updated. So right now the tar logic above actually extracts the right timestamps based on the git tags, by extracting the git-archive archives, so shim and gnu-efi have different timestamps - which is nice to have, but not doable by setting the time on the command-line, so this uses the shim tag timestamp - if it exists, for everything.

steve-mcintyre:
looks better, cool!

frozencemetery:
Hmm, I thought bzip2 was reproducible, but your commit suggests otherwise - am I misremembering?

PR #528, opened 2022-11-16
Shim 15.7 version update

vathpela:
Final changes for shim-15.7 (not including tagging)
- bump grub's required SBAT_LEVEL version
- bump shim's .sbat version
- minor coverity related changes

### DIFF
--- a/include/sbat_var_defs.h
+++ b/include/sbat_var_defs.h
@@ -3,6 +3,9 @@
 #ifndef SBAT_VAR_DEFS_H_
 #define SBAT_VAR_DEFS_H_
 
+/*
+ * This is the entry for the sbat data format
+ */
 #define SBAT_VAR_SIG "sbat,"
 #define SBAT_VAR_VERSION "1,"
 #define SBAT_VAR_ORIGINAL_DATE "2021030218"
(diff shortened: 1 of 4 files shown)


include/sbat_var_defs.h
```diff
@@ -22,14 +25,14 @@
 	SBAT_VAR_SIG SBAT_VAR_VERSION SBAT_VAR_LATEST_DATE "\n" \
 	SBAT_VAR_LATEST_REVOCATIONS
 #else /* !ENABLE_SHIM_DEVEL */
-#define SBAT_VAR_PREVIOUS_DATE SBAT_VAR_ORIGINAL_DATE
-#define SBAT_VAR_PREVIOUS_REVOCATIONS
+#define SBAT_VAR_PREVIOUS_DATE "2022052400"
+#define SBAT_VAR_PREVIOUS_REVOCATIONS "shim,2\ngrub,2\n"
```
julian-klode, line 29: We said we don't bump previous to shim,2 because most folks (including Ubuntu, SUSE, openSUSE) don't have a shim,2 yet and would end up unbootable.

julian-klode: Maybe add a comment stating so above it.

julian-klode (changes requested):

include/sbat_var_defs.h
vathpela: So you're suggesting `"grub,2\n"`, or `"shim1,\ngrub,2\n"`?

julian-klode: I only tested `"grub,2\n"` (and @jsetje had the same in the end) and we didn't have `shim,1\ngrub,1\n` before so the long one would be a bit inconsistent.

jsetje: They are functionally equivalent, but "grub,2\n" is the more compact way of stating it.

vathpela: Fixed in https://github.com/rhboot/shim/pull/528/commits/f38e63aa61deb926d51cf07d05a7a2e140d1be58.

julian-klode (approved):
:tada:

jsetje (approved):

PR #531, opened 2022-11-24
Add validation function for Microsoft signing

dennis-tseng99:
A validation function is added to post-process-pe.c to verify 3 flags required by Microsoft signing right after each EFI file has been built.

Like other functions existing in post-process-pe.c, this function can be enabled through the "-m" flag by CLI command. Default is enabled.

The output is something like this:
./post-process-pe -vv shim.efi
set_dll_characteristics():358: Updating DLL Characteristics from 0x0000 to 0x0100 ms_validation():373: NX-Compat-Flag: PASS
ms_validation():378:   4K-Alignment: PASS
ms_validation():390: Section-Wr-Exe: PASS

Microsoft is enforcing the validationis for all applications to be signed by the Microsoft third-party Unified Extensible Firmware Interface (UEFI) Certificate Authority (CA). These validations include:

- Section Alignment of the submitted PE file must be aligned with page size. This must be 4kb.
- Section Flags must not combine IMAGE_SCN_MEM_WRITE and IMAGE_SCN_MEM_EXECUTE for any given section.
- DLL Characteristics must include IMAGE_DLLCHARACTERISTICS_NX_COMPAT

Please refer the following for more details:
https://techcommunity.microsoft.com/t5/hardware-dev-center/new-uefi-ca-memory-mitigation-requirements-for-signing/ba-p/3608714

### DIFF
--- a/post-process-pe.c
+++ b/post-process-pe.c
@@ -43,6 +43,7 @@ static int verbosity;
 	})
 
 static bool set_nx_compat = false;
+static bool set_ms_validation = true;
 
 typedef uint8_t UINT8;
 typedef uint16_t UINT16;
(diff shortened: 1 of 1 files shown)


julian-klode (changes requested):
static storage modifier has been removed from variables and functions. This is (a) an unrelated change and (b) wrong, they are static (this also removes the need for the = 0 initializer at the top).

dennis-tseng99:
Thank Julian's comment. About:
(a) I agree. I will put "static" back.
(b) I don't agree. The initial value of non-static global variable in .bss is also 0. I put  0 to it is just for easy to read code.

frozencemetery (changes requested):
Changing the nx compat default is a separate PR.  Either drop the change or depend on that commit.

dennis-tseng99:
Thank Frozencemetery's comment. Yes, you are right. The nx compat default is a separate PR. I will drop the NX change.

PR #537, opened 2022-12-16
Drop invalid calls to `CRYPTO_set_mem_functions`

nicholasbishop:
These calls did not check the return value. If they had, it would have shown that the calls were failing due to passing `NULL` for the `realloc` function pointer. That [causes an early return](https://github.com/rhboot/shim/blob/main/Cryptlib/OpenSSL/crypto/mem.c#L157), so the calls weren't actually doing anything.

The `malloc`/`realloc`/`free` functions defined in [Cryptlib/SysCall/BaseMemAllocation.c](https://github.com/rhboot/shim/blob/main/Cryptlib/SysCall/BaseMemAllocation.c) are what actually get used, so just drop the explicit call to `CRYPTO_set_mem_functions`.

### DIFF
--- a/shim.c
+++ b/shim.c
@@ -1275,24 +1275,10 @@ EFI_STATUS set_second_stage (EFI_HANDLE image_handle)
 	return EFI_SUCCESS;
 }
 
-static void *
-ossl_malloc(size_t num)
-{
-	return AllocatePool(num);
-}
-
-static void
-ossl_free(void *addr)
-{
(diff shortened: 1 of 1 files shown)


dennis-tseng99 (approved):
Agree. Because CRYPTO_set_mem_functions() will always return if input parameter is NULL. ```if ((m == 0) || (r == 0) || (f == 0))   return 0```

PR #541, opened 2022-12-20
pe: Add IS_PAGE_ALIGNED macro

nicholasbishop:
This makes some checks in `get_mem_attrs` and `update_mem_attrs` clearer.

Also add `test-pe-util.c` with a test for the new macro. The file is named that way instead of `test-pe.c` to avoid having to get `pe.c` building in the unit test environment.


dennis-tseng99:
@Nicholasbishop, Actually, there already has a definition ```#define EFI_PAGE_MASK   0xFFF``` in ```gnu-efi/inc/efidef.h```. Maybe this definition could let your codes more perfect. Many thanks.

vathpela:
I've pushed this (with @dennis-tseng99's suggestion) as e4f40ae862b5389c8cc7d4f938a34238421456a1

Issue #538, opened 2022-12-17
`realloc` copies incorrect amount of memory, fails running under Mu firmware


nicholasbishop:
`realloc` is implemented here: https://github.com/rhboot/shim/blob/main/Cryptlib/SysCall/BaseMemAllocation.c#L29

It has a comment already pointing out the problem:
```
 // BUG: hardcode OldSize == size! We have no any knowledge about
 // memory size of original pointer ptr.
```

As currently implemented, it passes the desired size to `ReallocatePool` for both the old and new size. This happens to work fine usually, but I found when trying to test NX protection using https://github.com/microsoft/mu_tiano_platforms it reliably faults there.

I tried a very quick & hacky fix by statically keeping track of allocations in a fixed-size buffer so that `realloc` knows the allocation size, and that seemed to do the trick. As far as I can tell openssl requires a functioning `realloc`, so I think some variation of manually tracking allocations will be needed since UEFI doesn't provide such an interface. I don't have a PR for this yet, but figured I'd go ahead and file an issue for awareness.

dennis-tseng99:
```I tried a very quick & hacky fix by statically keeping track of allocations in a fixed-size buffer```

Yes. Because ```ReallocatePool()``` will copy your old buffer content to its new allocated buffer, the caller of ```realloc(void *ptr, size_t size)``` must prepare a fixed buffer by calling ```AllocatePool(size)``` with at least ```size``` bytes. Otherwise, some unknown garbage bytes will be copied to its new buffer. Many thanks.

nicholasbishop:
https://github.com/rhboot/shim/pull/543#issuecomment-1374861905 suggests importing fixes from EDK2, which sounds good to me.

Issue #382, opened 2021-06-11
Incorrect EFI_FILE_PROTOCOL->Open() usage in fallback


vathpela:
Laszlo reported this in a RHEL bugzilla ( https://bugzilla.redhat.com/show_bug.cgi?id=1966973 ), but I need to reference it here, so here's the story:

*** Description of problem:

In "fallback.c", in the following call tree:

  find_boot_csv()
    try_boot_csv()
      read_file()
        fh->Open()

the "fh" base handle, relative to which Open() is supposed to open a
file, is not a directory, but a regular file. This works with some
EFI_SIMPLE_FILE_SYSTEM_PROTOCOL implementations (notably
FatPkg/EnhancedFatDxe from edk2), but not with others (for example,
OvmfPkg/VirtioFsDxe).

The passage of the UEFI spec that governs the expected behavior of
EFI_FILE_PROTOCOL (aka (*EFI_FILE_HANDLE)) in the above context is
itself buggy.

*** Version-Release number of selected component (if applicable):
- shim-unsigned-x64-15.4-4.el8_1 (dist-git commit 1999eeb52fd2)
- via shim-15.4-2.el8_1 (dist-git commit eb38a06891a3)

*** How reproducible:
100%

*** Steps to Reproduce:

1. Install an OVMF binary no earlier than edk2-stable202102, on the virt
   host.

2. Install libvirtd packages no earlier than v7.1.0, on the virt host.

3. Install a reasonably recent QEMU, on the virt host.

4. Define a new libvirt domain with the following XML fragment
   (customize the host-side location of the virtio-fs root directory as
   needed):

  <filesystem type='mount' accessmode='passthrough'>
    <driver type='virtiofs'/>
    <binary path='/usr/libexec/virtiofsd'/>
    <source dir='.../virtio-fs-root'/>
    <target dir='VirtioFilesystem'/>
    <address type='pci' domain='0x0000'
     bus='0x00' slot='0x02' function='0x0'/>
    <boot order='1'/>
  </filesystem>

  No other bootable device should be present (disk, cd-rom, network
  card).

5. Using an installed (virtual or physical) RHEL-8.5 system as origin,
   recursively copy the "EFI" directory from "/boot/efi" to
   ".../virtio-fs-root" on the virt host:

   ssh rhel85 tar -C /boot/efi -c EFI | tar -C .../virtio-fs-root -x -v

6. Launch the domain.

*** Actual results:

shim logs:

[...]

PR #543, opened 2023-01-05
Cryptlib: Track allocation sizes to fix realloc

nicholasbishop:
Implementing `realloc` properly requires knowing the size of the allocation, so that the memory in the old allocation can be copied over to the new allocation. This information isn't provided by UEFI or gnu-efi. Previously the implementation of `realloc` copied over the new allocation size's number of bytes, which can fail with stricter memory protections.

Fix by expanding each allocation by 8 bytes, and storing the size of the allocation in the first 8 bytes of the allocation. The pointer returned by `malloc` is then 8 bytes after the actual pool allocation. The `realloc` and `free` functions subtract 8 bytes to get the "real" allocation pointer again for passing to the `ReallocatePool` and `FreePool` functions. The `realloc` function reads the original allocation size from the beginning of the allocation so that the correct number of bytes are copied to the new allocation.

Fixes https://github.com/rhboot/shim/issues/538

### DIFF
--- a/Cryptlib/SysCall/BaseMemAllocation.c
+++ b/Cryptlib/SysCall/BaseMemAllocation.c
@@ -19,20 +19,63 @@ WITHOUT WARRANTIES OR REPRESENTATIONS OF ANY KIND, EITHER EXPRESS OR IMPLIED.
 // -- Memory-Allocation Routines --
 //
 
+/* Prefix allocations with the allocation's total size.
+ *
+ * This is used to implement `realloc`, which needs to know the
+ * allocation's current size so that the data can be copied over to the
+ * new allocation.
+ */
+struct AllocWithSize {
+  /* Size of the whole allocation, including this `size` field.
+   *
+   * Allocations returned by AllocatePool are always eight-byte aligned
+   * (according to the UEFI spec), so use an eight-byte `size` field to
+   * make the `data` field also eight-byte aligned. */
+  UINT64 size;
+
+  /* The beginning of the allocation returned to the caller of malloc.
+   *
+   * The pointer to this data is what will then get passed into realloc
+   * and free; those functions use ptr_to_alloc_with_size to get the
+   * pointer to the underlying pool allocation. */
+  UINT8 data[0];
+};
+
+/* Convert a `ptr` (as returned by malloc) to an AllocWithSize by
+ * subtracting eight bytes. */
+static struct AllocWithSize *ptr_to_alloc_with_size (void *ptr) {
+  UINT8 *cptr = (UINT8*) ptr;
+  return (struct AllocWithSize*) (cptr - sizeof(struct AllocWithSize));
+}
+
 /* Allocates memory blocks */
 void *malloc (size_t size)
 {
-  return AllocatePool ((UINTN) size);
+  UINTN alloc_size = (UINTN) (size + sizeof(struct AllocWithSize));
+
+  struct AllocWithSize *alloc = AllocatePool (alloc_size);
+  if (alloc) {
+    alloc->size = alloc_size;
+    return alloc->data;
+  } else {
+    return NULL;
+  }
 }
 
 /* Reallocate memory blocks */
 void *realloc (void *ptr, size_t size)
 {
-  //
-  // BUG: hardcode OldSize == size! We have no any knowledge about
-  // memory size of original pointer ptr.
-  //
-  return ReallocatePool (ptr, (UINTN) size, (UINTN) size);
+  struct AllocWithSize *alloc = ptr_to_alloc_with_size (ptr);
+  UINTN old_size = alloc->size;
+  UINTN new_size = (UINTN) (size + sizeof(struct AllocWithSize));
+
+  alloc = ReallocatePool (alloc, old_size, new_size);
+  if (alloc) {
+    alloc->size = new_size;
+    return alloc->data;
+  } else {
+    return NULL;
+  }
 }
 
 /* De-allocates or frees a memory block */
@@ -43,6 +86,8 @@ void free (void *ptr)
   // is not true of FreePool() below, so protect it.
   //
   if (ptr != NULL) {
+    ptr = ptr_to_alloc_with_size (ptr);
+
     FreePool (ptr);
   }
 }


Cryptlib/SysCall/BaseMemAllocation.c
```diff
@@ -19,20 +19,63 @@ WITHOUT WARRANTIES OR REPRESENTATIONS OF ANY KIND, EITHER EXPRESS OR IMPLIED.
...
-  //
-  // BUG: hardcode OldSize == size! We have no any knowledge about
-  // memory size of original pointer ptr.
-  //
-  return ReallocatePool (ptr, (UINTN) size, (UINTN) size);
+  struct AllocWithSize *alloc = ptr_to_alloc_with_size (ptr);
+  UINTN old_size = alloc->size;
+  UINTN new_size = (UINTN) (size + sizeof(struct AllocWithSize));
+
+  alloc = ReallocatePool (alloc, old_size, new_size);
```
lersek, line 72: Two things to consider here.

**(1)** At this point, we're past the `malloc()` shortcut (which I pointed out above, for `ptr==NULL`); that is, we're actually resizing. It's still possible that we're resizing to zero (`size==0`). In that case, the logic here will preserve a just the "empty" `AllocWithSize` structure. That actually does conform to POSIX, again:

https://pubs.opengroup.org/onlinepubs/9699919799/functions/realloc.html

and returning the `alloc->data` flexible array member again conforms to the ISO C99 standard; **but* again this deserves a comment in the code.

**(2)** Unfortunately, the other thing to consider is a nasty preexistent bug, inside ReallocatePool().

As of commit 657b2483ca6e, shim consumes its own fork of gnu-efi (the submodule) at commit 03670e14f263. And at that point, the `ReallocatePool` function in `shim/gnu-efi/lib/misc.c` looks like this:

~~~
VOID *
ReallocatePool (
    IN VOID                 *OldPool,
    IN UINTN                OldSize,
    IN UINTN                NewSize
    )
{
    VOID                    *NewPool;

    NewPool = NULL;
    if (NewSize) {
        NewPool = AllocatePool (NewSize);
    }

    if (OldPool) {
        if (NewPool) {
            CopyMem (NewPool, OldPool, OldSize < NewSize ? OldSize : NewSize);
        }
    
        FreePool (OldPool);
    }
    
    return NewPool;
}
~~~

You do realize that this `ReallocatePool()` implementation frees `OldPool` even in case it attempts to, but *fails*, to allocate `NewPool`, right? That is, even in case the caller does not intend to use `ReallocatePool()` in a `FreePool()` role, but in an actual resizing role. Note that the releasing of `OldPool` *only* depends on `OldPool` being non-NULL on entry.

So, from our `realloc()` function, we're certainly calling `ReallocatePool()` with:
* a nonzero `old_size` (see the assertion I point out above),
* a nonzero `new_size`,
* and a non-NULL "old pool" pointer (because we're going to handle `ptr==NULL` on its own separate path, earlier).

That means this ReallocatePool() execution, internal to our realloc() implementation, is an actual *resizing* operation. And if the new area cannot be allocated, we're going to trash the existent area nevertheless. We're going to propagate `NULL` out to the caller of `realloc()`, and they will think that, while `realloc()` surely failed, they could continue accessing the *old* (original) allocation. They're going to think that because that's what POSIX guarantees -- but this implementation of ReallocatePool() violates.

In other words, `ReallocatePool()` itself is buggy.

... It's so very  demoralizing that *no matter where* I look in shim, I run into errors. `fallback.c` is full of issues (#382), and I've just briefly looked at `lib/simple_file.c`, and immediately spotted a leak in `simple_file_open_by_handle()` (namely that of `root`). There's basically no healthy code to build fixes upon; one doesn't even know where to start contributing patches (only for them to be ignored nearly forever). Oh well.

lersek (changes requested):

Cryptlib/SysCall/BaseMemAllocation.c
```diff
@@ -19,20 +19,63 @@ WITHOUT WARRANTIES OR REPRESENTATIONS OF ANY KIND, EITHER EXPRESS OR IMPLIED.
...
-  return AllocatePool ((UINTN) size);
+  UINTN alloc_size = (UINTN) (size + sizeof(struct AllocWithSize));
+
+  struct AllocWithSize *alloc = AllocatePool (alloc_size);
+  if (alloc) {
+    alloc->size = alloc_size;
+    return alloc->data;
+  } else {
+    return NULL;
+  }
```
lersek, line 62: It looks like github lost a review comment of mine here.

Here I pointed out that this should be formatted as follows -- no `else` after `return` please:

~~~
  if (alloc) {
    alloc->size = alloc_size;
    return alloc->data;
  }
  return NULL;
~~~

dennis-tseng99:
I merged ```malloc()```, ```realloc()``` and ```free()``` from _**https://github.com/tianocore/edk2/blob/master/CryptoPkg/Library/BaseCryptLib/SysCall/BaseMemAllocation.c**_  into shim in my local site, and did a testing roughly. It seems working well. Maybe we should align our shim functions to EDK2 codes.

Here is my understanding:

```[OldPoolHdr]

<--- CRYPTMEM_OVERHEAD ---><---- OldSize ------>
+-------------------------+--------------------+
|     PoolHdr             | real ptr   ......  |             
+-------------------------+--------------------+
^                         ^
OldPoolHdr               ptr

OldPoolHdr = (CRYPTMEM_HEAD *)ptr - 1;
OldPoolHdr will be freed by FreePool()

[NewPoolHdr]

<-------------------------- NewSize ---------------------------------->
<--- CRYPTMEM_OVERHEAD ---><------------------ size ------------------>
+-------------------------+-------------------------------------------+
|        NewPoolHdr       |             empty space                   |             
+-------------------------+-------------------------------------------+
^                         ^
NewPoolHdr                NewPoolHdr + 1 <--- return vaule
```

Here is the orginal EDK2 committed patch:
```commit cf8197a39d07179027455421a182598bd6989999
Author: Long Qin <qin.long@intel.com>
Date:   Wed Nov 1 16:10:04 2017 +0800

    CryptoPkg/BaseCryptLib: Fix buffer overflow issue in realloc wrapper
    
    There is one long-standing problem in CRT realloc wrapper, which will
    cause the obvious buffer overflow issue when re-allocating one bigger
    memory block:
        void *realloc (void *ptr, size_t size)
        {
          //
          // BUG: hardcode OldSize == size! We have no any knowledge about
          // memory size of original pointer ptr.
          //
          return ReallocatePool ((UINTN) size, (UINTN) size, ptr);
        }
    This patch introduces one extra header to record the memory buffer size
    information when allocating memory block from malloc routine, and re-wrap
    the realloc() and free() routines to remove this BUG.
    
    Cc: Laszlo Ersek <lersek@redhat.com>
    Cc: Ting Ye <ting.ye@intel.com>
    Cc: Jian J Wang <jian.j.wang@intel.com>
    Contributed-under: TianoCore Contribution Agreement 1.0
    Signed-off-by: Qin Long <qin.long@intel.com>
    Reviewed-by: Jian J Wang <jian.j.wang@intel.com>
    Validated-by: Jian J Wang <jian.j.wang@intel.com>

diff --git a/CryptoPkg/Library/BaseCryptLib/SysCall/BaseMemAllocation.c b/CryptoPkg/
Library/BaseCryptLib/SysCall/BaseMemAllocation.c
index f390e0d449..19c071e2bf 100644
--- a/CryptoPkg/Library/BaseCryptLib/SysCall/BaseMemAllocation.c
+++ b/CryptoPkg/Library/BaseCryptLib/SysCall/BaseMemAllocation.c
@@ -16,6 +16,18 @@ WITHOUT WARRANTIES OR REPRESENTATIONS OF ANY KIND, EITHER EXPRESS
 OR IMPLIED.
 #include <CrtLibSupport.h>
 #include <Library/MemoryAllocationLib.h>
 
+//
+// Extra header to record the memory buffer size from malloc routine.
+//
+#define CRYPTMEM_HEAD_SIGNATURE    SIGNATURE_32('c','m','h','d')
+typedef struct {
+  UINT32    Signature;
+  UINT32    Reserved;
+  UINTN     Size;
+} CRYPTMEM_HEAD;
+
+#define CRYPTMEM_OVERHEAD      sizeof(CRYPTMEM_HEAD)
+  
 //
 // -- Memory-Allocation Routines --
 //
@@ -23,27 +35,84 @@ WITHOUT WARRANTIES OR REPRESENTATIONS OF ANY KIND, EITHER EXPRES
S OR IMPLIED.
 /* Allocates memory blocks */
 void *malloc (size_t size)
 {
-  return AllocatePool ((UINTN) size);
+  CRYPTMEM_HEAD  *PoolHdr;
+  UINTN          NewSize;
+  VOID           *Data;
+
+  //
+  // Adjust the size by the buffer header overhead
+  //
+  NewSize = (UINTN)(size) + CRYPTMEM_OVERHEAD;
+
+  Data  = AllocatePool (NewSize);
+  if (Data != NULL) {
+    PoolHdr = (CRYPTMEM_HEAD *)Data;
+    //
+    // Record the memory brief information
+    //
+    PoolHdr->Signature = CRYPTMEM_HEAD_SIGNATURE;
+    PoolHdr->Size      = size;
+
+    return (VOID *)(PoolHdr + 1);
+  } else {
+    //
+    // The buffer allocation failed.
+    //
+    return NULL;
+  }
 }
 
 /* Reallocate memory blocks */
 void *realloc (void *ptr, size_t size)
 {
-  //
-  // BUG: hardcode OldSize == size! We have no any knowledge about
-  // memory size of original pointer ptr.
-  //
-  return ReallocatePool ((UINTN) size, (UINTN) size, ptr);
+  CRYPTMEM_HEAD  *OldPoolHdr;
+  CRYPTMEM_HEAD  *NewPoolHdr;
+  UINTN          OldSize;
+  UINTN          NewSize;
+  VOID           *Data;
+
+  NewSize = (UINTN)size + CRYPTMEM_OVERHEAD;
+  Data = AllocatePool (NewSize);
+  if (Data != NULL) {
+    NewPoolHdr = (CRYPTMEM_HEAD *)Data;
+    NewPoolHdr->Signature = CRYPTMEM_HEAD_SIGNATURE;
+    NewPoolHdr->Size      = size;
+    if (ptr != NULL) {
+      //
+      // Retrieve the original size from the buffer header.
+      //
+      OldPoolHdr = (CRYPTMEM_HEAD *)ptr - 1;
+      ASSERT (OldPoolHdr->Signature == CRYPTMEM_HEAD_SIGNATURE);
+      OldSize = OldPoolHdr->Size;
+
+      //
+      // Duplicate the buffer content.
+      //
+      CopyMem ((VOID *)(NewPoolHdr + 1), ptr, MIN (OldSize, size));
+      FreePool ((VOID *)OldPoolHdr);
+    }
+
+    return (VOID *)(NewPoolHdr + 1);
+  } else {
+    //
+    // The buffer allocation failed.
+    //
+    return NULL;
+  }
 }
 
 /* De-allocates or frees a memory block */
 void free (void *ptr)
 {
+  CRYPTMEM_HEAD  *PoolHdr;
+
   //
   // In Standard C, free() handles a null pointer argument transparently. This
   // is not true of FreePool() below, so protect it.
   //
   if (ptr != NULL) {
-    FreePool (ptr);
+    PoolHdr = (CRYPTMEM_HEAD *)ptr - 1;
+    ASSERT (PoolHdr->Signature == CRYPTMEM_HEAD_SIGNATURE);
+    FreePool (PoolHdr);
   }
 }```

nicholasbishop:
Thanks @dennis-tseng99, good find. Importing those changes from edk2 is probably the better path here, so I'll close this PR.

Cryptlib/SysCall/BaseMemAllocation.c
nicholasbishop: Thanks for your detailed review comments. I ended up closing this PR as discussed below. The new PR does not call `ReallocatePool` anymore.

> It's so very demoralizing that no matter where I look in shim, I run into errors.

Well, we just have to fix things one step at a time :) I know it can be frustrating to find errors, but taking a step back I don't think it's fair to stay there's no healthy code to build fixes on. Just because there are additional errors that need to be fixed doesn't mean it's not worth it to go ahead and fix some bugs.

lersek: Thanks for the comment. In general I agree with that attitude. It's just that having to fix (say) three bugs as a dependency chain for the one bug you actually want to fix, is frustrating enough in itself. On top, not seeing any maintainer feedback for a good while in the discussion phase, does not evoke a lot of trust that your work, which has just grown unexpectedly in scope, will not be in vain. "One step at a time" is definitely the way to go, but once it grows into "one series at a time", maintainer overload becomes an even worse obstacle, and a strong discouragement. It makes me doubt my effort is worthwhile. But yes, in general you are right.

PR #546, opened 2023-01-09
CryptoPkg/BaseCryptLib: Fix buffer overflow issue in realloc wrapper

nicholasbishop:
There is one long-standing problem in CRT realloc wrapper, which will cause the obvious buffer overflow issue when re-allocating one bigger memory block:
    void *realloc (void *ptr, size_t size)
    {
      //
      // BUG: hardcode OldSize == size! We have no any knowledge about
      // memory size of original pointer ptr.
      //
      return ReallocatePool ((UINTN) size, (UINTN) size, ptr);
    }
This patch introduces one extra header to record the memory buffer size information when allocating memory block from malloc routine, and re-wrap the realloc() and free() routines to remove this BUG.

Cc: Laszlo Ersek <lersek@redhat.com>
Cc: Ting Ye <ting.ye@intel.com>
Cc: Jian J Wang <jian.j.wang@intel.com>
Contributed-under: TianoCore Contribution Agreement 1.0
Signed-off-by: Qin Long <qin.long@intel.com>
Reviewed-by: Jian J Wang <jian.j.wang@intel.com>
Validated-by: Jian J Wang <jian.j.wang@intel.com>

Cherry picked from https://github.com/tianocore/edk2.git, commit cf8197a39d07179027455421a182598bd6989999. Changes:
* `SIGNATURE_32` -> `EFI_SIGNATURE_32`
* Added definition of `MIN`

Fixes https://github.com/rhboot/shim/issues/538

### DIFF
--- a/Cryptlib/SysCall/BaseMemAllocation.c
+++ b/Cryptlib/SysCall/BaseMemAllocation.c
@@ -16,33 +16,104 @@ WITHOUT WARRANTIES OR REPRESENTATIONS OF ANY KIND, EITHER EXPRESS OR IMPLIED.
 #include <OpenSslSupport.h>
 
 //
+// Extra header to record the memory buffer size from malloc routine.
+//
+#define CRYPTMEM_HEAD_SIGNATURE    EFI_SIGNATURE_32('c','m','h','d')
+typedef struct {
+  UINT32    Signature;
+  UINT32    Reserved;
+  UINTN     Size;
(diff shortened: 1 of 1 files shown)


dennis-tseng99 (approved):
Align malloc(), realloc() and free() to EDK2 upstream codes is good idea.

vathpela (approved):

Issue #524, opened 2022-11-07
Shim unable to locate grubx64 in PXE boot mode when grubx64 is stored in a different file path.


Alberto-David-Perez-Guevara:
Shim unable to locate grubx64 in PXE boot mode when grubx64 is stored in a different file path.
**SETUP**: 

I have a  system attempting to use shim (tried rel 15.5 & 15.4 with same results) to do PXE boot of Linux OS distro from a  PXE server. both bootx64.efi (shim) & grubx64.efi are in the same directory. The chosen directory for this experiment is for example: 
MaintenanceOS\ESXi_LinuxISO\bootx64.efi
MaintenanceOS\ESXi_LinuxISO\grubx64.efi

**RESULT**:

system is able able to connect using  to PXE server and fetch the Shim i.e. bootx64.efi file. However system then Hangs and displays the following information in the console:
 
start_image() returned TFTP Error
ASSERT [DxeCore] d:\a\_work\1\s\development\biossource\MdeModulePkg\Core\Dxe\Mem\Pool.c(624): CR has Bad Signature

**INFO & Logs**:

from our network trace we see the following TFTP packets for fetching the shim and grub efi files as follows:

for shim:
 Trivial File Transfer Protocol
     Opcode: Read Request (1)
     Source File: _\MaintenanceOS\ESXi_LinuxISO\_bootx64.efi

for grub:
 Trivial File Transfer Protocol
     Opcode: Read Request (1)
     Source File_:grubx64.efi_

**DEBUG**:

It seems the main issue is that Shim is unable to correctly deduce the file path, on this particular boot option Shim behaves as follows, 

efi_main(entry point)-> shim_init->   parse_load options->    set_second_stage-> generate_path_from_image_path -> start_image->parseDhcp4.

adding Debug prints we can see the error more closely

`DEBUG parse_load_options ->  bootpath: PciRoot(9)/Pci(0x0,0x0)/Pci(0x0,0x0)/Mac(XXYYZZ)/IPv4(0.0.0.0,0,DHCP)/\MaintenanceOS\ESXi_LinuxISO\bootx64.efi

DEBUG set_second_stage ->  second_stage: [\\grubx64.efi](file:///)

DEBUG generate_path_from_image_path ->  ImagePath: [\\grubx64.efi](file:///)  | PathName: (null)

[...]

PR #551, opened 2023-01-30
 Issue #524 Shim unable to locate grubx64 in PXE boot mode when grubx…

Alberto-Perez-Guevara:
…64 is stored in a different file path.

 Problem: grubx64 not found when file path contains '\'

 Root Cause: parseDhcp4 function is unable to handle file path with '\'

 Changes to be committed:
	modified:   netboot.c

### DIFF
--- a/netboot.c
+++ b/netboot.c
@@ -263,9 +263,10 @@ static EFI_STATUS parseDhcp4()
 	UINT8 *dir = pkt_v4->BootpBootFile;
 
 	for (i = dir_len; i >= 0; i--) {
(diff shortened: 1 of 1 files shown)


Alberto-Perez-Guevara:
Hi any feedback on this is most welcome

p3ck:
This would be really helpful for our test automation. Please consider

vathpela:
Pushed as 0bfc3978f4a6a10e4427fdab222b0e50c3c7283c.

Issue #558, opened 2023-03-29
Possible regression: 2021.08.12 shim boots whereas 2023.01.18 freezes in a chain-loaded Secure Boot context


pbatard:
### Issue description

It appears that, when running in a Secure Boot environment, the current shim introduced a regression, that is currently manifesting itself and affecting users trying to create NTFS-based boot media from Ubuntu Studio 22.04.2 LTS (this was tested with `ubuntustudio-22.04.2-dvd-amd64.iso`) when using the [NTFS chain loader](https://github.com/pbatard/uefi-ntfs#readme) from the [Rufus boot media creation utility](https://github.com/pbatard/rufus).

Basically, because the Ubuntu ISO contains a file that is larger than 4 GB, Rufus extracts the ISO content onto an NTFS partition rather than a FAT32 partition (since FAT cannot accommodate files larger than 4 GB) and then, for UEFI systems that do not have a native NTFS driver, chains load the Red Hat Shim from a small additional FAT partition that contains a UEFI boot loader that can load a UEFI NTFS driver and then hands of boot to the NTFS-residing bootloaders.

This chain loader, called [UEFI:NTFS](https://github.com/pbatard/uefi-ntfs) is itself Secure Boot signed, along with its **read-only** NTFS driver [that is derived from ntfs-3g](https://github.com/pbatard/ntfs-3g). Note that this driver was restricted to read-only as a means to improve its resilience against potential vulnerabilities and therefore facilitate the Secure Boot validation process.

However, we have recently found out that, when Secure Boot is active (the issue does **not** manifest itself when Secure Boot is Disabled or in Setup mode), the process above appears to freeze during shim execution when using the 2023.01.18 Secure Boot signed version. Yet, when replacing the shim with a 2021.08.12 Secure Boot signed version, everything works as expected.

Thus, it would appear that a regression was introduced between 2021.07 and 2023.01.18 that causes a freezout when shim is chain loaded from a read-only NTFS partition.

### Steps to replicate

The attached [`shim.zip`](https://github.com/rhboot/shim/files/11101146/shim.zip) contains a 8 MB uncompressed VHD disk image, which, for all intents and purposes **will** work the same as a `dd` disk image (since uncompressed VHD = regular disk image + 512 bytes footer), that you should be able to use with any USB disk media to recreate a test disk that includes:

[...]

PR #547, opened 2023-01-16
Don't loop forever in load_certs() with buggy firmware

rmetrich:
On DELL R350 booting DVD through RFS with BIOS 1.4.2 in Secure Boot, firmware returns *EFI_BUFFER_TOO_SMALL* but with new *buffersize* set to 0, which causes the `load_certs()` code to loop forever:

~~~
while (1) {
    efi_status = dir->Read(dir, &buffersize, buffer);
    if (efi_status == EFI_BUFFER_TOO_SMALL) {
        ...
        continue;
    }
    ...
}
~~~

This commit prevents such infinite loop. The new code doesn't check if new *buffersize* is 0 only, but if it's different from *old* one we passed, which may cover more buggy firmware cases at minor cost.

PR #560, opened 2023-03-31
Further improve load_certs() for non-compliant drivers/firmwares

pbatard:
Following the discovery of more problematic firmwares and drivers affected by the issue f23883ccf78f1f605a272f9e5700f47e5494a71d is designed to address (e.g. https://github.com/rhboot/shim/issues/558), this patch further improves the code so that, instead of simply bailing out, we progressively increase the buffer sizes, until either success or a maximum size limit is reached.

In most cases, this workaround should be enough to ensure completion of the directory read and thus provide full shim functionality (while still warning the user about the non-compliance of their environment).

### DIFF
--- a/shim.c
+++ b/shim.c
@@ -1486,18 +1486,28 @@ load_certs(EFI_HANDLE image_handle)
 		UINTN old = buffersize;
 		efi_status = dir->Read(dir, &buffersize, buffer);
 		if (efi_status == EFI_BUFFER_TOO_SMALL) {
-			if (buffersize != old) {
-				buffer = ReallocatePool(buffer, old, buffersize);
-				if (buffer == NULL) {
-					perror(L"Failed to read directory %s - %r\n",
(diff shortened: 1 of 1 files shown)


pbatard:
Just going to add that this is a follow up to #547.

We tested this patch in the context of #558 (i.e. with a file system driver that returned `0` instead of the required size on `EFI_BUFFER_TOO_SMALL`) and found that it was enough to complete the directory read.

Note that, as buffer size will be increased, multiple messages might be displayed, but since the point is also to alert the user about the non-compliance of their driver or firmware, we don't see these potential repeated alerts as an issue.

We set the buffer size limit for bail out to 1024, which we think should be more than enough to handle file names that are expected to be less than 512 UTF-16 characters...

vathpela:
Pushed with a minor fixup as cf59f3452d478455c5f3d83790b37a372d2837ea .

pbatard:
Thanks!

Issue #561, opened 2023-04-10
shim image loader assumes padding relocs are added by GenFw, but not all efi files have this


mikebeaton:
The shim handle_image method assumes that RelocDir->Size == Section->Misc.VirtualSize, but in fact this is not required.

While current EDK-II GenFw generates files where the above-mentioned two values are equal (and also pads the .reloc section above RelocDir->Size with dummy reloc entries, which the shim loader processes), in general RelocDir->Size may be smaller than the padding aligned Section->Misc.VirtualSize of the .reloc section, and the padding area need not contain dummy reloc entries.

When an .efi file loaded by shim does not match this current requirement, the file loads but then crashes with a page fault because the relocations have been silently dropped by shim. However the same .efi file(s) [see message below] start fine on the firmware of multiple systems, including OVMF - and on shim when modified by the linked PR.

Attached is a version of UEFI Shell showing this issue. It is loaded fine by the firmware of many systems, and we believe is fully valid, but is not currently loaded successfully by the shim image handler, instead showing this issue.

[OpenShell.zip](https://github.com/rhboot/shim/files/11188602/OpenShell.zip)

mikebeaton:
Related: https://github.com/acidanthera/bugtracker/issues/2239
Possibly related(?): https://github.com/rhboot/shim/issues/97

vit9696:
@vathpela, any progress on this? The issue is currently a blocker for us. Thanks!

mikebeaton:
Closed with https://github.com/rhboot/shim/pull/562#issuecomment-1601343882 - tyvm

PR #562, opened 2023-04-10
pe: only process RelocDir->Size of reloc section

mikebeaton:
Previously processing full padding-aligned Section->Misc.VirtualSize relied on padding reloc entries being inserted by GenFw, which is not required by spec.

Addresses https://github.com/rhboot/shim/issues/561

### DIFF
--- a/pe.c
+++ b/pe.c
@@ -87,7 +87,7 @@ relocate_coff (PE_COFF_LOADER_IMAGE_CONTEXT *context,
 	/* RelocBaseEnd here is the address of the first entry /past/ the
 	 * table.  */
 	RelocBaseEnd = ImageAddress(orig, size, Section->PointerToRawData +
-						Section->Misc.VirtualSize);
+						context->RelocDir->Size);
 
 	if (!RelocBase && !RelocBaseEnd)
 		return EFI_SUCCESS;
@@ -741,7 +741,7 @@ read_header(void *data, unsigned int datasize,
 	context->NumberOfSections = PEHdr->Pe32.FileHeader.NumberOfSections;
 
 	if (EFI_IMAGE_NUMBER_OF_DIRECTORY_ENTRIES < context->NumberOfRvaAndSizes) {
-		perror(L"Image header too small\n");
+		perror(L"Image header too large\n");
 		return EFI_UNSUPPORTED;
 	}
 
@@ -1277,8 +1277,11 @@ handle_image (void *data, unsigned int datasize,
 					Section->Misc.VirtualSize &&
 					base && end &&
 					RelocBase == base &&
-					RelocBaseEnd == end) {
+					RelocBaseEnd <= end) {
 				RelocSection = Section;
+			} else {
+				perror(L"Relocation section is invalid \n");
+				return EFI_UNSUPPORTED;
 			}
 		}
 


pe.c
```diff
@@ -1277,8 +1277,11 @@ handle_image (void *data, unsigned int datasize,
 					Section->Misc.VirtualSize &&
 					base && end &&
 					RelocBase == base &&
-					RelocBaseEnd == end) {
+					RelocBaseEnd <= end) {
```
mikebeaton, line 1280: Without this change, the file loads but crashes (since shim will attempt to use the file as if there is no reloc section when there is).

pe.c
```diff
@@ -1277,8 +1277,11 @@ handle_image (void *data, unsigned int datasize,
...
 				RelocSection = Section;
+			} else {
+				perror(L"Relocation section is invalid \n");
+				return EFI_UNSUPPORTED;
```
mikebeaton, line 1284: Suggest in the case where a reloc section is found but rejected, aborting with an error such as this is helpful (unless there are known cases where this behaviour is required?).

pe.c
```diff
@@ -87,7 +87,7 @@ relocate_coff (PE_COFF_LOADER_IMAGE_CONTEXT *context,
 	/* RelocBaseEnd here is the address of the first entry /past/ the
 	 * table.  */
 	RelocBaseEnd = ImageAddress(orig, size, Section->PointerToRawData +
-						Section->Misc.VirtualSize);
+						context->RelocDir->Size);
```
mikebeaton, line 90: Once the test below is modifed to accept the .reloc section, we also have to process only the used size.

pe.c
```diff
@@ -741,7 +741,7 @@ read_header(void *data, unsigned int datasize,
 	context->NumberOfSections = PEHdr->Pe32.FileHeader.NumberOfSections;
 
 	if (EFI_IMAGE_NUMBER_OF_DIRECTORY_ENTRIES < context->NumberOfRvaAndSizes) {
-		perror(L"Image header too small\n");
+		perror(L"Image header too large\n");
```
mikebeaton, line 744: Unrelated really - apologies if I'm misreading this? (Could make separate tiny PR; or remove, if the wording change does not seem right. Tyvm!)

pe.c
mikebeaton: Hi,

Thanks for this input.

Is BS->FreePages correct, here? The various other (pre-existing) `return EFI_UNSUPPORTED` nearby (and in the same allocation context, I believe), do not do it.

Then on the general idea, yes, in the current version of this PR the code continues if no reloc section is finally found (because this is what the code previously did), but the correct place to decide not to do so (which I tend to agree would be an improvement) would be around line 1331: https://github.com/rhboot/shim/blob/0640e136842200eb5873e9b2b02b159e15940591/pe.c#L1331

If you (possibly) meant the suggestion as an alternative fix - i.e. to avoid my crash - then I think that misunderstands the nature of the PR - the file _should_ load and run (not abort, and also - ofc - not crash) and this PR allows it to load.

vathpela:
Pushed as a8b0b600ddcf02605da8582b4eac1932a3bb13fa.

Issue #251, opened 2021-01-27
msleep uses stall which expects microseconds


jirislaby:
In the definition of `msleep`, `Stall` is used:
https://github.com/rhboot/shim/blob/8a2632d35ed41e244aa795d1ee45d3090a848471/lib/console.c#L687

But it expects microseconds (usually marked as `usecs`), not miliseconds (usually `msecs`). The calling code seems to pass microseconds at all places, it seems. So it seems to confuse only me. Would it be possible to rename the variable or add a comment, so noone else is confused later too?

vathpela:
Sure, send a pull request.

PR #563, opened 2023-04-12
Rename 'msecs' to 'usecs' to avoid potential confusion

aronowski:
As suggested in https://github.com/rhboot/shim/issues/251.

The function `msleep` uses `gBS->Stall` which waits for a specified number of microseconds.

Reference: https://edk2-docs.gitbook.io/edk-ii-uefi-driver-writer-s-guide/5_uefi_services/51_services_that_uefi_drivers_commonly_use/517_stall

This reference even mentions an example slepping for 10 microseconds: `// Wait 10 uS`. Notice the letter 'u'.

Therefore it's a good idea to call the argument `msleep` uses 'usecs' rather than 'msecs' so no one confuses it with miliseconds.

### DIFF
--- a/fallback.c
+++ b/fallback.c
@@ -1006,7 +1006,7 @@ try_start_first_option(EFI_HANDLE parent_image_handle)
 	EFI_HANDLE image_handle;
 
 	if (get_fallback_verbose()) {
-		int fallback_verbose_wait = 500000; /* default to 0.5s */
+		unsigned long fallback_verbose_wait = 500000; /* default to 0.5s */
 #ifdef FALLBACK_VERBOSE_WAIT
 		fallback_verbose_wait = FALLBACK_VERBOSE_WAIT;
 #endif
@@ -1211,7 +1211,7 @@ reset:
 	console_print(L"Reset System\n");
 
 	if (get_fallback_verbose()) {
-		int fallback_verbose_wait = 500000; /* default to 0.5s */
+		unsigned long fallback_verbose_wait = 500000; /* default to 0.5s */
 #ifdef FALLBACK_VERBOSE_WAIT
 		fallback_verbose_wait = FALLBACK_VERBOSE_WAIT;
 #endif
--- a/fallback.c
+++ b/fallback.c
@@ -1013,7 +1013,7 @@ try_start_first_option(EFI_HANDLE parent_image_handle)
 		console_print(L"Verbose enabled, sleeping for %d mseconds... "
 			      L"Press the Pause key now to hold for longer.\n",
 			      fallback_verbose_wait);
-		msleep(fallback_verbose_wait);
+		usleep(fallback_verbose_wait);
 	}
 
 	if (!first_new_option) {
@@ -1036,7 +1036,7 @@ try_start_first_option(EFI_HANDLE parent_image_handle)
 		}
 		console_print(L"\n");
 
-		msleep(500000000);
+		usleep(500000000);
 		return efi_status;
 	}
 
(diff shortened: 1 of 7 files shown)


dennis-tseng99:
Normally, parameter usecs is used for usleep(useconds_t usec), and msecs is used for msleep(unsigned int msecs). Hence, to avoid confusion, could we also change the function name from msleep() to usleep() ?

aronowski:
I like this idea but the implementation of it may be more complex than I initially thought.

I'm not a developer as of today and I'm unsure if such a change won't break anything crucial. I don't want to introduce major changes especially for a security-critical component such as Shim.

Changing argument names only resulted in the exact same binary, while changing function names results in a binary with a different checksum. While it works fine and allows for booting a system with Secure Boot enabled, I'm unsure if I should do this.

Maybe I'll let the Red Hat Bootloader Team decide on this.

vathpela:
The function name should be changed as well.

dennis-tseng99:
@aronowski  If you don't mind, please let me change the function name and its argument as well in another PR according to vathpela's suggestion.

aronowski:
The function has been renamed according to Peter's request.

@dennis-tseng99 I did this change once I saw my first email notification rather than following the complete conversation. You can, however, review the latest commit so I can be sure I haven't made any mistake.

dennis-tseng99 (approved):
In fallback.c, would you please also change **int** fallback_verbose_wait to **unsigned long** ?
```
1009                 - int fallback_verbose_wait = 500000; /* default to 0.5s */
1009                 + unsigned long fallback_verbose_wait = 500000; /* default to 0.5s */
1016                 usleep(fallback_verbose_wait);
```
```
1214                 - int fallback_verbose_wait = 500000; /* default to 0.5s */
1214                 + unsigned long fallback_verbose_wait = 500000; /* default to 0.5s */
1221                 usleep(fallback_verbose_wait);
```

vathpela:
I think wrapping the decl in a test for `SHIM_UNIT_TEST` should fix the unit test failures:
```
diff --git a/include/console.h b/include/console.h
index 04f02da7e01..8eb4b47f912 100644
--- a/include/console.h
+++ b/include/console.h
@@ -122,7 +122,9 @@ extern EFI_STATUS EFIAPI vdprint_(const CHAR16 *fmt, const char *file, int line,
 extern EFI_STATUS print_crypto_errors(EFI_STATUS rc, char *file, const char *func, int line);
 #define crypterr(rc) print_crypto_errors((rc), __FILE__, __func__, __LINE__)

+#ifndef SHIM_UNIT_TEST
 extern VOID usleep(unsigned long usecs);
+#endif

 /* This is used in various things to determine if we should print to the
  * console */
```

aronowski:
@dennis-tseng99 @vathpela
Thanks a lot for the hints. I've added the two changes.
In both cases I've confirmed in my public fork that unit tests pass.

dennis-tseng99 (commented):
Except header file, may I suggest we also add "#ifndef SHIM_UNIT_TEST" to the function itself ?
```
/* lib/console.c */
#ifndef SHIM_UNIT_TEST
VOID
usleep(unsigned long usecs)
{
	BS->Stall(usecs);
}
#endif
```

aronowski:
Sure thing. It's right there! :grinning:

dennis-tseng99 (approved):

vathpela:
I've squashed this up and re-ordered it a bit so that each commit passes the tests, and pushed it as 549d34691d68518e55c2edd6e759b19de7f8ddef and 908c388c922c6369cace0b76660198becee2284e.

PR #564, opened 2023-04-23
Add more clear comments when decoding asn.1 tag and length bytes

dennis-tseng99:
Actually, 
```
SpcIndirectDataContent += 2;
SpcIndirectDataContent += 3;
SpcIndirectDataContent += 4;
```
will skip Tag and Length fields simultaneously, rather then just skip Tag.
Related comments are modified for easier tracing thereafter.


vathpela:
This should be fixed in https://github.com/tianocore/edk2/blob/master/CryptoPkg/Library/BaseCryptLib/Pk/CryptAuthenticode.c , to be picked up next time we rebase, instead.

PR #569, opened 2023-05-09
Add SbatLevel_Variable.txt to document the various revocations

jsetje:
This serves to document the SbatLevel Boot Services variable so that other boot services code, such as bootmgr can update the revocation level.

We did not do the best job documenting these CVE lists in the past, hopefully this file will fix that as well. Reviews of the lists in particular would be appreciated.

### DIFF
--- a/SbatLevel_Variable.txt
+++ b/SbatLevel_Variable.txt
@@ -1,0 +1,42 @@
+In order to apply SBAT based revocations on systems that will never
+run shim, code running in boot services context needs to set the
+following variable:
+
+Name: SbatLevel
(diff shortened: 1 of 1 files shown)


SbatLevel_Variable.txt
```diff
@@ -0,0 +1,42 @@
...
+
+To revoke the above and also grub binaries impacted by
+
+* CVE-2022-2601
+* CVE-2022-3775
+
+sbat,1,2022111500
+grub,3
+
+Ad additonal bug was fixed in shim that was not considered exploitable
```
Foxboron, line 37: An additional bug*

Foxboron (changes requested):

vathpela (approved):

PR #577, opened 2023-06-23
mok: fix LogError() invocation

vathpela:
On some ARM platform, jlinton noticed that when we fail to set a variable (because it isn't supported at all, presumably), our error message has an extra argument that doesn't match the format string.

This patch removes the extra argument.


steve-mcintyre (approved):
Looks good to me

eslerm:
Is this issue actually a security vulnerability, or just a bug? Why was a CVE assigned?

PR #580, opened 2023-06-25
FreePages() is used to return memory allocated by AllocatePages() 

dennis-tseng99:
There are several **EFI_UNSUPPORTED** exceptions in ```handle_image()``` of pe.c
Some of them are missing returning page memory to firmware by ```FreePages()```; while some of them are using unmatched ```FreePoll()``` function which implicitly is ```uefi_call_wrapper(BS->FreePool, 1, Buffer)``` to return memory. According to UEFI spec 2.0, ```FreePoll()``` is only used for ```AllocatePoll()``` when memory is returned to the available pool.

### DIFF
--- a/pe.c
+++ b/pe.c
@@ -747,6 +747,7 @@ handle_image (void *data, unsigned int datasize,
 		    (Section->Characteristics & EFI_IMAGE_SCN_MEM_EXECUTE) &&
 		    (mok_policy & MOK_POLICY_REQUIRE_NX)) {
 			perror(L"Section %d is writable and executable\n", i);
+			BS->FreePages(*alloc_address, *alloc_pages);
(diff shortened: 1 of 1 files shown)


jsetje (changes requested):
The diff in relocate_coff() is already merged and can be dropped. Can you merge and re-push? Otherwise this looks like a good fix. Thank you!

(If I don't hear from you I can try to figure out the edit feature.)

dennis-tseng99:
@jsetje, I've dropped the diff in relocate_coff(). Thank you so much.

vathpela:
I've pushed this as 1e985a3a238100ca5f4bda3e269a9eaec9bda74b.

PR #581, opened 2023-06-27
Size should minus 1 when calculating 'RelocBaseEnd'

dennis-tseng99:
```context->RelocDir->Size``` should minus one when calling ```ImageAddress()```. Otherwise
**RelocBaseEnd** will get NULL address after running ImageAddress() in which ```address=size```.
```
ImageAddress (void *image, uint64_t size, uint64_t address)
{
        uintptr_t img_addr;

        /* ensure our local pointer isn't bigger than our size */
        if (address >= size)   <---- address=size
                return NULL;
        ........
}
```

```RelocBaseEnd = ImageAddress(orig, size, Section->PointerToRawData + context->RelocDir->Size);```

### DIFF
--- a/pe-relocate.c
+++ b/pe-relocate.c
@@ -75,7 +75,7 @@ relocate_coff (PE_COFF_LOADER_IMAGE_CONTEXT *context,
 	/* RelocBaseEnd here is the address of the first entry /past/ the
 	 * table.  */
 	RelocBaseEnd = ImageAddress(orig, size, Section->PointerToRawData +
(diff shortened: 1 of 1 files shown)


jsetje (approved):
I ran into this as well, and short of creating new versions of the ImageAddress() range check that test for things like being past the end of the section, this seems reasonable.

vathpela:
I've force pushed fd43edacfe1eaa64d1c7b37ba23dae74e05d1a04 over this to make the commit message more informative.

PR #583, opened 2023-06-27
Verify signature before verifying sbat levels

jsetje:
Verifying the validity of a files signature can protect from an attacker creating a file that exploits a pontential issue in the sbat validation. If the singature is not checked first, an attacker can just create a file with a valid .sbat section and can still attack the signature valiation.


vathpela (approved):
I fixed a few typos in the commit message, other than that it looks good.

PR #584, opened 2023-06-28
Add libFuzzer support for csv.c and sbat.c

vathpela:
shim takes several forms of input from several sources that are not necessarily trustworthy.  As such, we need to take measures to validate that we don't have unacceptable results from bad inputs.  One such measure is "fuzzing" the inputs which parse untrusted data by running them with randomized or partially randomized input.

This change adds such testing using clang's "libFuzzer" to our CSV parser and the parser for `.sbat` sections.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,6 +35,10 @@ Make.local
 /compile_commands.json
 /compile_commands.events.json
 /cov-int/
+/crash-*
+/fuzz-*
+!/fuzz-*.c
+/leak-*
 /post-process-pe
 /random.bin
 /sbat.*.csv
--- a/Makefile
+++ b/Makefile
@@ -290,6 +290,15 @@ else
 	$(PESIGN) -n certdb -i $< -c "shim" -s -o $@ -f
 endif
 
+fuzz fuzz-clean fuzz-coverage fuzz-lto :
+	@make -f $(TOPDIR)/include/fuzz.mk \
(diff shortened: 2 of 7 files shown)


jsetje (approved):
Did you mean to have the generic "clean" and "all" in fuzz.mk?

Otherwise this looks ok, I do get noise from -print-multiarch on a system that doesn't support it, although I did confirm it's just noise.

vathpela:
> Did you mean to have the generic "clean" and "all" in fuzz.mk?

Yeah - those wind up appending the build dependency, so e.g. `clean` will automatically rebuild `fuzz-clean`.
 
> Otherwise this looks ok, I do get noise from -print-multiarch on a system that doesn't support it, although I did confirm it's just noise.

I don't think I'm going to fix "clang made this argument print a warning instead of silent when there's nothing to do" in this one, especially since you don't need to run that to ship it at all.

dennis-tseng99:
@vathpela How are you doing. Because ```.ascii``` directive expects zero or more string literals separated by **commas** (http://web.mit.edu/gnu/doc/html/as_7.html), rather than by **space**, could you change sbat_var.S file something like **fig-1** ? So that the compiling error(**fig-2**) can be avoided with **clang** when ```make fuzz```. Thanks so much.

**[fig-1]**
image

**[fig-2]**
image

PR #587, opened 2023-07-08
mok: Avoid underflow in maximum variable size calculation

alpernebbi:
The code that mirrors MOK database to EFI variables gets the remaining variable storage size from the firmware and subtracts the size needed for any overhead to see if there is enough space to create a new entry.

However these calculations are on unsigned integer types, they can underflow and result in huge values when the firmware is about to run out of usable variable space. Explicitly check against this.


jsetje:
@chrisccoulson is setting max_var_sz to 0 the right thing to do in this case? It would be new behavior, but it also might be the sane thing to do if we're super tight.

vathpela:
Looks good to me.

PR #594, opened 2023-07-31
pe: Enhance debug report for update_mem_attrs

jongwu:
When memory attributes cannot be updated due to misalignment with 4K or when the size is 0, the debug printout lacks sufficient clarity to indicate the issue. To enhance troubleshooting, it is crucial to generate an error log that explicitly states the failure to execute the expected action. This error log will be visible even when the debug level log is not enabled, thereby significantly reducing debugging time.


jongwu:
Hello @vathpela -, any comments?

vathpela:
There are three aspects of this I don't like:
- The new message says it failed, but that's not necessarily true, which is why it's already separate from actually displaying an error
- I'd rather leave the concise print intact and add another to say specifically which value may be problematic
- holy crap that is one long line of text being printed.

PR #604, opened 2023-08-22
Generate Authenticode for the entire PE file

esnowberg:
The Authenticode is a hash calculation that excludes parts of the PE file that are altered in the signing process itself.  The Authenticode for a PE file should be the same for both the signed and unsigned versions. Both sbsign and pesign have chosen to modify a portion of the PE file that falls outside the actual digital signature.  This is done by zero padding the end of the PE file to align the signature table that is later added.  After doing this padding, the new zero padded data is included within the Authenticode hash calculation.

Both pesign and sbsign can display the Authenticode for an unsigned binary with the padding included within the calculation.  Adding this hash to the MOK does not allow the program to run.  The pesign program also has an option to generate the Authenticode without the padding included. Adding this hash to the MOK also does NOT allow the program to run. When shim finds a PE file without a digital signature, it completely stops calculating the hash towards the end of the file.  Part of the file is excluded. Testing has shown that the last 3K of the file can be omitted from the calculation.

If the Authenticode is generated using Shim’s MokManager, it will calculate a hash without the last part and allow the program to run. Since the end of the file is not included within the calculation, other things could be added.

Fix all this by hashing the entire file that is outside the digital signature to calculate the Authenticode.  Also add zero padding when necessary to do the Authenticode calculation.  If the program is signed, this code should never be referenced.  However, if this code is entered by a signed PE file, there is potentially something nefarious going on.

link: https://blog.hansenpartnership.com/problems-with-tianocore-after-multi-sign-r14141-fixed/
link: https://github.com/osresearch/sbsigntools/commit/370abb7c49ec2a600f64fcbd441d9297124a5cb7
link: https://git.kernel.org/pub/scm/linux/kernel/git/jejb/sbsigntools.git/commit/?id=592ec2188f7b9cf003fe7cb0835e93559f19156f

### DIFF
--- a/pe.c
+++ b/pe.c
@@ -285,14 +285,16 @@ generate_hash(char *data, unsigned int datasize,
 			goto done;
 		}
 
-#if 1
-	}
-#else // we have to migrate to doing this later :/
 		SumOfBytesHashed += hashsize;
 	}
 
-	/* Hash all remaining data */
-	if (datasize > SumOfBytesHashed) {
+	/* Hash all remaining data. If SecDir->Size is > 0 this code should not
+	 * be entered.  If it is, there are still things to hash.  For a file
+	 * without a SecDir, we need to hash what remains. */
+	if (datasize > SumOfBytesHashed + context->SecDir->Size) {
+		char padbuf[8];
+		ZeroMem(padbuf, 8);
+
 		hashbase = data + SumOfBytesHashed;
 		hashsize = datasize - SumOfBytesHashed;
 
@@ -306,8 +308,17 @@ generate_hash(char *data, unsigned int datasize,
 		}
 
 		SumOfBytesHashed += hashsize;
+		hashsize = ALIGN_VALUE(SumOfBytesHashed, 8) - SumOfBytesHashed;
+
+		if (hashsize) {
+			if (!(Sha256Update(sha256ctx, padbuf, hashsize)) ||
+			    !(Sha1Update(sha1ctx, padbuf, hashsize))) {
+				perror(L"Unable to generate hash\n");
+				efi_status = EFI_OUT_OF_RESOURCES;
+				goto done;
+			}
+		}
 	}
-#endif
 
 	if (!(Sha256Final(sha256ctx, sha256hash)) ||
 	    !(Sha1Final(sha1ctx, sha1hash))) {


esnowberg:
> Naively I would have expected firmware and shim to use the same algorithm to determine which part of a file to hash so that the same hashes could be used in db/dbx no matter how a file is loaded. Why isn't that a goal?

This could be done if the first half of the patch was taken.  As I brought up above, we would need to accept the fact that there can be two Authenticodes for the same binary.  

> Are any of sbsign or pesign capable of producing hashes that match what firmware uses?

Yes, you can use pesign without default padding.

` pesign -h -p -i <filename>
`
This hash could be loaded directly into the db.  Tianocore could then boot it directly.  Currently shim would not allow it to load with or without this current patch.

I could change the patch to remove the padding calculation.  But pesign should be modified to not add the pad within the calculation by default.  The added padding pesign does when calculating the hash will only happen with unsigned programs.  When the file is actually signed, the zero padding is written to the file.  Therefore having it reversed in pesign will show the correct Authenticode in both cases.  What may trip up some people is the Authenticodes will not match.

jsetje:
Thanks!

I suspect @chrisccoulson and/or @xnox will have some opinions here since they've been down this road as well.

xnox:
Similar issues are being addressed elsewhere too with slightly different approaches. I think kernel arm64 & riscv64 zimg format tries to intentionally pad the image to ensure that it is always aligned and thus it is never needed to pad things. It is sort of a backwards compatible approach.

I wish we could for example, prohibit unpadded things to be ever verified or loaded. And refuse signing thing that require padding. Such that unsigned, signed, stip signature binaries must be round trip safe. And apply this across the board for all binaries (shim, mm, FB, grub, Linux kernel, UKI, fwupd, fwupd capsules, etc).

This is a general observation that can be implemented "backwards" compatible way. On my phone now, so will look into this in more detail when back on my laptop.

jsetje:
I had the same thought to just enforce padding, but really that ship has sailed. We could start adding that to the builds for various projects that produce artifacts and might get there in some amount of time, but it's not a quick fix.

vathpela:
I pretty strongly feel (not saying I can't be convinced otherwise) that these things are all true:
- shim should be able to match the padded or unpadded binary (especially in dbx/vendor_dbx/mokx/etc)
- that should be true regardless of the presence of signatutes
- that should be true regardless of the presence of padding
- we can ignore "unpadded but signed", it doesn't appear to happen and it doesn't make sense
- the pesign output should default to being the padded version, because people get confused when signing it changes the digest, which it's fundamentally not supposed to do.

Honestly I think edk2 got this wrong, but mostly because Authenticode failed to specify it well.  So I think we have to accept that there are two correct hashes for every binary.

vathpela:
> I wish we could for example, prohibit unpadded things to be ever verified or loaded. And refuse signing thing that require padding. Such that unsigned, signed, stip signature binaries must be round trip safe. And apply this across the board for all binaries (shim, mm, FB, grub, Linux kernel, UKI, fwupd, fwupd capsules, etc).

I wish we could too, but the PE spec and Authenticode just plain got this wrong.

esnowberg:
> I pretty strongly feel (not saying I can't be convinced otherwise) that these things are all true:
> 
> * shim should be able to match the padded or unpadded binary (especially in dbx/vendor_dbx/mokx/etc)
> * that should be true regardless of the presence of signatutes
> * that should be true regardless of the presence of padding

I can see your point here.

> * we can ignore "unpadded but signed", it doesn't appear to happen and it doesn't make sense
> * the pesign output should default to being the padded version, because people get confused when signing it changes the digest, which it's fundamentally not supposed to do.

I completely agree that fundamentally the Authenticode should match.  I guess it comes down to if the Authenticode generated in pesign is only useful when enrolling things into mok or mokx.  For me, what is confusing, is the output of pesign can only be used with signed binaries in EDK2.  Using pesign with an unsigned binary and enrolling it directly into the db or dbx is not going to match up with the generated Authenticode.

> Honestly I think edk2 got this wrong, but mostly because Authenticode failed to specify it well. So I think we have to accept that there are two correct hashes for every binary.

Today, with unsigned binaries, we have a shim that doesn't match up with either of the two Authenticode's being discussed.   It skips the end of the file.  With that said, if shim worked the same way as EDK2.  There wouldn't need to be two Authenticodes being tracked and we wouldn't be breaking anything that shouldn't already be fixed.  But this would require pesign's default behavior to change and the zero padding to be removed from this patch.

If there is a strong feeling to track both Authenticodes, I can modify this patch to make that work.  It certainly would be more invasive. The Authenticode's pesign displays would only work with shim.  

With the recent popularity of UKIs, I believe Authenticode usage is going to become more common.

jsetje:
It certainly seems unreasonable to change this in edk2 without providing compatibility there, but it may be worth engaging someone there so that we can come up with a combined patch to a less confusing future.

I'm fundamentally in favor of providing compatibility by supporting both methods, at least for a release or two. If the plan is phase out one of the methods, then we should print a warning when it is seen in use.

jsetje:
@vincentjzimmer we really should get this aligned with edk2. I just heard a rumor about an updated authenticode spec that will hopefully be pushed that may provide some answers.

jsetje:
This diff will bring shim in line with https://aka.ms/AuthenticodeSpec as well as edk2 behavior..

This has largely been unnoticed prior to UKIs since it shouldn't be an issue for objects like GRUB or Linux bzImage files, which are aligned by their builds even when they are unsigned. Such aligned images will end up with the same hash before and after this change, so the need for a compatibility mode is questionable.

jsetje (approved):

Issue #606, opened 2023-08-27
Debugging shim in gdb gives 'Dwarf Error: DW_FORM_line_strp used without required section'


mikebeaton:
When attempting to debug shim in gdb, it currently gives the error: `Dwarf Error: DW_FORM_line_strp used without required section`

This appears to be the same issue reported here https://bugs.launchpad.net/ubuntu/+source/linux/+bug/2014231 (also https://github.com/golang/vscode-go/issues/1914 and https://github.com/haikuports/haikuports/issues/4987), which is nothing to do with shim: current versions of gcc produce DWARF5 by default, but current versions of gdb only accept DWARF4.

Applying the gcc flags suggested in those issues (`-gdwarf-4 -gstrict-dwarf`) makes shim debuggable again.

vathpela:
Apologies for the delay here - is this still an issue with gdb?  If so, the change looks good to me.

mikebeaton:
I can confirm that the issue still applies with `gcc (GCC) 14.2.1 20240912 (Red Hat 14.2.1-3)` and `GNU gdb (Fedora Linux) 15.2-2.fc41` on Fedora 41, and the PR still fixes it.

PR #607, opened 2023-08-27
Force gcc to produce DWARF4 so that gdb can use it

mikebeaton:
Addresses #606.

I am not sure if the location where I have added the flags is ideal, though it does make shim debuggable again.

### DIFF
--- a/Make.defaults
+++ b/Make.defaults
@@ -109,7 +109,7 @@ INCLUDES	= -nostdinc \
 
 override DEFAULT_FEATUREFLAGS = \
 		-std=gnu11 \
-		-ggdb \
+		-ggdb -gdwarf-4 -gstrict-dwarf \
 		-ffreestanding \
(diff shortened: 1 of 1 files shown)


supperpiccle:
I ran into this exact problem and the solution here fixed it.  

However, I tested with a version of gdb that supports DWARF 5.  Both gdb 12.1 and 13.1 ran into this issue.  I suspect it's an issue with this piece in the makefile

`%.efi.debug : %.so
ifneq ($(OBJCOPY_GTE224),1)
	$(error objcopy >= 2.24 is required)
endif
	$(OBJCOPY) -D -j .text -j .sdata -j .data \
		-j .dynamic -j .rodata -j .rel* \
		-j .rela* -j .dyn -j .reloc -j .eh_frame -j .sbat \
		-j .sbatlevel \
		-j .debug_info -j .debug_abbrev -j .debug_aranges \
		-j .debug_line -j .debug_str -j .debug_ranges \
		-j .note.gnu.build-id \
		$< $@`

I don't know anything about the differences between version 4 and 5, but I suspect the objcopy there isn't conforming to whatever specification version 5 requires.  Long term, I think the solution would be to run down what's wrong here.

vathpela (approved):

Issue #573, opened 2023-06-07
RFE: instead of invoking exactly grubx64.efi iterate through drop-in dir


poettering:
Currently shim strictly chainloads grubx64.efi (well, at least on x86-64). This is less than ideal for two reasons:

1. not everyone uses grub (i.e. people use systemd-boot too, and currently when using that together with shim rename it grubx64.efi)
2. this is a single point of failure: if for some reason updating the grubx64.efi binary doesn't succeed properly the user will be thrown in mokmanager rather than some more suitable fallback that tries to recover from the hosed updated.

Hence I'd propose the following logic, that is dumb enough to be safe, but avoids both these issues:

enumerate EFI binaries from some to be defined directory (for example `\EFI\shim-payload\*.EFI` or so), sort them via strverscmp() (ideally the modified one the boot loader spec mandates, as per https://uapi-group.org/specifications/specs/boot_loader_specification/), and then invoke the newest binary. if that fails, go for the next, and so on, and so on, until one works.

This should be reasonably simple and robust and decouple shim from the boot loader implementation while increasing safety for participants in this: they can drop a new entry in the dir, and keep a couple of them around and delete the old ones when they see fit. versions never fight for the same filename anymore.

dennis-tseng99:
This must be a good idea. Though, I'm just worried Microsoft is willing to sign those uncertain combinations of shim + (non-grub2 + some proprietary 2nd stage boot loaders) located in ```\EFI\shim-payload\*.EFI```

poettering:
@dennis-tseng99 msft has no problem if people sign other boot loaders. noone at msft is glued to grub.

bluca:
The signing discussion is off-topic here, it's really orthogonal to how binaries are searched, let's focus on this

esnowberg:
@poettering @bluca I sent the following pull request: https://github.com/rhboot/shim/pull/608 that will solve this problem with a configuration file.  I would be interested in your feedback on this approach.

poettering:
Uh, I am not sure what the benefit would be to have a config file for that. (and csv! yikes). it's much simpler to just look for the alternative files and use them, rather than making things indirect.

config files kind mean yu also want to measure them, and so on, it just makes things a ton more complex for little gain.

julian-klode:
[...]

Issue #482, opened 2022-05-25
certmule discovery doesn't work when network booting


vathpela:
In https://github.com/rhboot/shim/commit/35d7378d29b9ad6f664df20efc4121e210859e65, we add a method of discovering wrapped certificates to import for validation, but it relies on opening a directory and doing `efi_status = dir->Read(dir, &buffersize, buffer);`.  This doesn't work on TFTP, HTTP, or HTTPS targets, which don't present directories in a structured way.

jsetje:
Since those files contain an ESL, there really is no need to support this many, at least if this is a special case for transports that don't support readdir.

PR #608, opened 2023-08-31
Add configuration option to boot an alternative 2nd stage

esnowberg:
Add runtime option to load a different 2nd stage boot loader.

Most projects working with Unified Kernel Images (UKI) have one thing in common.  If they want to use a distro signed shim to boot it, they have to rename their UKI to grub{arch}.efi.  Having a UKI named grub can be confusing.

Add the ability for shim to load an optional configuration file.  This new file is called "options.csv".  The configuration file is completely optional. If used, it is located in the same directory as the booted shim.  The "options.csv" file currently allows a single entry.  Other options could be added to it in the future.  The first and only entry in the file is the name of the secondary boot loader shim will load. The "options.csv" file is in Unicode LE format.

This allows a signed shim to directly load a UKI without the need to rename it to grub. Shim's transitive trust is maintained. If the alternative 2nd stage can not be verified, it will not boot.

Example on how to create the options.csv file:

`echo "uki.efi," | iconv -f us-ascii -t utf-16le > options.csv`

Unlike the RFE request here: https://github.com/rhboot/shim/issues/573 a csv file is being used instead of walking through the directory.  Shim already has precedents for loading a csv file.  Also, walking an ESP directory has shown to have problems when network booting with shim.


bluca:
> walking an ESP directory has shown to have problems when network booting with shim.

If that is the issue with discovering what's available, maybe there could be a build-time option to disable it for systems where things are that broken? Both sd-stub and sd-boot walk the ESP in various places, so if you are using either a UKI or sd-boot you'd already be out of luck on such systems.

Maybe an alternative approach: look for grub first at the fixed path, with no filesystem spelunking at first - and only if not found iterate over the directory content? That way if you are using shim+grub there is no change in functionality nor operations, and if you use shim+uki or shim+sd-boot then you'll already get filesystem spelunking in the second stage so there would be no difference.

esnowberg:
> > walking an ESP directory has shown to have problems when network booting with shim.
> 
> If that is the issue with discovering what's available, maybe there could be a build-time option to disable it for systems where things are that broken? Both sd-stub and sd-boot walk the ESP in various places, so if you are using either a UKI or sd-boot you'd already be out of luck on such systems.

Here is a reference to the issue: https://github.com/rhboot/shim/issues/482

Distros use a single build-time option for their shim. 
 
> Maybe an alternative approach: look for grub first at the fixed path, with no filesystem spelunking at first - and only if not found iterate over the directory content? That way if you are using shim+grub there is no change in functionality nor operations, and if you use shim+uki or shim+sd-boot then you'll already get filesystem spelunking in the second stage so there would be no difference.

Where does systemd-stub do directory discovery during boot?  How does  systemd-stub solve this when booting from TFTP, HTTP, or HTTPS?

bluca:
> > Maybe an alternative approach: look for grub first at the fixed path, with no filesystem spelunking at first - and only if not found iterate over the directory content? That way if you are using shim+grub there is no change in functionality nor operations, and if you use shim+uki or shim+sd-boot then you'll already get filesystem spelunking in the second stage so there would be no difference.
> 
> Where does systemd-stub do directory discovery during boot? How does systemd-stub solve this when booting from TFTP, HTTP, or HTTPS?

At least for addons, type 1 configs, drivers, some of these happen in the stub, some in sd-boot. I don't believe there's any particular care taken for network booting, the filesystem protocol is used.

esnowberg:
> > > Maybe an alternative approach: look for grub first at the fixed path, with no filesystem spelunking at first - and only if not found iterate over the directory content? That way if you are using shim+grub there is no change in functionality nor operations, and if you use shim+uki or shim+sd-boot then you'll already get filesystem spelunking in the second stage so there would be no difference.
> > 
> > 
> > Where does systemd-stub do directory discovery during boot? How does systemd-stub solve this when booting from TFTP, HTTP, or HTTPS?
> 
> At least for addons, type 1 configs, drivers, some of these happen in the stub, some in sd-boot. I don't believe there's any particular care taken for network booting, the filesystem protocol is used.

The solution in this pull request is general enough to get to the second stage without the need to do directory discovery.  It would be up to the second stage to determine if it wants to support network booting or not.  Why place a limitation on shim if the next stage currently doesn’t support something?

julian-klode:
I think this is better served by the changes to integrate fallback loader and a patch to disable the creation of boot entries for the fallback loader ones.

bluca:
> > > > Maybe an alternative approach: look for grub first at the fixed path, with no filesystem spelunking at first - and only if not found iterate over the directory content? That way if you are using shim+grub there is no change in functionality nor operations, and if you use shim+uki or shim+sd-boot then you'll already get filesystem spelunking in the second stage so there would be no difference.
> > > 
> > > 
> > > Where does systemd-stub do directory discovery during boot? How does systemd-stub solve this when booting from TFTP, HTTP, or HTTPS?
> > 
> > 
> > At least for addons, type 1 configs, drivers, some of these happen in the stub, some in sd-boot. I don't believe there's any particular care taken for network booting, the filesystem protocol is used.
> 
> The solution in this pull request is general enough to get to the second stage without the need to do directory discovery. It would be up to the second stage to determine if it wants to support network booting or not. Why place a limitation on shim if the next stage currently doesn’t support something?

Because extra config files means extra moving parts, and also extra things that have to be measured. It's much nicer to just do the right thing.

esnowberg:
> Because extra config files means extra moving parts, and also extra things that have to be measured. It's much nicer to just do the right thing.

With the changes currently in this pull request, would the config file really need to be measured?  Currently if someone renames their UKI to grub, this should already be handled in the current measurement.  

In the future if a new option were added, it may be determined if measurement on the config would be appropriate.  But I don’t see why it would be needed at this time.

bluca:
> > Because extra config files means extra moving parts, and also extra things that have to be measured. It's much nicer to just do the right thing.
> 
> With the changes currently in this pull request, would the config file really need to be measured? Currently if someone renames their UKI to grub, this should already be handled in the current measurement.
> 
> In the future if a new option were added, it may be determined if measurement on the config would be appropriate. But I don’t see why it would be needed at this time.

It's an (unsigned) artifact that changes the behaviour of the first stage loader, so that sounds like something that would need to be measured. Besides that, and other than config files being a pain to handle, we also rename UKIs for boot counting, having to keep a config file in sync with that would make it extra fragile and complicated.

esnowberg:
> > > Because extra config files means extra moving parts, and also extra things that have to be measured. It's much nicer to just do the right thing.
> > 
> > 
> > With the changes currently in this pull request, would the config file really need to be measured? Currently if someone renames their UKI to grub, this should already be handled in the current measurement.
> > In the future if a new option were added, it may be determined if measurement on the config would be appropriate. But I don’t see why it would be needed at this time.
> 
> It's an (unsigned) artifact that changes the behaviour of the first stage loader, so that sounds like something that would need to be measured. 

If it is determined appropriate to measure the config file, a follow on patch for shim could do that.  No other entity other than shim would need to be responsible for doing the measurement.

> Besides that, and other than config files being a pain to handle, we also rename UKIs for boot counting, having to keep a config file in sync with that would make it extra fragile and complicated.

The intention of this change was to add a general purpose way for any secondary boot loader to be used with a signed shim, while maintaining shim’s transitive trust.  This is not geared specifically for sd-boot but any secondary boot loader.

It would be good to hear from the shim maintainers on their opinion of adding new code to shim that does discovery on a directory.  It was my believe that adding new code that did this would be a no-go.

bluca:
> The intention of this change was to add a general purpose way for any secondary boot loader to be used with a signed shim, while maintaining shim’s transitive trust. This is not geared specifically for sd-boot but any secondary boot loader.

Sure, I don't know how many other second stages for uefi + sb are out there, but if this cuts out, or at the very least makes it unnecessarily difficult and awkward to use, with sd-stub based UKIs or sd-boot, then that should be a concern, as that's at the very least a good chunk of the total, no?

bluca:
This looks fine as a starting point to me, then I can propose a follow-up PR to auto-discovery if the config file is not found. That should satisfy all constraints: on systems where fragile firmware implementations are used, one can ensure the new config file is provided, and all will be well. For systems where this is not an issue, users can choose to omit it, and do the autodiscovery. Would this work as an approach?

bluca:
Ping - would it be possible to bring this forward?

bluca (approved):

shim.c
```diff
@@ -1513,6 +1513,84 @@ load_certs(EFI_HANDLE image_handle)
...
+	if (EFI_ERROR(efi_status)) {
+		perror(L"Unable to init protocol\n");
+		return efi_status;
+	}
+
+	efi_status = generate_path_from_image_path(li, L"options.csv", &PathName);
+	if (EFI_ERROR(efi_status))
+		goto done;
+
+	console_print(L"Found directory named \"%s\"\n", PathName);
```
vathpela, line 1542: Who is this message for?  Seems like it'll just confuse users.

shim.c
```diff
@@ -1513,6 +1513,84 @@ load_certs(EFI_HANDLE image_handle)
...
+
+	device = li->DeviceHandle;
+
+	efi_status = BS->HandleProtocol(device, &EFI_SIMPLE_FILE_SYSTEM_GUID,
+					(void **) &drive);
+
```
vathpela, line 1548: This needs error checks.

fallback.c
```diff
@@ -4,6 +4,7 @@
  * Copyright Peter Jones <pjones@redhat.com>
  */
 #include "shim.h"
+#include "utils.h"
```
vathpela, line 7: If you move the `shim.h` change from the second commit into this one, this becomes unnecessary.

shim.h
```diff
@@ -186,6 +186,7 @@
 #include "include/simple_file.h"
 #include "include/str.h"
 #include "include/tpm.h"
+#include "include/utils.h"
```
vathpela, line 189: This should be in the other commit.

shim.c
```diff
@@ -1513,6 +1513,84 @@ load_certs(EFI_HANDLE image_handle)
...
+	efi_status = drive->OpenVolume(drive, &root);
+	if (EFI_ERROR(efi_status)) {
+		perror(L"Failed to open fs: %r\n", efi_status);
+		goto done;
+	}
+
+	efi_status = root->Open(root, &ofile, PathName, EFI_FILE_READ_ONLY, 0);
+	if (EFI_ERROR(efi_status)) {
```
vathpela, line 1556: Shouldn't this special-case `EFI_NOT_FOUND`?

load-options.c
```diff
@@ -6,6 +6,7 @@
 #include "shim.h"
 
 CHAR16 *second_stage;
+CHAR16 *optional_second_stage;
```
vathpela, line 9: I think this needs to be initialized to NULL.

vathpela (changes requested):
Mostly looks good, but a few problems do need to be addressed.

load-options.c
esnowberg: I'll initialize it to NULL in the next version

shim.c
esnowberg: If you mean to not display the error message here if the file is not found, I'll fix that in the next version.

shim.c
esnowberg: I'll remove it.  This was just for my debugging and should have been removed.

shim.c
esnowberg: I'll add the error check here

fallback.c
esnowberg: I'll fix this in the next version

shim.h
esnowberg: I'll move this to the other commit

vathpela:
I pushed another tiny change here - added a dprint() and made the console_print() on success a dprint() as well.

vathpela (approved):

PR #611, opened 2023-09-06
Null-terminate 'arguments' in fallback

vittyvk:
In case CSV entry contains boot argument (e.g. an image to load for shim) it must be null-terminated. While populate_stanza() makes sure 'arguments' end with '\0', add_boot_option() doesn't account for it in 'size' calculations. E.g. for the following CSV entry:

```
 shimx64.efi,6.6.0-0.rc0.20230904git708283abf896.6.fc40.x86_64,\EFI\Linux\5f93b3c9cf1c488a99786fb8e99fb840-6.6.0-0.rc0.20230904git708283abf896.6.fc40.x86_64.efi,Comment
```

the resulting variable after 'fallback' looks like:

```
 # hexdump /sys/firmware/efi/efivars/Boot0004-8be4df61-93ca-11d2-aa0d-00e098032b8c | tail -3
 0000180 0038 0036 005f 0036 0034 002e 0065 0066
 0000190 0069
 0000192
```

Add trailing '\0' to 'size' calculations in add_boot_option() when 'arguments' is not empty. The resulting variable looks like:

```
 # hexdump /sys/firmware/efi/efivars/Boot0004-8be4df61-93ca-11d2-aa0d-00e098032b8c | tail -3
 0000180 0038 0036 005f 0036 0034 002e 0065 0066
 0000190 0069 0000
 0000194
```

and the specified image is loaded by shim without issues.

### DIFF
--- a/fallback.c
+++ b/fallback.c
@@ -226,9 +226,11 @@ add_boot_option(EFI_DEVICE_PATH *hddp, EFI_DEVICE_PATH *fulldp,
 
 		void *var = LibGetVariable(varname, &GV_GUID);
 		if (!var) {
+			int arg_size = StrLen(arguments) ? StrLen(arguments) * sizeof (CHAR16) +
+				sizeof (CHAR16) : 0;
 			int size = sizeof(UINT32) + sizeof (UINT16) +
 				StrLen(label)*2 + 2 + DevicePathSize(hddp) +
-				StrLen(arguments) * 2;
+				arg_size;
 
 			CHAR8 *data, *cursor;
 			cursor = data = AllocateZeroPool(size + 2);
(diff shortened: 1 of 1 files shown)


vittyvk:
ping?

julian-klode:
FWIW, the message is misleading you into thinking it's a single \0 byte, but it's a \0 UCS-2 character (i.e. 2 bytes).

But yes I believe this is useful. It's not strictly necessary: As I'm sure you know, you can always append a space to the entry in the .csv and it works. But it's an ugly workaround for sure.

vittyvk:
The workaround using extra space is indeed known and is already used in the wild, e.g.
https://gitlab.com/kraxel/virt-firmware/-/commit/77d3801ccc963da747ba576da0fe586d352615d3
but it probably takes some time to discover it. It'd be nice to have a permanent solution.

PR #613, opened 2023-09-15
Process OVERRIDE_SECURITY_POLICY in lib/Makefile

mikebeaton:
Without this `make OVERRIDE_SECURITY_POLICY=1` fails with

```
ld: shim.o: in function `install_shim_protocols':
/home/mjsbeaton/shim_source/shim.c:1354: undefined reference to `security_policy_install'
ld: shim.o: in function `uninstall_shim_protocols':
/home/mjsbeaton/shim_source/shim.c:1376: undefined reference to `security_policy_uninstall'
```

Fixes issue #596 and is an alternative to PR #600.

### DIFF
--- a/lib/Makefile
+++ b/lib/Makefile
@@ -31,6 +31,10 @@ ifeq ($(ARCH),arm)
 DEFINES		+= -DMDE_CPU_ARM
 endif
 
+ifneq ($(origin OVERRIDE_SECURITY_POLICY), undefined)
+	DEFINES	+= -DOVERRIDE_SECURITY_POLICY
+endif
+
 LDFLAGS		= -nostdlib -znocombreloc
 
 


dennis-tseng99:
Consider the code expanding in the near future, I still suggest we should make use of "export DEFINES" in Make.defaults to pass down variable(s) to sub-make. Please refer our original discussion https://github.com/rhboot/shim/pull/600

mikebeaton:
Just `export DEFINES` anywhere inside `Make.defaults` but not inside a conditional also sounds fine to me. Tyvm.

Am intrigued what the changes are in the near future.

As mentioned in the related issue, it'd be very appreciated if anyone (cc @jejb @vathpela @bluca - if I may) can point to any public discussion (or provide any info here) of the status of `OVERRIDE_SECURITY_POLICY` (&/or any issues  with using it). As far as I can make out, it dates back to the original [Linux Foundation `Preloader.efi`](https://blog.hansenpartnership.com/linux-foundation-secure-boot-system-released/), and pretty soon after being added to Shim [was made conditional](https://github.com/rhboot/shim/commits/main/lib/security_policy.c) (apparently with the issue being addressed here always present, although I may be misunderstanding something). However it seems like it would be an essential feature for any third-party second stage bootloader (for instance a BLSpec compliant one) which wants to:

 - Defer to distro certificates (rather than requiring user-signing of everything)
 - Use SBAT revocations
 - Not implemented own PE loader
 
Since, AFAICT, a user-build of Shim is the only way to achieve this at the moment (e.g. tentative approach to exactly this is [here](https://github.com/acidanthera/OpenCorePkg/tree/master/Utilities/ShimUtils)).

Issue #370, opened 2021-05-06
It doesn't seem to find grubx64.efi under certain conditions.


jc-lab:
* PC Model : DELL Latitude 3510

I replaced `/EFI/Microsoft/bootmgfw.efi` with shimx64.efi. And Here have grubx64.efi in the same location.

However, I have seen a screen like the one below.
![image](https://user-images.githubusercontent.com/30275706/117378492-86276b80-af10-11eb-807a-267e1c894b5c.png)
```
Failed to open \EFI\Microsoft\Boot\S - Invalid Parameter
Failed to load image \EFI\Microsoft\Boot\S: Invalid Parameter
start_image() returned Invalid Parameter
```

Is there any character limit, etc.?
I also don't know why the 'S' goes into the path shim looking for.

miray-tf:
My guess is that shim uses some information of the LoadOptions.
Parsing for that information is done in set_second_stage() in shim.c in the source.

You might get some more information from shim if you manage to boot a linux system on that laptop and enable shim verbose move with 
mokutil --set-verbosity true

The BOOTXXXX EFI vafiables might also provide some information.

jc-lab:
I already ran it with verbose, but didn't find anything else.

Here is the modified code to print a few:
https://gist.github.com/jc-lab/dfeec0765bff8af97da51243a6a73efe

![image](https://user-images.githubusercontent.com/30275706/117418067-f60c1500-af55-11eb-9cbc-1fd75e2f1d25.png)
li->LoadOptions doesn't seem to be normal.
Is the laptop firmware the problem?

miray-tf:
It could be a problem with the firmware or could be that Windows has set some boot options. 

You seem to be able to recompile shim for testing, so you could try to use the dhexdumpat() function create a hex dump of the LoadOptions.

By the way, in the output from line shim.c:1585 you get the 'start = S' output. That seems to be where the 'S' filename in the load path comes from.

Metabolix:
Windows bcdedit.exe seems to always put "WINDOWS\0" (in 8-bit chars) in the beginning of the optional data, followed by some other MS-specific stuff. It causes shim error "Failed to open \EFI\mypath\䥗䑎坏S - Invalid Parameter". Maybe the easiest fix would be to check for "WINDOWS\0" specifically.

Metabolix:
> When I reproduced this problem the text was a bit different: "S xBCDOBJECT={9dea862c-5cdd-4e70-acc1-f32b344d4795}". 

[...]

PR #620, opened 2023-11-13
We need to check if "loader_str" is an actual path or not.

BogdanAriton:
The string "loader_str" we got from split_load_options() is considered as a path but it could be other things, for example:
		"S xBCDOBJECT={9dea862c-5cdd-4e70-acc1-f32b344d4795}"
		This is a Microsoft BCD object which represents the MS Boot loader:
		BCDEDIT Friendly Name: {bootmgr}
 		Symbolic Name: GUID_WINDOWS_BOOTMGR
		GUID: {9dea862c-5cdd-4e70-acc1-f32b344d4795}
This value is what normally the bootmgfw.efi file is expecting. The shim in some situations could replace bootmgfw.efi for various reasons.

We need to ignore such arguments and use the DEFAULT_LOADER.

Link to opened issue: https://github.com/rhboot/shim/issues/370

### DIFF
--- a/load-options.c
+++ b/load-options.c
@@ -264,7 +264,7 @@ is_valid_path(EFI_LOADED_IMAGE *li, CHAR16 *path)
 	if (EFI_ERROR(efi_status)) {
 		perror(L"Error getting SimpleFileSystemProtocol: %r\n", efi_status);
 		return FALSE;
-    }
+	}
 
 	efi_status = generate_path_from_image_path(li, path, &PathName);
 	if (EFI_ERROR(efi_status)) {
@@ -311,7 +311,7 @@ is_valid_path(EFI_LOADED_IMAGE *li, CHAR16 *path)
 		}
 
 		BS->FreePool(FileInfo);
-    }
+	}
 
 	FileHandle->Close(FileHandle);
 
--- a/load-options.c
+++ b/load-options.c
@@ -249,6 +249,76 @@ done:
 }
 
 /*
+* Check if path is a valid path to a file
+* Even if it's a valid directory we cannot load a directory.
+*/
+static int 
+is_valid_path(EFI_LOADED_IMAGE *li, CHAR16 *path) 
+{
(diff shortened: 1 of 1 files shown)


BogdanAriton:
Thanks for reviewing the code. I've updated the white space issues.
Thanks!

Metabolix:
> If the load option is correct, then use that, if not keep the default value.

That already happens. The necessary checks are already in place (when shim tries to load the file), and now you are just duplicating them in another place. The functional difference is in output. I think you should just quiet the offending error message instead of duplicating all the checks. That's a smaller change and makes the intention clear.

However, I see the goal differently: if the user creates a boot entry with a new path, that's what they want to use, and they want to know if it's not working. The user should have control over their boot options, and only special cases (such as `bcdedit`) prevent it. So instead of ignoring all errors, we should ignore only those where the user is clearly not in control.

Let a mainainer say which option is preferable.

BogdanAriton:
> > If the load option is correct, then use that, if not keep the default value.
> 
> That already happens. The necessary checks are already in place (when shim tries to load the file), and now you are just duplicating them in another place. The functional difference is in output. I think you should just quiet the offending error message instead of duplicating all the checks. That's a smaller change and makes the intention clear.
> 
> However, I see the goal differently: if the user creates a boot entry with a new path, that's what they want to use, and they want to know if it's not working. The user should have control over their boot options, and only special cases (such as `bcdedit`) prevent it. So instead of ignoring all errors, we should ignore only those where the user is clearly not in control.
> 
> Let a mainainer say which option is preferable.

OK, I'm fine with closing this PR out and just check for WINDOWS\0 case added by bcedit. This is probably the only case, thus far that causes issues which most people don't understand. (like the ones described in the ticket)

vathpela:
I think just handling bcdedit is probably a better way forward.  In either case, this current patch breaks several test cases in "make test", so at the least that would need to be fixed.

vathpela:
I'm going to close this in favor of https://github.com/rhboot/shim/pull/621 .

PR #625, opened 2023-12-23
Improving the robustness of value retention for the variable second_stage

Jurij-Ivastsuk:
in function: shim_init(void)
in function: EFI_STATUS set_second_stage (EFI_HANDLE image_handle)
second_stage is set to \\grubx64.efi
in the further course of time:
in function: efi_status = init_grub(image_handle);
EFI_STATUS init_grub(EFI_HANDLE image_handle)
second_stage variable - no longer has any value !!! the value \\grubx64.efi has been lost
see attached screen foto:
IMG_5120
In order to improve the robustness of value retention for the variable second_stage we have done following changes:
in /shim/include/load-options.h
removed line with: extern CHAR16 *second_stage;
in /shim/shim.c
we set new variable just after the #define OID_EKU_MODSIGN "1.3.6.1.4.1.2312.16.1.2"
static CHAR16 *second_stage;
After making this improvement, I didn't get an error message when booting anymore.

### DIFF
--- a/include/load-options.h
+++ b/include/load-options.h
@@ -12,7 +12,6 @@ EFI_STATUS generate_path_from_image_path(EFI_LOADED_IMAGE *li,
 
 EFI_STATUS parse_load_options(EFI_LOADED_IMAGE *li);
 
-extern CHAR16 *second_stage;
 extern void *load_options;
 extern UINT32 load_options_size;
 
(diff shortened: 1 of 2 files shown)


julian-klode:
It's quite obviously not the solution to make it static in shim.c, it's just going to cause wherever else it is defined with external linkage to have a different value and stuff not to work at all.

Jurij-Ivastsuk:
@julian-klode Thank you very much for your review! This really isn't a solution for everyone. By redeclaring the "second_stage" variable from external to static and defining it only in shim.c, we have limited the visibility of the variable. This variable was only visible in the translation unit of shim.c. And was no longer visible in the translation unit of load-options.c, for example. But that really helped me to solve the problem. The next pull request is coming soon.

PR #629, opened 2024-01-23
Fixing EFI_CERT_foo in docs

SherifNagy:
Fixing a minor document error


vathpela:
I don't think this is right?  I think Paul was intending to compare the proposed EFI_CERT_SBAT to, for example, EFI_CERT_TYPE_X509_GUID, EFI_CERT_RSA2048_GUID, etc., without having to list them all.

SherifNagy:
Ah got what you mean, was a bit confusing when I read it :)

PR #643, opened 2024-02-29
Fix parallel build of gnu-efi

jepio:
When building with make 4.4.1, gnu-efi may be built twice and in parallel:

  $ make -n -j ARCH=x86_64 2>&1 | tee build.log
  $ grep -e 'make -C gnu-efi' build.log
  make -C gnu-efi \
  make -C gnu-efi \

This has been seen to cause linking failures when building libgnuefi.a because some object files may end up truncated.

The reason for this is that make interprets multiple targets in the same rule as independent and can run the rule multiple times. The solution is to define a grouped target with &:, which causes make to behave the way one expects: runs the rule once and expects it to create all targets.

Grouped target support was added in make 4.3 released 4 years ago.


vathpela:
"4 years ago" does indeed _seem_ like it's a long time, but quite a few of our consumers are stuck (for now) on make 3.82.  Is there another way we can avoid this behavior?

jsetje:
I'm not sure I follow why back grounding that invocation would make things more rather than less deterministic.

jepio:
This isn't backgrounding, it's defining a grouped target:
https://www.gnu.org/software/make/manual/html_node/Multiple-Targets.html

PR #644, opened 2024-02-29
Multiple certificate handling improvements

rosslagerwall:
A couple of improvements for performing verification with multiple certificates.

### DIFF
--- a/shim.c
+++ b/shim.c
@@ -1488,6 +1488,7 @@ load_cert_file(EFI_HANDLE image_handle, CHAR16 *filename, CHAR16 *PathName)
 	EFI_SIGNATURE_LIST *certlist;
 	void *pointer;
 	UINT32 original;
+	UINT32 offset;
 	int datasize = 0;
 	void *data = NULL;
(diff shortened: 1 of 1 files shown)


vathpela:
Is there existing usage of these packed certificates?  It would seem like the obvious way to do this is multiple EFI_SIGNATURE_LIST entries each with an EFI_CERT_TYPE_X509_GUID entry.  If something else is already supporting the packed version, though, I don't have a big problem with it.

rosslagerwall:
> Is there existing usage of these packed certificates? It would seem like the obvious way to do this is multiple EFI_SIGNATURE_LIST entries each with an EFI_CERT_TYPE_X509_GUID entry. If something else is already supporting the packed version, though, I don't have a big problem with it.

Nothing that I am aware of. On reflection, that change is unnecessary so I've dropped the commit. The remaining commits fix a leak and allow having multiple EFI_SIGNATURE_LIST entries each with an EFI_CERT_TYPE_X509_GUID entry.

Thanks

PR #646, opened 2024-03-19
Validate that a supplied vendor cert is not in PEM format

### LINKED ISSUE #645: Fail the build if VENDOR_CERT_FILE is PEM-encoded
A common failure mode in shim reviews is people embedding certificates which are PEM-encoded rather than DER-encoded. It's a very easy mistake to make, and easy to miss in reviews too.

I've added an extra message in https://github.com/rhboot/shim-review/pull/402 , but it would be even nicer if the shim build process would notice this mistake and fail the build with an appropriate error.

steve-mcintyre:
If we see "BEGIN CERTIFICATE", it's a PEM certificate and won't work. Fail the build early and say so.

Fixes #645

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -69,11 +69,19 @@ ifneq ($(origin FALLBACK_VERBOSE_WAIT), undefined)
 	CFLAGS += -DFALLBACK_VERBOSE_WAIT=$(FALLBACK_VERBOSE_WAIT)
 endif
 
-all: confcheck $(TARGETS)
+all: confcheck certcheck $(TARGETS)
 
 confcheck:
 ifneq ($(origin EFI_PATH),undefined)
 	$(error EFI_PATH is no longer supported, you must build using the supplied copy of gnu-efi)
+endif
+
+certcheck:
+ifneq ($(origin VENDOR_CERT_FILE), undefined)
+	@if grep -q "BEGIN" $(VENDOR_CERT_FILE); then \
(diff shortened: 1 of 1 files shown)


vathpela:
Probably should just check for "BEGIN"?  There's other PEM stuff people might be dumb enough to shove in there.

steve-mcintyre:
> Probably should just check for "BEGIN"? There's other PEM stuff people might be dumb enough to shove in there. I don't know why someone would shove e.g. a GPG pubkey in there, but I have faith in humanity's abilities.

Done!

jsetje:
I guess we can't count on siglist tools to be installed? Otherwise we could test the siglist with them explicitly.

steve-mcintyre:
> I guess we can't count on siglist tools to be installed? Otherwise we could test the siglist with them explicitly.

Yeah, deliberately trying to stay simple here, hence not even using `file` to check the contents.

jsetje:
I'm happy with not creating additional build dependencies, since I really don't want to have to fix them for 7. :)

The complete version would probably be to use openssl to validate vendor cert and the siglist tools to validate vendor db

jsetje (approved):

steve-mcintyre:
> I'm happy with not creating additional build dependencies, since I really don't want to have to fix them for 7. :)
> 
> The complete version would probably be to use openssl to validate vendor cert and the siglist tools to validate vendor db

Sure, we can revisit this to add more features. I'd just like to get something in to catch the obvious failures now. :-)

PR #651, opened 2024-04-05
sbat: Add grub.peimage,2 to latest (CVE-2024-2312)

julian-klode:
Add the previous latest level to the switch for automatic.

### DIFF
--- a/include/sbat_var_defs.h
+++ b/include/sbat_var_defs.h
@@ -47,6 +47,8 @@
 #define SBAT_VAR_AUTOMATIC_REVOCATIONS "shim,2\ngrub,3\n"
 #elif SBAT_AUTOMATIC_DATE == 2023012900
 #define SBAT_VAR_AUTOMATIC_REVOCATIONS "shim,2\ngrub,3\ngrub.debian,4\n"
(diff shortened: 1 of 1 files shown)


julian-klode:
I guess to note:

- This only affects Debian testing, unstable, Ubuntu mantic and the in-development noble
- I've not bumped the default previous level, but recorded the previous latest level as a possible previous one

julian-klode:
There's a question if we should also bump that latest to `grub,4` now and drop the `grub,debian` entry. It's been what, 3 months now? We should do it before we get another round of CVEs and need to bump grub again, otherwise `grub` is going to be at `grub,3` in previous and `grub,5` in latest which would be messy.

kukrimate:
I think latest should be "grub,4".

I think ideally latest should have been "grub,4" at the time of the 15.8, it would make sense that the _actually secure_ revocations at the time of a specific shim release are always available as an opt in. In this case it made sense for people not affected by the NTFS CVE but still.

vathpela (approved):

vathpela:
This looks good to me, as would updating to `grub,4` and removing `grub.debian` in a different PR.

PR #656, opened 2024-04-18
Loader protocol (with Device Path resolution support)

kukrimate:
This rebases the current loader protocol work on main and adds support for loading executables if only a device path is provided. Currently the EFI_SIMPLE_FILE_SYSTEM_PROTOCOL and EFI_LOAD_FILE2_PROTOCOL are supported, which should be sufficient for most users.


loader-proto.c
```diff
@@ -0,0 +1,411 @@
...
+	bprop->buffer = NULL;
+
+	/* look for a handle with LF2 support from the input DP */
+	bprop->dp = dp;
+	status = BS->LocateDevicePath(&EFI_LOAD_FILE2_GUID, &bprop->dp, &bprop->hnd);
+	if (EFI_ERROR(status))
+		goto out;
+
+	/* find protocol */
+	status = BS->LocateProtocol(bprop->hnd, &EFI_LOAD_FILE2_GUID, (void **) &lf2);
```
kukrimate, line 129: This should be BS->HandleProtocol()

loader-proto.c
```diff
@@ -0,0 +1,411 @@
...
+
+	if (bprop.buffer)
+		FreePool(bprop.buffer);
+
+	return EFI_SUCCESS;
+
+free_alloc:
+	BS->FreePages(image->alloc_address, image->alloc_pages);
+free_image:
+	FreePool(image);
```
kukrimate, line 240: We should also FreePool(image->li.FilePath) and FreePool(image->loaded_image_device_path).

loader-proto.c
kukrimate: Fixed now.

include/lf2.h
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+///
+/// Protocol Guid defined by UEFI2.1.
+///
+#define LOAD_FILE2_PROTOCOL  EFI_LOAD_FILE2_PROTOCOL_GUID
+
+typedef struct _EFI_LOAD_FILE2_PROTOCOL EFI_LOAD_FILE2_PROTOCOL;
+
+/**
```
jsetje, line 30: Nit: Please use some sort of a comment style that makes it clear that this is a multi-line comment.

/*
 * XXX
 */

loader-proto.c
```diff
@@ -0,0 +1,436 @@
...
+
+	/* find protocol */
+	status = BS->HandleProtocol(bprop->hnd, &EFI_LOAD_FILE2_GUID, (void **) &lf2);
+	if (EFI_ERROR(status))
+		goto out;
+
+	/* get file size */
+	bprop->size = 0; /* this shouldn't be read when Buffer=NULL but better be safe */
+	status = lf2->LoadFile(lf2, bprop->dp, /*BootPolicy=*/false, &bprop->size, NULL);
+	/* NOTE: the spec is somewhat ambiguous what is the correct return status code
```
jsetje, line 136: Nit: Comment format.

jsetje:
This is actually looking pretty good. Thank you for getting it this far!

Some of the comment style is unfamiliar to me and may be fine, but it might be worth using a single style for the multi-line ones. That's really a nit though.

I have gotten this to work, but will do more testing with this on Monday.

jsetje:
When trying to load a kernel with an untrusted signature, we end up with the following error message:

`Verification failed: Security Policy Violation
error: bad shim signature.
error: you need to load the kernel first.        
Press any key to continue...`

"bad shim signature" seems misleading.

shim.c
```diff
@@ -1908,6 +1915,91 @@ devel_egress(devel_egress_action action UNUSED)
...
+		efi_status = image->exit_status;
+	}
+
+	//
+	// We only support EFI applications, so we can unload and free the
+	// image unconditionally.
+	//
+	BS->UninstallMultipleProtocolInterfaces(ImageHandle,
+	                                &EFI_LOADED_IMAGE_GUID, image,
+					&SHIM_LOADED_IMAGE_GUID, &image->li,
```
ardbiesheuvel: These appear to be the wrong way around.

PR #657, opened 2024-04-18
tpm: Don't return an error if event log is full

kukrimate:
Don't return an error if the firmware's event log is full. The PCR should still be extended in this case so there is no reason to make the system unbootable.

As discussed previously, it seems like we shouldn't fail to boot if the firmware's (limited size) TPM event log fills up. The events should still cause the target PCRs to be extended in this case.

Maybe we should warn on console and pass this information to the OS too?

### DIFF
--- a/tpm.c
+++ b/tpm.c
@@ -201,11 +201,18 @@ static EFI_STATUS tpm_log_event_raw(EFI_PHYSICAL_ADDRESS buf, UINTN size,
 			*/
 			efi_status = tpm2->hash_log_extend_event(tpm2,
 				PE_COFF_IMAGE, buf, (UINT64) size, event);
+			/* don't return an error if the firmware's event log is
+			 * full. the PCR should still be extended in this case
+			 * so there is no reason to make the system unbootable */
+			if (efi_status == EFI_VOLUME_FULL)
+				efi_status = EFI_SUCCESS;
 		}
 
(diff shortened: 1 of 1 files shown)


vathpela:
Yeah, I'd like to see it logged once.  Otherwise this looks reasonable.

kukrimate:
@vathpela Added the warning, and it should cover all places the EFI_VOLUME_FULL thing can be returned to represent this condition.

marosg42:
@kukrimate in https://bugs.launchpad.net/ubuntu/+source/shim/+bug/2089320 you mentioned this should solve it. Any chance this can get merged?

sgtsnookums (approved):
Ran into this exact underlying issue this week, found the issue and this PR in github, and tried applying this patch, which solved my issue booting with the latest upstream UEFI dbx entries. Take my thumbs up.

NHDLarry:
I ran into this recently, when coincidentally, I enabled TPM2.0 and updated Windows 10 in a dual boot configuration. I presumed that it was the SBAT issue, exacerbated by the Asus X99 UEFI refusing to work with Mokutil. It had been working well for a very long time prior to that, rarely running W10.  On closer examination of this issue and some related reports, I'm certain this is it. It appears MSFT not to blame.  Will build and confirm, but hopefully, this will make life easier and get into distribution soon.

vathpela (approved):

PR #658, opened 2024-04-28
Add LoongArch support

yetist:
This PR can add LoongArch support for shim, but first requires updating the gnu-efi subproject to the latest commit.


gmbr3 (changes requested):
Master can't be used for shim purposes due to changes
Shim needs to rebase gnu-efi first
Blocked by rhboot/gnu-efi#8

vathpela:
I don't think we have any interest in adding another arch that doesn't have support for `objcopy --target efi-app-FOO` in binutils.  Last time we accepted this it was an unmitigated disaster.  Get that done and make it use that instead.

PR #659, opened 2024-04-30
undo change that limits certificate files to a single file

jsetje:
In the ea0f9df "Allow SbatLevel data from external binary" patch the code that tries to load any certificate file that starts with shim_certificate was changed to only accept shim_certificate.efi which defeats the entire point of reading through the directory. This change reverts that.


jsetje:
This needs testing.

jsetje:
Unsurprisingly testing confirms that this restores the previous behavior and shim will load files named:

shim_certificate_001.efi shim_certificate_002.efi etc.

PR #661, opened 2024-05-23
Apply the EKU checks

dennis-tseng99:
Implement the CodeSign EKU check to fulfill the requirements of NIAP OS_PP. Also modify the ModSign EKU check to use VerifyEKUsInPkcs7Signature() to check the signer certificate instead of the certificate directly from the key database.

The EKU check can be enabled by setting ENABLE_EKU_CODESIGN=1 when make. This commit supersedes PR-232 which was closed on Jul 1, 2021.

 Signed-off-by: Gary Lin <glin@suse.com> and
                           Dennis Tseng <dennis.tseng@suse.com>


dennis-tseng99:
To not change the original author's commit, this PR will be closed and replaced by PR#664 (Apply EKU check with compile option).

PR #663, opened 2024-06-04
Provide better error message when MokManager is not found

rmetrich:
If MokManager has to be entered but system is booting on disk on `EFI/BOOT/BOOTx.EFI` entry, MokManager cannot be found because it's not in that directory.
This indicates an issue with the BootOrder or the UEFI firmware is just not taking BootOrder into account (seen on Lenovo ThinkPad P1 Gen 6 and VMWare).
This patch prints a related message and reboots after 10 seconds.
    
### Reproducer

1. Import a certificate using mokutil
2. Tell UEFI to boot on BOOTX64.EFI entry on next boot
    
### Result without the patch (with verbosity enabled)

~~~
mok.c:1045:import_mok_state() checking mok request
shim.c:866:load_image() attempting to load \EFI\BOOT\mmx64.efi
Failed to open \EFI\BOOT\mmx64.efi - Not Found
Failed to load image 貘給: Not Found
shim.c:888 load_image() Failed to open \EFI\BOOT\mmx64.efi - Not Found
shim.c:1115 read_image() Failed to load image 貘給: Not Found
Failed to start MokManager: Not Found
mok.c:1047:import_mok_state() mok returned Not Found
Something has gone seriously wrong: import_mok_state() failed: Not Found
~~~

### Result with the patch

Box with title "Could not find MokManager" and message "Boot Order must be misconfigured or not honored by the UEFI firmware.", then system rebooted after 10 seconds (in hope the BootOrder will be good next time).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,10 +38,10 @@ CFLAGS += -DENABLE_SHIM_CERT
 else
 TARGETS += $(MMNAME) $(FBNAME)
 endif
-OBJS	= shim.o globals.o mok.o netboot.o cert.o replacements.o tpm.o version.o errlog.o sbat.o sbat_data.o sbat_var.o pe.o pe-relocate.o httpboot.o csv.o load-options.o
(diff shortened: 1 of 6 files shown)


vathpela (approved):
I've made a few minor changes, but basically this looks good to me.

PR #664, opened 2024-06-06
Apply EKU check with compile option

dennis-tseng99:
Implement the CodeSign EKU check to fulfill the requirements of NIAP OS_PP.
Also modify the ModSign EKU check to use VerifyEKUsInPkcs7Signature() to
check the signer certificate instead of the certificate directly from the
key database.
    
These commits supersede the PR#232 and PR#661 (Apply the EKU checks) and
separate the original author's codes from other commit so that the author's
original codes can be quite independent.

The author also answered the question in PR#232 and  changed the conditional
statement to EFI_Status != EFI_SUCCESS right after VerifyEKUsInPkcs7Signature()
in Cryptlib/Pk/CryptPkcs7Verify.c

User can optionally enable codesign EKU check in compiling time.
To enable the codesign check, set ENABLE_CODESIGN_EKU to 1.
To disable the codesign check, set ENABLE_CODESIGN_EKU to 0
or just omit this flag.
    
For example:
make xxxx ENABLE_CODESIGN_EKU=1 xxxx shim.efi

### DIFF
--- a/Cryptlib/Pk/CryptPkcs7VerifyEku.c
+++ b/Cryptlib/Pk/CryptPkcs7VerifyEku.c
@@ -1,0 +1,516 @@
+/** @file
+  This module verifies that Enhanced Key Usages (EKU's) are present within
+  a PKCS7 signature blob using OpenSSL.
+
+  Copyright (C) Microsoft Corporation. All Rights Reserved.
+  Copyright (c) 2019, Intel Corporation. All rights reserved.<BR>
+
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+
+**/
+
+#include <Base.h>
+#include "InternalCryptLib.h"
+#include <openssl/x509v3.h>
+#include <openssl/asn1.h>
+#include <openssl/x509.h>
+#include <openssl/bio.h>
+#include <openssl/x509.h>
+#include <openssl/pkcs7.h>
+#include <openssl/bn.h>
+#include <openssl/x509_vfy.h>
+#include <openssl/pem.h>
+#include <openssl/evp.h>
+#include <openssl/asn1.h>
+
+/**
+  This function will return the leaf signer certificate in a chain.  This is
+  required because certificate chains are not guaranteed to have the
+  certificates in the order that they were issued.
+
+  A typical certificate chain looks like this:
+
+
+                 ----------------------------
+                |            Root            |
+                 ----------------------------
+                               ^
+                               |
+                 ----------------------------
+                |          Policy CA         | <-- Typical Trust Anchor.
+                 ----------------------------
+                               ^
+                               |
+                 ----------------------------
+                |         Issuing CA         |
+                 ----------------------------
+                               ^
+                               |
+                 -----------------------------
+                /  End-Entity (leaf) signer  / <-- Bottom certificate.
+                -----------------------------  EKU: "1.3.6.1.4.1.311.76.9.21.1"
+                                                    (Firmware Signing)
+
+
+  @param[in]   CertChain            Certificate chain.
+
+  @param[out]  SignerCert           Last certificate in the chain.  For PKCS7 signatures,
+                                    this will be the end-entity (leaf) signer cert.
+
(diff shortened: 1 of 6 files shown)


Cryptlib/Pk/CryptPkcs7VerifyEku.c
```diff
@@ -0,0 +1,516 @@
...
+  if (Status != EFI_SUCCESS || SignerCert == NULL) {
+    //
+    // Fail to get the end-entity leaf signer certificate.
+    //
+    Status = EFI_INVALID_PARAMETER;
+    goto Exit;
+  }
+
+  Status = CheckEKUs (SignerCert, RequiredEKUs, RequiredEKUsSize, RequireAllPresent);
+  if (Status != EFI_SUCCESS) {
```
mikebeaton, line 494: The correct idiom in EDK-2 code, also mainly used in shim, is `if (EFI_ERROR(Status))`. The reasoning is that various EDK-2 functions can also return warnings (which are positive non-zero values; negative non-zero values are errors). So unless one explicitly wants the code to stop on warnings and errors (not usual) one should use `EFI_ERROR`.

Apologies if I've misunderstood and there's a reason to stop on warnings. If not, the comment applies three times that I can see in this PR.

By the way, I'm an interested outsider, so don't take this as a definitive comment from the shim project, but I believe it's correct!

dennis-tseng99: @mikebeaton Thanks for your comment. Although your suggestion is reasonable, we just want to keep consistency with the newest EDK-2 codes.

dennis-tseng99: Sure, [here](https://github.com/tianocore/edk2/blob/master/CryptoPkg/Library/BaseCryptLib/Pk/CryptPkcs7VerifyEku.c) is the EDK2 codes we referred to. Thanks so much.

mikebeaton: Okay, it is indeed done the same way here. Thank you for the clarification.

vathpela:
What's the rationale for this being conditional at build time?  To me, it seems likely we either always want this or never do, but making it so some builds get it and some don't feels weird.  Help me understand the criteria for choosing when I would want it enabled and disabled?

joeyli:
> What's the rationale for this being conditional at build time? To me, it seems likely we either always want this or never do, but making it so some builds get it and some don't feels weird. Help me understand the criteria for choosing when I would want it enabled and disabled?

Hi vathpela,

Those two patches are for checking the Code Signing extendedKeyUsage field in X.509 certification. (OID 1.3.6.1.5.5.7.3.3) 
It's for supporting NIAP PP_OS spec:

https://www.niap-ccevs.org/MMO/PP/PP_OS_V4.2.1.pdif

FIA_X509_EXT.1 X.509 Certificate Validation
[...snip]
The OS shall validate the extendedKeyUsage field according to the following
rules:
Certificates used for trusted updates and executable code integrity
verification shall have the Code Signing purpose (id-kp 3 with OID
1.3.6.1.5.5.7.3.3) in the extendedKeyUsage field.

And
FPT_TST_EXT.1 Boot Integrity
FPT_TST_EXT.1.1
Application Note: The bootchain of the OS is the sequence of software, to include
the OS loader, the kernel, system drivers or modules, and system files, which
ultimately result in loading the OS. The first part of the OS, usually referred to as the
first-stage bootloader, must be loaded by the platform. Assessing its integrity, while
critical, is the platform's responsibility; and therefore outside the scope of this PP. All
software loaded after this stage is potentially within the control of the OS and is in
scope.

If those patches be applied to shim and user's MOK or embedded key in shim does NOT have Code Signing EKU. Then the verification of second bootloader (grub2) and kernel will fail. Which means that the key can NOT be used to sign grub2 and kernel.

So we provide a build option for shim builder to enable or disable the Code Signing EKU feature when building. If shim builder wants to use this feature, they can add ENABLE_CODESIGN_EKU=1 when running make command.

Thanks!

PR #665, opened 2024-06-13
Load shim_certificates.efi during netboot

rosslagerwall:
Network boot cases do not support reading through a directory but it can still be useful to load unbundled certificates. Try to load them directly.

### DIFF
--- a/shim.c
+++ b/shim.c
@@ -1558,10 +1558,13 @@ load_unbundled_trust(EFI_HANDLE image_handle)
 		/*
 		 * Network boot cases do not support reading a directory. Try
 		 * to read revocations.efi to pull in any unbundled SBATLevel
-		 * updates unconditionally in those cases. This may produce
-		 * console noise when the file is not present.
+		 * updates unconditionally in those cases. Also try to read
+		 * shim_certificate.efi to load additional certificates.
+		 * This may produce console noise when the files are not
+		 * present.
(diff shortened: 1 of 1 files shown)


jsetje:
The naming convention for those files is supposed to be "shim_certificate*.efi", but obviously that won't work without some sort of a readdir. Do you want to support a single, different filename for netboot?

Also, that will generate another line of noise when it's not there. I should probably just break up the interface to properly suppress that message. (I didn't want to risk that at the time when I created the initial netboot revocations.efi changes.)

rosslagerwall:
> Do you want to support a single, different filename for netboot? (Edit: this may well make sense, I just want to make sure it's intentional.)

Apologies for not replying earlier. Yes, I think that it would be useful for Shim to load certificates from a separate file during network boot, even if it has to be a fixed path rather than allowing multiple files.

jsetje:
Hi, I pushed ahead with this as well, and now I should probably figure out how to leverage your patch so you get proper credit.

If you could take a look at this, I'm happy to credit you there or replace the last patch with a variation of this one: https://github.com/rhboot/shim/pull/686/commits

I do want to support loading at least 3-5 (so I did 10) filenames.

rosslagerwall:
I tested #686 and it works for me, so I'm going to decline this PR since yours is a superset of it.

PR #666, opened 2024-06-20
Allow fallback to default loader when encountering errors on network boot

nathan-omeara:
Only certain errors trigger fallback to the default loader name.  This change allows fallback when encountering `EFI_TFTP_ERROR` as well.  And ideally would also handle `EFI_HTTP_ERROR` the same way, but that requires updating gnu-efi to a version newer than the shim-15.8 branch.

This fixes the issue reported in https://github.com/rhboot/shim/issues/649 that prevents boot on some models of PC.

### DIFF
--- a/shim.c
+++ b/shim.c
@@ -1222,10 +1222,15 @@ EFI_STATUS init_grub(EFI_HANDLE image_handle)
 					 use_fb ? FALLBACK : second_stage);
 	}
 
-	// If the filename is invalid, or the file does not exist,
-	// just fallback to the default loader.
+	/*
+	 * If the filename is invalid, or the file does not exist, just fall
+	 * back to the default loader.  Also fall back to the default loader
+	 * if we get a TFTP error or HTTP error.
+	 */
 	if (!use_fb && (efi_status == EFI_INVALID_PARAMETER ||
-	                efi_status == EFI_NOT_FOUND)) {
+	                efi_status == EFI_NOT_FOUND ||
+	                efi_status == EFI_HTTP_ERROR ||
+	                efi_status == EFI_TFTP_ERROR)) {
 		console_print(
 			L"start_image() returned %r, falling back to default loader\n",
 			efi_status);


nathan-omeara:
@vathpela @jsetje 

Could one of you take a look at this PR to address #649 ?

shim.c
```diff
@@ -1262,8 +1262,9 @@ EFI_STATUS init_grub(EFI_HANDLE image_handle)
 	// default loader if we get a TFTP error or HTTP error.
 	if (!use_fb && (efi_status == EFI_INVALID_PARAMETER ||
 	                efi_status == EFI_NOT_FOUND ||
-	                // TODO: Add EFI_HTTP_ERROR when gnu-efi is updated to a version that includes it
-	                // efi_status == EFI_HTTP_ERROR ||
+					// TODO: Change to EFI_HTTP_ERROR when gnu-efi is updated to a version that includes it
+					// I already get error 35 on my test systems, even with this build of shim.
+	                efi_status == EFIERR(35) ||
```
dbnicholson, line 1267: I think it would be nicer to handle this would be with a compatibility macro at the top of this source file (or even in shim.h):

```
// FIXME: Drop this when gnu-efi is updated to include de6f9259e8476495c78babbc25250a59de7f3942.
#ifndef EFI_HTTP_ERROR
#define EFI_HTTP_ERROR EFIERR(35)
#endif
```

Then you can make this conditional more normal and the compatibility macro can be dropped when gnu-efi is updated. It's already on the gnu-efi [shim-15.9 branch](https://github.com/rhboot/gnu-efi/tree/shim-15.9), but the submodule here hasn't been updated.

nathan-omeara: Thank you, I've made this change.

vathpela (approved):
Because we've got EFI_HTTP_ERROR and similar from another PR, I've consolidated these patches into one and done some minor fix-ups in the process.

Looks good to me.

PR #669, opened 2024-07-16
[Comments only] Canonical NX enablement delta

kukrimate:
This is the patch we are planning on using for the next Canonical shim-review submission.

I am posting it here, not necessarily for upstreaming in this exact form, but to give people an opportunity for comments / expressing potential concerns we submit a shim-review proper.


kukrimate:
This already was part of an approved shim-review so closing, it wasn't intended for merging anyways.

PR #671, opened 2024-07-23
Add user facing message when SBAT self check fails

THS-on:
As with the upcoming versions we are going to revoke more and more older shims, I've been testing this more extensively and have found that the current message is not ideal for users who don't normally interact with the shim.

This is an initial attempt at adding additional message with more context. We could also include an URL with additional information, but the question here is if this should be generic or distro specific?


shim.c
```diff
@@ -1966,10 +1966,18 @@ efi_main (EFI_HANDLE passed_image_handle, EFI_SYSTEM_TABLE *passed_systab)
 die:
 		console_print(L"Something has gone seriously wrong: %s: %r\n",
 			      msgs[msg], efi_status);
+
+		/*
+		 * Provide additional information when the SBAT self check fails
+		 */
+		if ( msg == SBAT_SELF_CHECK ) {
+			console_print(L"\nThe current shim has likely been revoked. "
```
steve-mcintyre, line 1974: Can we be sure here? "likely" isn't great?

shim.c
```diff
@@ -1966,10 +1966,18 @@ efi_main (EFI_HANDLE passed_image_handle, EFI_SYSTEM_TABLE *passed_systab)
...
+		 */
+		if ( msg == SBAT_SELF_CHECK ) {
+			console_print(L"\nThe current shim has likely been revoked. " 
+						   "Please update to a newer shim version.\n");
+		}
 #if defined(ENABLE_SHIM_DEVEL)
 		devel_egress(COLD_RESET);
 #else
-		usleep(5000000);
+		usleep(8000000);
```
steve-mcintyre, line 1980: Do we need to change the delay??

THS-on: No, but I found 5 sec to short to read the messages. Another option would be to just count down or wait on user input. Need to see if we can do that nicely.

shim.c
jsetje: The other option is to have the message clearly cover all the cases that could land you here:

1) You are somehow trying to boot an otherwise valid out of date OS image and your shim has been revoked.
2) You are somehow booting an out of date shim (perhaps from the wrong device?)
3) You are somehow trying to boot an old file that was inserted maliciously.
4) You are dealing with a file or EFI variable that was somehow corrupted
5) We ran into an internal error (bug in the shim code)

Although that's enough cases that it may make sense to split them up and handle them appropriately. Certainly verify_sbat_section() could differentiate between EFI_SECURITY_VIOLATION and EFI_ERROR, which would group 1-3 and 4&5.

That wording could probably use some more work, but let me know what you think.

I do want to get something like this merged soon, especially since Windows pre-installed systems are very likely to have SbatLevel pre-applied.

shim.c
jsetje: Since the next thing we do is power down, there's really no reason not so have a delay that's plenty long enough to read the message. My instinct would be for 90 seconds or so.

That's probably more reason to sort the messages as clearly as we can above.

shim.c
THS-on: @jsetje my plan is then the following:
- differentiate errors in verify_sbat_section()  (thanks for the hint with EFI_ERROR)
- propagate which SBAT rule check failed: Something like:  expected at least level 5  for "shim" got "4"
- Output error messages similar to the ones you suggested depending on the information (sbat is wrong, something else failed)

jsetje: Printing the exact levels may be more than we need here. In the end a customer needs to know which package to update, and the package versions have nothing to do with the sbat label.

shim.c
THS-on: Yeah sounds good to me. I increased it to 90 seconds for now.

shim.c
THS-on: I now updated the error message handling to differentiate between likely .sbat section or internal errors and when an actual revocation happens.

As I had most of the code for showing the component name and expected and found SBAT levels already done, I also now included it in the error message shown to the user.

Error message for revocation looks like this:
revocation

Error message for everything else (in this case I just set the .sbat section to nothing):
error

I think this gives users and also distros enough info if something happens.

jsetje: On the one hand I don't love "Something has gone seriously wrong", on the other hand it does convey that we did not want anyone to end up in that situation and it's worth escalating so that we can fix anything that may have automatically put it there, such as a dependency check before applying a revocation. While it made headlines recently, I didn't actually see complaints about that string, just that there wasn't additional useful information.

THS-on: There is two types of information that we want to convey: what went wrong (e.g. shim was revoked via SBAT) and what to do for the user.

The first one should contain enough information for us or the distributions to see what is wrong and the second something that a user can easily understand. 

We already print a decent amount of technical error messages, that I wouldn't want to hide, because those are helpful to debug issues. Maybe something like this:

```
[...]
Something has gone seriously wrong: [...]
SBAT self-check failed! Expected "shim,5" or higher. Got :"shim,4".

Your shim has been revoked via the UEFI Secure Boot SBAT policy on your machine. Please update to a compliant version.

[Vendor custom string here]
```

We can add the option for distributions to include some information where to redirect people, this is then something we would need to audit during reviews and have policies around. @jsetje what do you think? Also wouldn't direct users directly to upstream as most errors should go through the distribution first.

THS-on:
@steve-mcintyre @jsetje friendly ping on how we want to continue with this PR.

shim.c
steve-mcintyre: We can add a vendor string option at build time quite readily here - is that all we need to do to move this forward? I'm happy with this last suggesrtion from @aronowski here.

THS-on:
@vathpela @jsetje do we want to try to get this still in 16.0?

vathpela:
It didn't look ready when I was triaging last month, so I was assuming we'd target 17 for this.

THS-on:
> It didn't look ready when I was triaging last month, so I was assuming we'd target 17 for this.

Yeah still needs some agreement on how/if we allow a custom user facing string to be compiled in. Then let's target 17 with this.

Issue #634, opened 2024-02-02
SBAT.md recommends resetting vendor generation numbers after global generation number went up


kukrimate:
For example:
> However, once the global number is bumped for the next upstream CVE fix there will be no further need to carry that product-specific generation number. Satisfying the check of the global number will also exclude any of the older product-specific binaries.

> If this same Vendor C has a similar event after the global number is incremented, they would again set their product-specific or version-specific number to 1. If they have a second event on the same component, they would set their product-specific or version-specific number to 2.

The current consensus seems to be for both generation numbers to only ever go up, so we should stop recommending people do this.

PR #672, opened 2024-07-23
Fix SBAT.md for today's consensus about numbers

aronowski:
As per https://github.com/rhboot/shim/issues/634, the current consensus seems to be for generation numbers to only ever go up and not get reset. This has been clarified and an example related to this behavior has been described.

### DIFF
--- a/SBAT.md
+++ b/SBAT.md
@@ -214,8 +214,7 @@ product. Setting a product-specific generation number for such an event
 eliminates the need for other vendors to have to re-release the binaries for
 their products with an incremented global number.
 
-The current consensus is for both generation numbers to only ever go up and not
-be reset.
+Both generation numbers should only ever go up; they should never be reset.
 
 #### Example: a vendor forking a global project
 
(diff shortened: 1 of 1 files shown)


SBAT.md
```diff
@@ -213,68 +214,95 @@ product. Setting a product-specific generation number for such an event
 eliminates the need for other vendors to have to re-release the binaries for
 their products with an incremented global number.
 
-However, once the global number is bumped for the next upstream CVE fix there
-will be no further need to carry that product-specific generation number.
-Satisfying the check of the global number will also exclude any of the older
-product-specific binaries.
+The current consensus is for both generation numbers to only ever go up and not
```
steve-mcintyre, line 217: No need to say "current consensus" here IMHO. How about simply:

Both generation numbers should only ever go up; they should never be reset.

SBAT.md
```diff
@@ -213,68 +214,95 @@ product. Setting a product-specific generation number for such an event
...
+```
+sbat,1,SBAT Version,sbat,1,https://github.com/rhboot/shim/blob/main/SBAT.md
+grub,4,Free Software Foundation,[...]
+grub.vendorc,1,Vendor C,[...]
+```
+
+In the perfect scenario, to provide the perfect security, the UEFI SBAT
+revocation variable (named *SbatLevel*) would then be set, so that GRUB with a
+global generation number of only 4 or higher would be able to be booted. (In
+reality there's naturally going to be a certain grace period)
```
steve-mcintyre, line 248: I don't think you need to emphasise the "perfect security" etc. so much here. Maybe just:

After this, the UEFI SBAT revocation variable (named *SbatLevel*) would be updated to raise the minimum global generation number for GRUB to 4.

?

SBAT.md
```diff
@@ -213,68 +214,95 @@ product. Setting a product-specific generation number for such an event
...
 
 However, Vendor C mis-merges the patches into one of their products and does
 not become aware of the fact that this mis-merge created an additional
-vulnerability until after they have published a signed binary in that,
-vulnerable, state.
+vulnerability until after they have published a signed binary in that vulnerable
+state. Vendor C's GRUB binary can now be used to compromise everyone's systems.
```
steve-mcintyre, line 253: Not sure we need "everyone's" here?

SBAT.md
```diff
@@ -213,68 +214,95 @@ product. Setting a product-specific generation number for such an event
...
+grub,5,Free Software Foundation,[...]
+grub.vendorc,2,Vendor C,[...]
+```
 
-In such an event, a vendor would set the product-specific or version-specific
-generation number based on whether the mis-merge occurred in all of their
-branches or in just a subset of them. The goal is generally to limit end
-customer impact with as few re-releases as possible, while not creating an
-unnecessarily large UEFI revocation variable payload.
+The goal is generally to limit end customer impact with as few
```
steve-mcintyre, line 284: s/end customer/end user/ maybe?

steve-mcintyre (commented):
A few suggested changes, otherwise looks good!

steve-mcintyre (approved):
Looks good, I think!

PR #686, opened 2024-08-20
netboot cleanup for additional files

jsetje:
These proposed changes are ready for comments, especially around the naming. Some XXX's in the commit comments.

### DIFF
--- a/shim.c
+++ b/shim.c
@@ -1464,7 +1464,10 @@ load_revocations_file(EFI_HANDLE image_handle, CHAR16 *PathName)
 	uint8_t *ssps_latest = NULL;
 	uint8_t *sspv_latest = NULL;
 
-	efi_status = read_image(image_handle, L"revocations.efi", &PathName,
+	efi_status = read_image(image_handle, L"revocations_sbat.efi", &PathName,
+				&data, &datasize,
+				SUPPRESS_NETBOOT_OPEN_FAILURE_NOISE);
+	efi_status = read_image(image_handle, L"revocations_skusi.efi", &PathName,
 				&data, &datasize,
 				SUPPRESS_NETBOOT_OPEN_FAILURE_NOISE);
 	if (EFI_ERROR(efi_status))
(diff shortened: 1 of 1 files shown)


shim.c
```diff
@@ -1464,7 +1464,10 @@ load_revocations_file(EFI_HANDLE image_handle, CHAR16 *PathName)
 	uint8_t *ssps_latest = NULL;
 	uint8_t *sspv_latest = NULL;
 
-	efi_status = read_image(image_handle, L"revocations.efi", &PathName,
+	efi_status = read_image(image_handle, L"revocations_sbat.efi", &PathName,
+				&data, &datasize,
```
hughsie, line 1469: are you missing a check for efi_status here?

shim.c
```diff
@@ -1693,13 +1696,18 @@ load_unbundled_trust(EFI_HANDLE image_handle)
 			if (EFI_ERROR(efi_status)) {
 				perror(L"Failed to open %s - %r\n",
 				       PathName, efi_status);
+				while (load_cert_file(image_handle, FileName, PathName,
+						SUPPRESS_NETBOOT_OPEN_FAILURE_NOISE)
+						== EFI_SUCCESS && i++ < 10) {
+					FileName[17]++;
```
hughsie, line 1702: I think it's going to be confusing having to rename shim_certificate_sbat.efi to shim_certificate_0.efi, shim_certificate_ski.efi to shim_certificate_1.efi etc. Can't we just use those names?

jsetje: For the netboot case we can't do a reddir. So fining something like shim_certificate_<vendor_name>.efi would require an awful lot permutations to try, and I really don't think that's the right thing to do.

So I'm putting it on whoever deploys the netboot config to deal with renaming or linking. It's not wonderful, but does seem better to have this flexibility in place.

jsetje:
I should re-test some of the failure variations. I'll make a note here when that's done.

rosslagerwall (approved):

Issue #457, opened 2022-03-02
shim-signed not working with Macbook Pro 2020


kami83:
Hi, i am trying to run a Macbook Pro (3 Distros, Windows, MacOS, Linux) with refind and i would like to activate shim with Secure Boot. Can anyone tell me why my Macbook always goes to Recovery Boot, when i try to launch shimx64.efi. I tried to use an Ubuntu version with Microsoft-signed. Is here anyone how can help me with this problem?

Thanks  a lot.

BG kami

afrodidact:
Did you find a resolution for this issue?

kami83:
No sorry. Nothing new. It doesnt work.

BR kami

frozencemetery:
Apple's UEFI is also more likely to be divergent than that of other vendors, and I don't believe any shim developers currently own Apple hardware.  Additionally, they don't anchor their root of trust with Microsoft (because they're Apple and sell an OS), so whether you're signed or unsigned I don't think would matter here.

While I understand booting multiple operating systems on the same machine is attractive, using a tool like refind as an additional boot manager introduces a fragile layer of complexity.  When I used to do this on Apple hardware, I was unable to get refind to ever work (I'm not convinced it does, and this is not a place to debug it).  I suggest you install one operating system per drive and either physically swap drives or use Apple's boot manager to switch OSs (hold I think it's the Option key during boot) - that's how I did it (in 2015, on a MacPro4,1).  Having them on the same drive means they'll quite likely step on each other later, be that partition tables or ESP shenanigans.

Finally, you are using a distro's shim, so they may be more helpful than upstream.  Concretely: if you have a generally reasonable setup, I suggest you file a bug with Ubuntu and see if they can help you more.  They'll probably want more information to start with, including how you installed and stuff.

Best of luck to you.

Issue #385, opened 2021-06-23
shim 15.4 is causing problems on new Intel-based Macbooks 


mikeeq:
Hi all,

I was trying to boot latest release of Fedora 34 on my 2020 13" Macbook Pro, but when I choose to boot the livecd in builtin Mac's bootloader is not showing anything else besides black screen. Downgrading shim version to 15-8 resolves the issue. I've checked all 15.4-* packages of shim, but none of them boot correctly to grub2.

https://koji.fedoraproject.org/koji/packageinfo?packageID=14502

I'm not sure how should I debug it, so please share some information what exactly do I need to do/grab/share etc. Thanks ;)

julian-klode:
Presumably one of the known issues fixed in git. Fedora needs a new shim at some point, and we really need a new shim release.

You'll have to use `mokutil --set-verbosity true` and then capture the screen with a camera. Good luck.

thesourcehim:
I'm having similar issue on Dell 5510 laptop with shim 15.4 (Fedora 34). After I enrolled my mok, shim started to hang on boot unless secure boot is set to audit mode in bios. I.e audit mode - shim boots, deploy mode (default) - shim hangs, secure boot disabled - shim hangs

thesourcehim:
Here's shim's log with secure boot set to default mode
![IMG_20210727_144129](https://user-images.githubusercontent.com/4952491/127148339-088e7e9a-2a7e-4a9d-adc0-1e6d58b5bb12.jpg)

I can upload previous log segments if needed. After the last message (finished verifying SBAT data: success) nothing happens, this screen stays forever.

ValdikSS:
Have a black screen with infinite loop (laptop runs hot) on Lenovo ThinkPad X220, and old laptop without Secure Boot support, after updating from Fedora 33 to 34.
No prints, no anything, just black screen.
Booting grubx64 using EFI shell works fine.

Running `mokutil --set-verbosity true` says `This system doesn't support Secure Boot`.

thesourcehim:
I want to try latest git version, but I'm stuck: I have built shim and signed it according to this: https://en.opensuse.org/openSUSE:UEFI_Image_File_Sign_Tools
But bios fails to verify the signature. What am I doing wrong?

sudo pesign -n certdb -S -i /boot/efi/EFI/fedora/shimx64.efi 

certificate address is 0x7f6d61739a48
Content was not encrypted.
Content is detached; signature cannot be verified.
The signer's common name is thesource
[...]

PR #364, opened 2021-04-12
Don't call QueryVariableInfo() on EFI 1.10 machines

vathpela:
The EFI 1.10 spec (and presumably earlier revisions as well) didn't have
RT->QueryVariableInfo(), and on Chris Murphy's MacBookPro8,2 , that
memory appears to be initialized randomly.

This patch changes it to not call RT->QueryVariableInfo() if the
EFI_RUNTIME_SERVICES table's major revision is less than two, and
assumes our maximum variable size is 1024 in that case.

Signed-off-by: Peter Jones <pjones@redhat.com>

PR #690, opened 2024-08-29
avoid EFIv2 runtime services on Apple x86 machines

eduardacatrinei:
While booting a MacBookPro15,2 (the last Intel model, 2019). 
The shim 15.8 got stuck in `RT->QueryVariableInfo()`. 
Previously, these devices shipped with EFI firmware version 1.10, and we had a quirk in place for this (https://github.com/rhboot/shim/pull/364)
However, Apple updated the firmware to version 2.40, but it still doesn't implement runtime services.

Logs and a screen recording have been attached to this comment.
https://github.com/rhboot/shim/pull/364#issuecomment-2312795519

Related to:
https://github.com/rhboot/shim/issues/385
https://github.com/rhboot/shim/issues/457
https://nvd.nist.gov/vuln/detail/CVE-2022-48769


eduardacatrinei:
@vathpela can you take a look here?

AdityaGarg8:
Does this make Apple's T2 chip recognise the signature? Or it just makes it bootable with secure boot turned off?

eduardacatrinei:
This will help prevent a black screen when trying a distro based on shim.
According with [Apple_T2_Security_Chip_Overview](https://www.apple.com/jp/mac/docs/Apple_T2_Security_Chip_Overview.pdf), it will not recognize the signature.
```
NOTE: There is currently no trust provided for the the Microsoft Corporation
UEFI CA 2011, which would allow verification of code signed by Microsoft
partners. This UEFI CA is commonly used to verify the authenticity of
bootloaders for other operating systems such as Linux variants.
```

vathpela:
I've made a small change to not trust that `ST->FirmwareVendor` is a properly formatted string, but other than that it looks good to me.

vathpela (approved):

PR #695, opened 2024-10-03
Handle network file not found like disks

dbnicholson:
When the second stage can't be found, shim will try the default second stage. That only work when reading the image from disk since the network protocols don't return `EFI_NOT_FOUND`. Convert the appropriate network error codes to `EFI_NOT_FOUND` so they're treated the same way.

See #649.

### DIFF
--- a/httpboot.c
+++ b/httpboot.c
@@ -55,6 +55,51 @@ convert_http_status_code (EFI_HTTP_STATUS_CODE status_code)
 	return 0;
 }
 
+/* Convert an HTTP status code to an EFI status code. */
+static EFI_STATUS
+efi_status_from_http_status(EFI_HTTP_STATUS_CODE status_code)
+{
+	switch (status_code) {
+	case HTTP_STATUS_400_BAD_REQUEST:
+	case HTTP_STATUS_411_LENGTH_REQUIRED:
+	case HTTP_STATUS_413_REQUEST_ENTITY_TOO_LARGE:
+	case HTTP_STATUS_414_REQUEST_URI_TOO_LARGE:
+	case HTTP_STATUS_415_UNSUPPORTED_MEDIA_TYPE:
+	case HTTP_STATUS_416_REQUESTED_RANGE_NOT_SATISFIED:
+	case HTTP_STATUS_417_EXPECTATION_FAILED:
+		return EFI_INVALID_PARAMETER;
+	case HTTP_STATUS_401_UNAUTHORIZED:
+	case HTTP_STATUS_402_PAYMENT_REQUIRED:
+	case HTTP_STATUS_403_FORBIDDEN:
+	case HTTP_STATUS_407_PROXY_AUTHENTICATION_REQUIRED:
+		return EFI_ACCESS_DENIED;
+	case HTTP_STATUS_404_NOT_FOUND:
+	case HTTP_STATUS_410_GONE:
+		return EFI_NOT_FOUND;
+	case HTTP_STATUS_405_METHOD_NOT_ALLOWED:
+	case HTTP_STATUS_501_NOT_IMPLEMENTED:
+		return EFI_UNSUPPORTED;
+	case HTTP_STATUS_406_NOT_ACCEPTABLE:
+		return EFI_NO_MEDIA;
+	case HTTP_STATUS_408_REQUEST_TIME_OUT:
+	case HTTP_STATUS_504_GATEWAY_TIME_OUT:
+		return EFI_TIMEOUT;
+	case HTTP_STATUS_409_CONFLICT:
+	case HTTP_STATUS_412_PRECONDITION_FAILED:
+		return EFI_MEDIA_CHANGED;
+	case HTTP_STATUS_500_INTERNAL_SERVER_ERROR:
+	case HTTP_STATUS_502_BAD_GATEWAY:
+		return EFI_DEVICE_ERROR;
+	case HTTP_STATUS_503_SERVICE_UNAVAILABLE:
+		return EFI_NOT_READY;
+	case HTTP_STATUS_505_HTTP_VERSION_NOT_SUPPORTED:
+		return EFI_INCOMPATIBLE_VERSION;
+	default:
+		/* Use a generic HTTP error for anything else. */
+		return EFI_HTTP_ERROR;
+	}
+}
+
 static EFI_DEVICE_PATH *devpath;
 static EFI_MAC_ADDRESS mac_addr;
 static IPv4_DEVICE_PATH ip4_node;
@@ -565,7 +610,7 @@ receive_http_response(EFI_HTTP_PROTOCOL *http, VOID **buffer, UINT64 *buf_size)
 	if (http_status != HTTP_STATUS_200_OK) {
 		perror(L"HTTP Status Code: %d\n",
 		       convert_http_status_code(http_status));
-		efi_status = EFI_ABORTED;
+		efi_status = efi_status_from_http_status(http_status);
 		goto error;
 	}
 


dbnicholson:
I haven't tested this, but it looks appropriate to me.

mikebeaton:
@dbnicholson - Is this right? Just a flyby comment, but I would guess that most (all?) failures when connecting to TFTP or HTTP should be equivalent to Not Found for the purposes of failing over to the default second stage?

nathan-omeara:
This doesn't seem to work.  I tested in my lab, and shim does not try the fallback default loader.  I pulled up a wireshark capture to make sure the TFTP response set the error code right, and it appears right.

image

dbnicholson:
@mikebeaton Personally I would want to keep it narrow. Shim has a case to handle file not found and TFTP has an error for file not found. More specifically, when loading images from disk, shim only does the fallback for `EFI_INVALID_PARAMETER` and `EFI_NOT_FOUND`. For reference, there are lots of other errors that could be encountered:

- [OpenVolume](https://uefi.org/specs/UEFI/2.10/13_Protocols_Media_Access.html#efi-simple-file-system-protocol-openvolume)
- [Open](https://uefi.org/specs/UEFI/2.10/13_Protocols_Media_Access.html#efi-file-protocol-open)
- [GetInfo](https://uefi.org/specs/UEFI/2.10/13_Protocols_Media_Access.html#efi-file-protocol-getinfo)
- [Read](https://uefi.org/specs/UEFI/2.10/13_Protocols_Media_Access.html#efi-file-protocol-read)

Shim doesn't try to handle `EFI_NO_MEDIA` or `EFI_DEVICE_ERROR`, for example. You may notice that none of those functions return `EFI_INVALID_PARAMETER`. That only happens when shim sets the error that way. So, when loading images off disk, you really only get the fallback to the default loader when the volume or file can't be found. Not for anything else, and I don't think network booting should be different unless the disk rules change, too.

If I was to revise my PR, it would be to map more error codes to simple file equivalents. In other words, make the interface for reading an image from network the same as reading from disk. If that was the case, you could discuss expanding the cases where shim automatically tries the default loader instead of failing.

dbnicholson:
> This doesn't seem to work. I tested in my lab, and shim does not try the fallback default loader. I pulled up a wireshark capture to make sure the TFTP response set the error code right, and it appears right.

Thanks for trying that. Maybe I'll have to spin up a TFTP server to see what the error looks like from the firmware side.

nathan-omeara:
I added some debug prints.. TftpErrorRecevied is 0, TftpError.ErrorCode is 0.

the error isn't getting passed back correctly, I guess?

netboot.c
```diff
@@ -362,8 +372,24 @@ EFI_STATUS FetchNetbootimage(EFI_HANDLE image_handle UNUSED, VOID **buffer, UINT
...
-	if (EFI_ERROR(efi_status) && *buffer) {
-		FreePool(*buffer);
+	if (EFI_ERROR(efi_status)) {
+		if (pxe->Mode->TftpErrorReceived) {
+			console_print(L"TFTP error %u: %a\n",
+				      pxe->Mode->TftpError.ErrorCode,
+				      pxe->Mode->TftpError.ErrorString);
+
+			switch (pxe->Mode->TftpError.ErrorCode) {
+			case TFTP_ERROR_NOT_FOUND:
```
nathan-omeara, line 382: FWIW, I think `TFTP_ERROR_ACCESS` would also make sense to fall back on.. honestly, almost any error from the server side should probably trigger a retry with the default filename.

Same on the HTTP side, 'not found/404' is the obvious one, but 403s, 500s, other things could also be dependent on the filename requested, so it would make sense to retry with the default loader name with almost any error returned from the server.

On the other hand, network layer errors don't make sense to retry with a different filename - but I'm not sure how to differentiate those here.

mikebeaton:
> @mikebeaton Personally I would want to keep it narrow. Shim has a case to handle file not found and TFTP has an error for file not found.

Hi again - Just to be clear what I was thinking of, I was especially considering HTTP, where 404 Not Found is very specific - it means everything regarding connecting to the site was successful, but that the specific named document is not on the site. For instance, the site itself might not be found, and that is not a not found error. So I guess at least the site itself not being found is quite similar to the volume not being found? That was the thinking behind the suggestion. I'm not sure if that realistically adds any errors to your list, since I don't think you'd even get an HTTP error in that case.

dbnicholson:
> I added some debug prints.. TftpErrorReceived is 0, TftpError.ErrorCode is 0.
> 
> the error isn't getting passed back correctly, I guess?

That's too bad. I think the firmware is responsible for filling that in, but maybe you have to tell it to first. I'll look closer but I suppose you could just map `EFI_TFTP_ERROR` to `EFI_NOT_FOUND` right there.

dbnicholson:
> Hi again - Just to be clear what I was thinking of, I was especially considering HTTP, where 404 Not Found is very specific - it means everything regarding connecting to the site was successful, but that the specific named document is not on the site. For instance, the site itself might not be found, and that is not a not found error. So I guess at least the site itself not being found is quite similar to the volume not being found? That was the thinking behind the suggestion. I'm not sure if that realistically adds any errors to your list, since I don't think you'd even get an HTTP error in that case.

Yeah, that's exactly what I was saying about the disk case. `EFI_NOT_FOUND` is only returned by `Open()` when you already opened the volume. All the other errors you'd encounter trying to load an image from disk will skip the fallback to the default loader. An HTTP 403, for example, doesn't mean the file wasn't found. It means that you're not authorized to access it. In the disk case you'd get `EFI_ACCESS_DENIED`. I think that if you couldn't access the site at all (DNS resolution failure, no route, etc), I think you'd end up with `EFI_DEVICE_ERROR` in the HTTP request. That would be very similar to the `EFI_NO_MEDIA` case for `OpenVolume()`. That still wouldn't get you to the fallback case, though.

dbnicholson:
> I added some debug prints.. TftpErrorReceived is 0, TftpError.ErrorCode is 0.
> 
> the error isn't getting passed back correctly, I guess?

I also got this with OVMF and it was a simple byte swapping issue. See https://github.com/tianocore/edk2/pull/6287. I imagine that most UEFI firmwares use this TFTP code rather than roll their own. It's largely unchanged from when Intel contributed it 14 years ago.

With that change, shim in verbose mode reports:

```
Fetching Netboot Image /grubx64.efi
TFTP error 1: File not found
Unable to fetch TFTP image: Not Found
start_image() returned Not Found, falling back to default loader
```

Since this is likely to affect a lot of people, I guess I'll just include a fallback to map `EFI_TFTP_ERROR` to `EFI_NOT_FOUND`.

nathan-omeara:
tested this branch on my latitude 5300 again - this updated branch does correctly fallback to default loader.

netboot.c
dbnicholson: There are really 2 issues here:

1. The shim frontend doesn't handle the errors raised by the network backends. That's what I'm trying to address here. I've chosen to translate the network backend errors to the file backend errors (although I only added file not found here). That way the shim frontend doesn't need to special case the errors returned from the network backends.
2. Shim should fallback to the default loader in more cases. I agree shim should try the default loader for an access denied error. While it's likely reading the default loader will also fail, trying is better than failing outright.

I'm going to rework the patches a bit more so more errors are mapped and then I'll add a commit that also tries the default loader on access denied.

dbnicholson: I updated the patches to map all TFTP and HTTP error codes to EFI status codes, but I decided not to change the types of errors that shim falls back to the default loader on. The policy that's there is from 50732ee91 and it's to cover the case where shim has incorrectly decided what the second stage loader filename is. If you want the policy to be more like "always try the default second stage if loading the specified one fails", then that's really a separate thing. FWIW, I wrote a patch to add `EFI_ACCESS_DENIED` [here](https://github.com/dbnicholson/shim/commit/dab2c2aa92fb18e27eb35d03c2eb4760970da6cc).

vathpela (approved):

PR #697, opened 2024-10-14
Add support for sm2 and sm3 algorithms into shim

costinchen:
Add EC and ECDSA components to bundled Openssl in the shim and add support for SM2 and SM3 algorithms. 
SM2 is a prominent elliptic curve cryptography standard and offers enhanced security and efficiency. SM3 provides a robust hashing mechanism with strong collision resistance. 
You can add the ENABLE_SHIM_HASH=true compilation option to enable it, like `make ENABLE_SHIM_HASH=true ...`


vathpela:
What's the use case for this?  I don't know of any firmware or UEFI signing authority that uses these algorithms for Secure Boot.

In any case, we're not taking patches to add functionality to the OpenSSL tree - we would consider a rebase to newer versions, but adding hashes and ciphers in this tree is right out.  So I'm going to close this, but if there's a good rationale to add the functionality, I'm open to a patch series that adds it /after/ moving to a new OpenSSL version that includes it.

julian-klode:
Support for SM algorithms is being mandated by Chinese equivalent to cyber security regulations like FIPS in the US.

During the review of a shim submission, we asked them to submit the patches to upstream shim instead of keeping them locally.

The algorithms are in newer OpenSSL versions, upgrading to them would be better but then I think reviewing the update to a new OpenSSL version would be even harder than reviewing just the backported hashes - in both cases we need to compare against upstream in some way. Just a thought.

costinchen:
As Julian mentioned, supporting the SMx algorithms is now one of the requirements of China's cyber security standards for operating systems, including using these algorithms for Secure Boot. 

Anyway, the SMx algorithms are relatively new and more secure cryptographic algorithms. So we hope you will consider supporting these algorithms in Shim after upgrading the OpenSSL version.

PR #700, opened 2024-10-28
shim: immediately stop if signature is matched.

dennis-tseng99:
If an image signed by multiple keys matches any one of vendor certificates, 
the image will be considered valid and should stop checking the next signature. 
This PR tries to immediately stop do-while() next signature checking in verify_buffer_authenticode() once key is matched.

### DIFF
--- a/shim.c
+++ b/shim.c
@@ -563,7 +563,7 @@ verify_buffer_authenticode (char *data, int datasize,
 			    PE_COFF_LOADER_IMAGE_CONTEXT *context,
 			    UINT8 *sha256hash, UINT8 *sha1hash)
 {
-	EFI_STATUS ret_efi_status;
+	EFI_STATUS efi_status;
 	size_t size = datasize;
 	size_t offset = 0;
 	unsigned int i = 0;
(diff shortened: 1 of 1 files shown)


mikebeaton:
@dennis-tseng99 - I think this change may not be correct:

https://github.com/rhboot/shim/commit/76c0447e

"If at any point any assertion about the
binary or signature list being well-formed fails, the binary is
immediately rejected, though we do allow skipping over signatures
which have an unsupported sig->Hdr.wCertificateType."

dennis-tseng99:
@mikebeaton, Thanks for your comment. Hope not to misunderstand your meaning.

```
If at any point any assertion about the
binary or signature list being well-formed fails, the binary is
immediately rejected
```

Actually, in this PR, do-while would not stop until verification gets SUCCESS or offset is too large. Even the well-formed fails happen, the binary would not be immediately rejected.

mikebeaton:
> @mikebeaton, Thanks for your comment. Hope not to misunderstand your meaning.
> 
> ```
> If at any point any assertion about the
> binary or signature list being well-formed fails, the binary is
> immediately rejected
> ```
> 
> Actually, in this PR, do-while would not stop until verification gets SUCCESS or offset is too large. Even the well-formed fails happen, the binary would not be immediately rejected.

In the PR, verification would stop if it gets SUCCESS. Before the PR verification would fail _after_ a success, if the binary was found to be malformed after the success.

mikebeaton:
Hi! I still think this is unfortunately probably wrong. If modifying code like this then (I humbly suggest!) it is important to check whether the code being skipped has any side effects that might be wanted. In this case, the calls to `verify_one_signature` which you are skipping make calls to `tpm_measure_variable` - so I believe the answer is yes, there are non-trivial, and wanted, side effects to 'pointlessly' verifying all signatures.

vathpela:
This doesn't look like it's ready for review, so I'm switching it to "draft" status.

PR #711, opened 2024-12-23
Generate SbatLevel Metadata from SbatLevel_Variable.txt

jsetje:
This makes SbatLevel_Variable.txt the single source of truth for SbatLevel metadata and generates the header file from it.


jsetje:
I failed to notice that the initial PR had broken out of place build. Fixed now.

PR #717, opened 2025-01-21
shim: ensure vendor_db gets logged into `db`

Foxboron:
README.tpm claims that `vendor_db` is logged into `db`, however the actual measurement was done towards `vendor_db` with the `check_db_hash_in_ram` function.

This causes some unexpected issues like in https://github.com/keylime/keylime/issues/1725


jsetje:
The alternative is to update the docs, which might result in fewer other things breaking:

https://github.com/rhboot/shim/pull/728

Thanks for raising this!

Foxboron:
I don't really have a particular preference to what we should do. The patch was submitted after some raising the issue on IRC.

vathpela:
For right now we're going with #728.  Thanks for bringing this up!

PR #660, opened 2024-05-15
Add features for the Host Security ID program

vathpela:

PR #723, opened 2025-02-18
Mirror some more efi variables to mok-variables

vathpela:
Some machines have EFI Boot Services variables but not Runtime variables, and thus it can be quite difficult to figure out what's going on once the system is booted.

This changes mok variable mirroring to also mirror the following variables to the mok variable config table:

      AuditMode
      BootOrder
      BootCurrent
      BootNext
      Boot0000
      Boot0001
      Boot0002
      Boot0003
      Boot0004
      Boot0005
      Boot0006
      DeployedMode
      SecureBoot
      SetupMode
      SignatureSupport
      PK
      KEK
      db
      dbx
      Kernel_SkuSiStatus

There's no attempt to do anything involving creating runtime or boot-services only variables, it just mirrors them into the config table so they'll be exposed there.


vathpela:
Note that this is now on top of https://github.com/rhboot/shim/pull/660

jsetje (approved):
I wonder if we want some way to disable the mirroring entirely. Something like MOK_VARIABLE_NOTHING ? I certainly don't want this by default, but some light memory only configs might want to be able to save the space.

PR #724, opened 2025-02-18
This adds DXE Services measurements to HSI and uses them for NX

vathpela:
(The NX work isn't here yet beyond exposing the API)


vathpela:
Note that this is now on top of https://github.com/rhboot/shim/pull/723

vathpela:
This all works, but currently there's nothing to make the NX permission setting *use* the DXE services, and also nothing to update the HSI bits using the DXE code, and I think we need that before I make this not a draft.

jsetje (approved):
I might pull the idef'd out stuff into a separate patch and leave that in a draft PR.

If you prefer to leave it in as a template for what should be there if these systems worked, I'm OK with that as well.

PR #742, opened 2025-03-26
Fix uncompressed ipv6 netboot

hrvach:
In case the ipv6 is not in a compressed format, the last chunk gets set to 0 because the str2ns() never executes.

As an example, without this fix provided, str2ip6 works correctly on a compressed address like 

2400:dd00::1 -> 24 00 dd 00 00 00 00 00 00 00 00 00 00 00 00 01

but fails on

2400:dd00:1234:1234:1234:1234:1234:1234 -> 24 00 dd 00 12 34 12 34 12 34 12 34 12 34  00 00

The fixed routine returns 24 00 dd 00 12 34 12 34 12 34 12 34 12 34 12 34 as expected.

### DIFF
--- a/netboot.c
+++ b/netboot.c
@@ -157,15 +157,20 @@ static CHAR8 *str2ip6(CHAR8 *str)
 		if (b[1] == ':' )
 			break;
 	}
-	a = b = (str + len);
-	for (j = len, p = 7; j > i; j--, a--) {
-		if (*a != ':')
-			continue;
-		t = *b;
-		*b = '\0';
-		ip[p--] = str2ns(a+1);
-		*b = t;
-		b = a;
+
+	if (i == len) {
+		ip[p++] = str2ns(a);
+	} else {
+		a = b = (str + len);
+		for (j = len, p = 7; j > i; j--, a--) {
+			if (*a != ':')
+				continue;
+			t = *b;
+			*b = '\0';
+			ip[p--] = str2ns(a+1);
+			*b = t;
+			b = a;
+		}
 	}
 	return (CHAR8 *)ip;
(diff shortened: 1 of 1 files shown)


vathpela:
I re-wrote the commit message with more of the info from the PR and fixed some formatting, but otherwise looks good to me, thanks!

hrvach:
Much appreciated, thanks for the review!

dennis-tseng99:
str2ip6() doesn't take care of multiple double-colon or illegal ipv6 digit, like:
**2400::dd00::1 or
2400:xyz0::1**

Maybe we could add something like this to deal with it:
```
static CHAR8 *str2ip6(CHAR8 *str)
{
        bool double_colon_found = false;
        .............................
        if (dotcount > MAX_IP6_DOTS)
                return (CHAR8 *)ip;

        len = strlen(str);
        a = b = str;

        for (i = 0; i < len; i++) {
                if (str[i] == ':') {
                        if (str[i+1] == ':') {
                                if (double_colon_found) {
				        printf("too many double-colon\n");
				        memset(ip, 0, sizeof(ip));
				        return (CHAR8 *)ip;
                                }
			        double_colon_found = true;
			        printf("double-colon 1st found!\n");
                        }
                }
                else {
                        if (str[i] < '0' || (str[i] > '9' && str[i] < 'A') || 
                           (str[i] > 'F' && str[i] < 'a') || str[i] > 'f') {
                        	memset(ip, 0, sizeof(ip));
                        	printf("unknown character %c\n",str[i]);
                        	return (CHAR8 *)ip;
                	}
                }
        }
        ................................
```

Issue #741, opened 2025-03-25
Untrusted components contained in a Trusted/Signed UKI fail to load with shim 16


jsetje:
Due to calling back into the loader protocol every object that's loaded ends up going through a verification. A UKI that's signed with a trusted key is loaded, but any object it contains, for example a kernel, is then validated again individually. Strictly speaking this verification is not required and will in fact result in measuring the kernel twice.

This is an issue for anyone that's building a UKI and then enrolling it in a trust differs from the contained components. Whether that's a kernel signed with a key that's not trusted in the context that the UKI is loaded in or simply an unsigned kernel.

Even just measuring the kernel twice will mean measurements for those components will change just by updating to shim 16, which is not desirable either.

Please consider if this impacts you before getting a shim 16 signed.

@ardbiesheuvel suggested we fix this by adding hashes of validated sections of a UKI to the db and then check objects that are subsequently loaded against that.

@kukrimate has prototyped a fix here: https://github.com/rhboot/shim/pull/740

kraxel:
I think @ardbiesheuvel idea is to add the authenticode hash to MokListRT instead of maintaining a separate list (as implemented by https://github.com/rhboot/shim/pull/740).  Not fully sure what side effects that would have on measurements though.

ardbiesheuvel:
> I think [@ardbiesheuvel](https://github.com/ardbiesheuvel) idea is to add the authenticode hash to MokListRT instead of maintaining a separate list (as implemented by [#740](https://github.com/rhboot/shim/pull/740)). Not fully sure what side effects that would have on measurements though.

That is what I suggested on the thread, and actually, what Jan is quoting here is more @vathpela's suggestion rather than mine.

What I suggested before is adding a PE section to an image carrying an ephemeral extension of `"db"`, which would be used in subsequent authentications. This proposal is basically a simplification of that approach, implemented in a way that preserves compatibility when upgrading to shim 16

bluca:
In case the implementation adds the hash to MokListRT, is the content of MokListRT measured in a PCR?

kukrimate:
I don't think adding it to any existing allowlist without special flagging is enough, if compatibility is the goal.

To not break comaptibility we need to avoid any extra measurements.

jsetje:
Is the extra measurement actually wrong though?

- the system loaded some UKI.
- the system loaded and booted a specific kernel in that UKI.

Not that I love breaking compatibility.

bluca:
[...]

PR #748, opened 2025-04-29
Loader proto v2

vathpela:
This patch series allows for transparent loading of sub-images from a section of a previously loaded image for UKIs.

It does this by adding a cache of section metadata and a sha256 digest when `LoadImage()` is called and verification is successful.  On subsequent `LoadImage()` calls, if `SourceBuffer` and `SourceSize` are provided and the size matches a section in the parent image's cache, the sha256 sum is calculated for comparison.  If it matches the cached entry, the new image is understood to have been verified during the caller's verification.  In this case, `dbx`/`MokListX` are still checked for specific revocations of the loaded image, and the `SBAT` check is performed, but `dbx`/`MokListX` are not checked for revocations of signatures.

Resolves issue #741 .

### DIFF
--- a/loader-proto.c
+++ b/loader-proto.c
@@ -45,6 +45,23 @@ typedef struct {
 } buffer_properties_t;
 
 static EFI_STATUS
+try_load_from_cached_section(EFI_HANDLE parent_image_handle,
+			     EFI_DEVICE_PATH *dp, buffer_properties_t *bprop)
+{
+	EFI_STATUS status;
+
+	status = validate_cached_section(parent_image_handle,
+					 bprop->buffer, bprop->size);
+	if (EFI_ERROR(status))
+		return status;
+
+	bprop->allocated_buffer = false;
+	bprop->dp = dp;
+
+	return EFI_SUCCESS;
+}
+
+static EFI_STATUS
 try_load_from_sfs(EFI_DEVICE_PATH *dp, buffer_properties_t *bprop)
 {
 	EFI_STATUS status = EFI_SUCCESS;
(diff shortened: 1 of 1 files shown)


pe.c
```diff
@@ -39,6 +39,130 @@
...
+	CopyMem(entry->section_name, section_name, sizeof(entry->section_name));
+	entry->addr = addr;
+	entry->size = size;
+
+	efi_status = _do_sha256_sum(addr, size, entry->digest);
+	if (EFI_ERROR(efi_status)) {
+		ZeroMem(entry, sizeof (*entry));
+		return efi_status;
+	}
+	return EFI_SUCCESS;
```
bluca, line 107: `num_section_cache_entries` needs to be incremented before returning, otherwise the validate/flush helpers below will not be able to iterate over the cache

pe.c
```diff
@@ -772,8 +899,18 @@ handle_image (void *data, unsigned int datasize,
 			clear_attrs &= ~MEM_ATTR_X;
 		}
 		update_mem_attrs(addr, length, set_attrs, clear_attrs);
-	}
 
+		if ((Section->Characteristics & EFI_IMAGE_SCN_CNT_CODE ||
+		     Section->Characteristics & EFI_IMAGE_SCN_CNT_INITIALIZED_DATA) &&
+		    Section->Characteristics & EFI_IMAGE_SCN_MEM_READ) {
+			efi_status = cache_section(li, Section->Name, base, raw_length);
```
bluca, line 911: The first parameter needs to be `li->ParentHandle` rather than `li`, as that's where LoadImage will save the handle passed by the caller, `li` is a handle local to shim

pe.c
```diff
@@ -772,8 +899,18 @@ handle_image (void *data, unsigned int datasize,
...
+			if (EFI_ERROR(efi_status)) {
+				perror(L"Failed to cache section details\n");
+				BS->FreePages(*alloc_address, *alloc_pages);
+				return efi_status;
+			}
```
bluca, line 916: A debug print here noting a successful caching would be nice - it did help me debug these small issues, something like:

```
dprint(L"Cached section %d (%a) at 0x%016llx, size 0x%016llx\n",
       i, Section->Name,
       (unsigned long long)(uintptr_t)base,
       (unsigned long long)raw_length);
```

pe.c
bluca: scratch that, user error

bluca:
Tested with sd-boot and sd-stub changes at https://github.com/systemd/systemd/compare/main...bluca:systemd:shim_loader_proto and works nicely

pe.c
```diff
@@ -39,6 +39,130 @@
...
+{
+	struct shim_section_cache_entry *section = NULL;
+	UINT8 digest[32];
+	EFI_STATUS efi_status;
+
+	ZeroMem(digest, sizeof(digest));
+
+	for (UINTN i = 0; i < num_section_cache_entries; i++) {
+		struct shim_section_cache_entry *this_entry = &section_cache[i];
+
```
bluca, line 122: A debug print here helped me test things and would be nice to have, something like:

```
dprint(L"Handles: 0x%016llx 0x%016llx sections: '%a' '%a'\n",
	(unsigned long long)(uintptr_t)this_entry->parent_image_handle, (unsigned long long)(uintptr_t)parent_image_handle,
	this_entry->section_name, section_name);
```

pe.c
vathpela: Fixed in my current tree.

pe.c
vathpela: I've added this in my current tree.

pe.c
```diff
@@ -772,8 +905,22 @@ handle_image (void *data, unsigned int datasize,
 			clear_attrs &= ~MEM_ATTR_X;
 		}
 		update_mem_attrs(addr, length, set_attrs, clear_attrs);
-	}
 
+		if ((Section->Characteristics & EFI_IMAGE_SCN_CNT_CODE ||
+		     Section->Characteristics & EFI_IMAGE_SCN_CNT_INITIALIZED_DATA) &&
+		    Section->Characteristics & EFI_IMAGE_SCN_MEM_READ) {
+			efi_status = cache_section(li->ParentHandle, Section->Name, base, raw_length);
```
bluca, line 912: This was actually my mistake, I marked the earlier comment as solved, it was right the first time with `li`

pe.c
vathpela: I've added this in my current tree as well.

kukrimate:
Seems like there is some type confusion in these patches.

What is called a `parent_image_handle` in `cache_section(EFI_HANDLE parent_image_handle`, etc isn't actually an `EFI_HANDLE` but the loaded image protocol pointer.

For example `efi_status = cache_section(li, Section->Name, base, raw_length);` clearly passes a loaded image and not a handle.

Of course EFI_HANDLE is just a typedef for void * so that compiles without an error but not semantically correct at all.

Then this tries to do a lookup against a EFI_HANDLE, but what was saved is actually the protocol pointer.
```
static EFI_STATUS
try_load_from_cached_section(EFI_HANDLE parent_image_handle,
				  EFI_DEVICE_PATH *dp,
				  buffer_properties_t *bprop)
{
	EFI_STATUS status;

	status = validate_cached_section(parent_image_handle,
```

bluca (approved):
Patches look good to me, and I've tested it with an existing sd-boot + sd-stub UKI, everything works. Ship it!

vathpela:
> Seems like there is some type confusion in these patches.
> 
> What is called a `parent_image_handle` in `cache_section(EFI_HANDLE parent_image_handle`, etc isn't actually an `EFI_HANDLE` but the loaded image protocol pointer.
> 
> For example `efi_status = cache_section(li, Section->Name, base, raw_length);` clearly passes a loaded image and not a handle.

This is fixed in the current version.

loader-proto.c
```diff
@@ -241,21 +241,22 @@ shim_load_image(BOOLEAN BootPolicy, EFI_HANDLE ParentImageHandle,
...
+		goto free_image;
+
+	in_protocol = 1;
+	efi_status = handle_image(SourceBuffer, SourceSize, &image->li,
+				  *ImageHandle, &image->entry_point,
+				  &image->alloc_address, &image->alloc_pages,
+				  parent_verified);
+	in_protocol = 0;
 	if (EFI_ERROR(efi_status))
 		goto free_alloc;
```
bluca, line 301: By swapping the handle_image and the protocol installation, if handle_image fails (because of a security violation or other issues) the caller is still left with the image handle installed. The `free_alloc` section section below should uninstall and null the variable to avoid this:

```
free_alloc:
	BS->UninstallMultipleProtocolInterfaces(ImageHandle,
	                                &SHIM_LOADED_IMAGE_GUID, image,
	                                &EFI_LOADED_IMAGE_GUID, &image->li,
	                                &gEfiLoadedImageDevicePathProtocolGuid,
					image->loaded_image_device_path,
					NULL);
	*ImageHandle = NULL;
```

vathpela: Thanks for noticing this.  I'll have an updated patch soon.

PR #769, opened 2025-08-05
Fix double free and leak in the loader protocol

rosslagerwall:


loader-proto.c
```diff
@@ -414,6 +414,8 @@ shim_unload_image(EFI_HANDLE ImageHandle)
 
 	if (efi_status == EFI_UNSUPPORTED)
 		return system_unload_image(ImageHandle);
+	else if (efi_status != EFI_SUCCESS)
+		return efi_status;
 
 	flush_cached_sections(ImageHandle);
```
bluca, line 420: don't we need to flush the cached sections in either case here?

PR #770, opened 2025-08-07
fix http boot

jsetje:
This is a simple typo that is currently breaking http boot. When the interior loop goes looking for a second Content-Length field, will always match the first one, and if that's some other field things break.


vathpela:
I think @rosslagerwall's way is fine.

jsetje:
@rosslagerwall thanks for pushing to clean that up as well!

buf_size_set is kind of required since someone could spoof a 0 content length, and if they do, we probably want to catch that.

rosslagerwall (approved):

PR #771, opened 2025-08-11
gitignore: add more mkosi dirs and vscode dir

bluca:


vathpela:
Looks like mkosi is failing on fedora because it's getting packages from rawhide during the f43 branch day, and so the gpg pubkeys in the new package are for f44 but the rebuild hasn't landed yet.

PR #772, opened 2025-08-15
pe: Fix PF in GRUB after memattrs call

rosslagerwall:
The call to update_mem_attrs() takes an aligned pointer within an allocated region but passes the entire size of the allocated region. The result is that Shim may remove execute permission from some pages belonging to GRUB causing a page fault upon returning from the LoadImage call.


rosslagerwall:
The three test failures are build errors and don't seem to be related to the change here.

PR #775, opened 2025-09-24
tests: fix build on musl-based systems

akodanev:
This PR addresses two issues related to glibc-specific usage in the tests to ensure they can be built with musl libc too:

* the use of `__GNUC_PREREQ` macro from `features.h` - replaced with GNUC_PREREQ that is provided by `shim.h` and implements the same functionality.
* and the use of `execinfo.h` (backtrace functions) - musl libc does not provide the backtrace functions in its implementation. These functions are typically provided by glibc's `execinfo.h`. As a fallback for musl-based systems, support linking against the external libunwind library and using dladdr() to obtain similar stack trace info. Note that the usage of dladdr() is similar to that of backtrace_symbols in glibc, and it can be used with glibc as well.

### DIFF
--- a/include/test.mk
+++ b/include/test.mk
@@ -13,6 +13,8 @@ DEBUG_PRINTS ?= 0
 OPTIMIZATIONS=-O2 -ggdb
 CFLAGS_LTO =
 CFLAGS_GCOV =
+LIBS = -lefivar
+
 CFLAGS = $(OPTIMIZATIONS) -std=gnu11 \
 	 -isystem $(TOPDIR)/include/system \
 	 $(EFI_INCLUDES) \
@@ -45,6 +47,11 @@ CFLAGS = $(OPTIMIZATIONS) -std=gnu11 \
 	 -DGNU_EFI_USE_MS_ABI -DPAGE_SIZE=4096 \
 	 -DSHIM_UNIT_TEST \
 	 "-DDEFAULT_DEBUG_PRINT_STATE=$(DEBUG_PRINTS)"
+
+ifneq ($(origin ENABLE_LIBUNWIND), undefined)
+CFLAGS += -DENABLE_LIBUNWIND
+LIBS += -lunwind
+endif
 
 # On some systems (e.g. Arch Linux), limits.h is in the "include-fixed" instead
 # of the "include" directory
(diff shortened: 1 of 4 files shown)


akodanev:
Only mkosi/boot fedora* checks fail. All are as follows:

```
ld -o shimx64.so --hash-style=sysv -nostdlib -znocombreloc -T /work/src/elf_x86_64_efi.lds -shared -Bsymbolic -Lgnu-efi/x86_64/gnuefi -Lgnu-efi/x86_64/lib -LCryptlib -LCryptlib/OpenSSL gnu-efi/x86_64/gnuefi/crt0-efi-x86_64.o --build-id=sha1  --no-undefined shim.o globals.o memattrs.o mok.o netboot.o cert.o dp.o loader-proto.o tpm.o version.o errlog.o sbat.o sbat_data.o sbat_var.o pe.o pe-relocate.o httpboot.o csv.o load-options.o utils.o verify.o Cryptlib/libcryptlib.a Cryptlib/OpenSSL/libopenssl.a lib/lib.a gnu-efi/x86_64/lib/libefi.a gnu-efi/x86_64/gnuefi/libgnuefi.a -lefi -lgnuefi --start-group Cryptlib/libcryptlib.a Cryptlib/OpenSSL/libopenssl.a --end-group /usr/lib/gcc/x86_64-redhat-linux/15/libgcc.a lib/lib.a
gcc -std=gnu11 -Og -g3 -Wall -Wextra -Wno-missing-field-initializers -Werror -o post-process-pe /work/src/post-process-pe.c
objcopy -D -j .text -j .sdata -j .data -j .data.ident \
	-j .dynamic -j .rodata -j .rel* \
	-j .rela* -j .dyn -j .reloc -j .eh_frame \
	-j .vendor_cert -j .sbat -j .sbatlevel \
	--file-alignment 0x1000 \
	--target efi-app-x86_64 shimx64.so shimx64.efi
objcopy: shimx64.so: file format not recognized
make: *** [/work/src/Makefile:280: shimx64.efi] Error 1
```
This is unrelated to the test build in the PR, but rather to recent changes in f44 that may be possible in binutils (2.45.50-3.fc44).

PR #777, opened 2025-09-26
Switch to upstream gnu-efi (v4.0.4)

andreabolognani:
This is part of an attempt to bring riscv64 support into shim.

See #420 for the initial attempt and #641 for a revived one; the latter PR contains a lot of discussion, including requests from the maintainers to consolidate the work floating around in various branches into a single PR that could be looked at.

Mandatory disclaimer: I know basically nothing about the shim code base, so what I'm doing here consists in 1) taking code other people have written followed by 2) doing whatever is needed to make the compiler stop shouting at me. I am under no delusion that any of this would clear the bar for merging as is, hence the draft status. It's intended to be primarily a conversation starter. With the maintainers' guidance, hopefully we can refine it into something suitable for upstream.

I will have a second PR building on this one soon, which will take care of actually adding riscv64 support.

PR #641, opened 2024-02-23
Updated support for SHIM on RISC-V

brianredbeard:
Add what is needed to build on riscv64.

Signed-off-by: Heinrich Schuchardt <heinrich.schuchardt@canonical.com>

This is an update which supersedes #420 which brings it in alignment with the current upstream and fixes a minor style nit around the definition of `__riscv`.

PR #420, opened 2021-09-30
Initial RISC-V support

xypron:
Add what is needed to build on riscv64.

Signed-off-by: Heinrich Schuchardt <heinrich.schuchardt@canonical.com>

PR #778, opened 2025-09-26
Add riscv64 support

andreabolognani:
This is part of an attempt to bring riscv64 support into shim.

See #420 and #641 for previous discussion, as well as #777 which is a prerequisite of this PR.

Using this branch I was able to successfully build shim on both x86_64 and riscv64, and I was able to use the resulting binaries to boot Fedora 42 cloud images for both architectures, replacing the stock ones.


andreabolognani:
Tagging @xypron, @brianredbeard and @jmontleon since some of the changes come from them.

andreabolognani:
Rebased on top of https://github.com/rhboot/shim/pull/777.

Additionally, I have dropped all definitions related to `va_list` from shim's `stdarg.h` since they no longer appear to be necessary when using the latest upstream version of gnu-efi. I believe their existence was a point of contention in the past.

@xypron perhaps you could revisit your patch, integrating the follow-up fixes and improvements from me, @brianredbeard and @jmontleon? And of course any feedback from @vathpela would be much appreciated too :)

Thanks in advance!

andreabolognani:
Rebased to match the current contents of the `main` branch, matching #777. Fixed a couple of signoffs with the author's explicit agreement.

andreabolognani:
Rebased once again on top of the updated #777. Squashed some patches (with agreement from the authors) and fixed VPATH builds.

andreabolognani:
Rebased on top of #777 and added CI jobs for riscv64 cross-builds.

andreabolognani:
Rebased on top of #777 once again.

andreabolognani:
@xypron can you please consider giving this some sprucing up? The CI now passes and #777 looks close to being ready. You can feel free to squash any changes of my own into your commit, and I'm pretty confident neither @brianredbeard nor @jmontleon would mind if you did the same with theirs. Thanks!

brianredbeard:
I've got one thing for you to cherry-pick: https://github.com/brianredbeard/redhat-efi-boot-shim/commit/29593a024755b1de3845fe0253b98d5f2cd1d8f8

The TL;DR is that `.vendor_cert`, `.data.ident`, and `.sbatlevel`. These sections hold (respectively) the embedded vendor certificates, the SBAT relocation level data, and the build identity information.  Makefile:281-284 uses objcopy to extract these values from the elf to produce the PE image and without building that correctly we would end up dropping the data.

Not having the cert verification or revocation would probably defeat the point of what we're working on 😆 

Like my previous attempt, I'm happy to play janitor and do the clean-up step as we go.

brianredbeard:
I've worked through #777 and this one (#778) I've done a review, found and fixed several issues, and rebased the combined work into a cleaner commit series.

Issues fixed in the riscv64 linker script (elf_riscv64_efi.lds):
- Added missing .vendor_cert, .sbatlevel, and .data.ident output sections (these were never carried over from the gnu-efi 4.0.4 template, breaking vendor cert verification, SBAT variable payload, and version identification)
- Merged .rodata into .data to match all other shim architectures
- Fixed ALIGN(512) → page-aligned _edata placement
- Removed manual _DYNAMIC definition and unnecessary KEEP on .reloc
- These were all artifacts of copying the gnu-efi 4.0.4 script without applying shim's standard adaptations — the same adaptations already made for x86_64, ia32, and aarch64

Commit structure (4 commits, down from 19):
1. gnu-efi: Update to upstream 4.0.4 — submodule switch + all API migrations (ReallocatePool, CompareGuid, VPrint, stdarg, string signedness)
2. Linker scripts: Fix section ordering and cleanup — .reloc reorder, .eh_frame/.build-id dedup, ARM .areloc
3. Add RISC-V 64-bit (riscv64) architecture support — corrected linker script, PE/COFF support, build system, icache flush
4. CI: Add riscv64 cross-build jobs

All original authors are preserved via --author, Co-Authored-By, and Signed-off-by. The final tree is identical to the current HEAD of #778 except for the linker script fixes and a minor whitespace fix in peimage.h.  The same whitespace nit effects the ARM entry (a tab instead of spaces as per all other entries).  I can take care of fixing that too if folks please.

Branch is at https://github.com/brianredbeard/redhat-efi-boot-shim/tree/riscv-squash — happy to open a PR superseding both #777 and #778 if that works for everyone.  

One note, the current tag for the submodule is master and not 4.0.4 (or the proper tag/branch).  I just want to confirm that's correct.

andreabolognani:
@brianredbeard thanks a lot for taking a look!

I've checked out your squash branch and I have to say that I'm personally not a big fan of lumping so many changes together into a small number of large commits: I much prefer having a larger number of smaller commits.

The exception to the above is when you have a commit that introduces a change, only for the next commit to (partially) undo it. That's very much the case here, mostly on account of @xypron providing the original implementation and several other people adding fixes and improvements on top. In this specific scenario, I would welcome squashing some of the commits together to avoid unnecessary churn. But then, the original implementation could itself potentially use some splitting into smaller incremental steps...

Anyway, I've cherry-picked your additional changes for now. Ultimately it's up to the maintainers whether they would rather have many small commits or few large ones, and I'd be happy to go in either direction once they will have made their preference known.

andreabolognani:
#777 is already organized pretty much optimally as far as I'm concerned, and there are no functional changes that you're proposing for that part, so I just left it alone. The whitespace changes you mentioned and any other cleanups should IMO be made as a follow-up, to avoid potentially holding up the functional parts due to disagreements on non-functional concerns.

andreabolognani:
Rebased on top of `main` now that #777 has been merged.

@xypron can you please take another look at this? The coast is finally clear for getting it merged, it just need a bit of tidying up.

In particular we want to avoid the situation where some code is introduced in one commit only to be dropped/altered in a follow-up one. Makes the git history confusing for no particular good reason. I think commits 2-6 should definitely be squashed into commit 1; jury's out when it comes to commits 7 and 8. It would be perfectly fine IMO to just squash the whole thing together, unless the maintainers disagree with that approach.

The DCO job complains about git authorship information disagreeing with the signoff. That will need to be fixed too.

I can take care of all of the above myself if that's more convenient for you. Just let me know which of the two email addresses should be associated with the change, and I'll do the rest.

Thanks!

xypron:
@andreabolognani

Thank you for driving this. Please, use my Canonical email address.

I have added reviewing the series to next weeks Jira cards.

andreabolognani:
@xypron fantastic!

I've squashed everything together, made a couple tiiiiiny whitespace tweaks, and adjusted authorship information. Once you get a chance to look at the results, if there are any further changes that you think are necessary, we can figure out how to go about them.

... and of course in the time between the previous update to the PR and now, Debian testing has become broken somehow and all mkosi jobs are failing :) We'll figure that out too.

xypron:
Hello @andreabolognani,

Without POST_PROCESS_PE_FLAGS='-n' I see these build warning:

./post-process-pe -vv  shimriscv64.efi
validate_nx_compat():379: NX Compatibility flag is not set

./post-process-pe -vv  mmriscv64.efi
validate_nx_compat():379: NX Compatibility flag is not set

./post-process-pe -vv  fbriscv64.efi
validate_nx_compat():379: NX Compatibility flag is not set

The only executable section is (as reported by [efi_analyzer](https://github.com/xypron/efi_analyzer)):

Section[1]: .text
  Virtual size: 0x4f930
  Virtual address: 0x1e000
  Size of raw data: 0x50000
  Pointer to raw data: 0x1a000
  End of raw data: 0x6a000
  Characteristics: 0x60000020
    * The section contains executable code.
    * The section can be executed as code.
    * The section can be read.

Hence the NX Compatibility flag should be set. gnu-efi/riscv64/gnuefi/crt0-efi-riscv64.S would be the place to do so.
I think we should add a patch to upstream gnu-efi to set the flag in files gnu-efi/riscv64/gnuefi/crt0-efi-*.S to avoid having to set it via POST_PROCESS_PE_FLAGS.

Best regards

Heinrich

andreabolognani:
@xypron thanks for checking things out!

A lot of this goes over my head, so please help me understand. Is the "NX Compatibility flag" something that is necessary for basic enablement, or could we realistically leave things as they are for now and fix things with a follow-up? What's the actual impact from it being missing?

I think getting gnu-efi patched and bumping the dependency on the shim side would be relatively straightforward now that we're tracking upstream more closely, but I'm still somewhat wary of introducing an additional step now that we're finally so tantalizingly close to getting riscv64 support merged.

So if that could be done as a follow-up enhancement, I would probably prefer going that route. But I'll trust your judgement on whether that's actually feasible.

Did you spot any other issues, or are you otherwise happy with the PR?

Thanks again!

xypron:
@andreabolognani 
The requirement by Microsoft is that EFI binaries should not have sections that are both executable and writable. Such binaries should be rejected in a secure boot scenario. The NX flag in the PE-COFF header is used as a signal that the condition is met. We should set it for shim, GRUB, and the kernel.

andreabolognani:
@xypron okay so IIUC it's a requirement for Secure Boot specifically, not for basic operations of shim (loading grub, running the fallback boot path).

If that's the case, considering that there is still no agreement on who should even sign the shim binary in the context of Secure Boot on riscv64, I'd be inclined to not consider that a blocker and proceed with the PR as is.

Perhaps in the meantime you could open a PR against gnu-efi that introduces the necessary changes? That would get the ball rolling for NX as a follow-up improvement.

xypron:
> Perhaps in the meantime you could open a PR against gnu-efi that introduces the necessary changes? That would get the ball rolling for NX as a follow-up improvement.

Sounds good

andreabolognani:
Attempting to fix / work around the issue currently preventing all mkosi jobs from running. If it works, I will open a separate PR for it.

xypron:
Thank you @andreabolognani for pushing this forward.

With signed shim, grub, kernel and a security database (PK, KEK, db) in U-Boot secure boot is working:
```
[    0.000000] Linux version 7.0.0-14-generic (buildd@bos03-riscv64-054) (riscv64-linux-gnu-gcc (Ubuntu 15.2.0-16ubuntu1) 15.2.0, GNU ld (GNU Binutils for Ubuntu) 2.46) #14.2-Ubuntu SMP PREEMPT_DYNAMIC Fri Apr 17 12:35:32 UTC 2026 (Ubuntu 7.0.0-14.14.2-generic 7.0.0)
...
[    0.000000] secureboot: Secure boot enabled
[    0.000000] Kernel is locked down from EFI Secure Boot mode; see man kernel_lockdown.7
...
Welcome to Ubuntu 26.04 LTS!
```
With a byte changed in grubriscv64.efi:
```
                    ERROR
Verification failed: (0x1A) Security Violation
```
LGTM

P.S. Most of the test preparation steps are available in https://github.com/xypron/test-shimriscv64.git

andreabolognani:
@xypron great to hear that!

The tentative changes I've made to the mkosi configuration seem to have restored CI to its previous state: two of the CentOS jobs are still failing, but we [already knew](https://github.com/rhboot/shim/pull/777#issuecomment-4790364304) about that and a fix is supposed to be in the works. I will prepare a separate PR for those changes.

We should still look into getting NX wired up properly. Please tag me in the gnu-efi PR when you get around to creating it. I'll keep an eye on it and look into following up on the shim side once it has landed.

xypron:
@andreabolognani
NX can be set via environment variable POST_PROCESS_PE_FLAGS='-n' during the shim build process.
There is also POST_PROCESS_PE_FLAGS='-N' to remove the flag.

Changing the default in gnu-efi makes sense but is orthogonal to this pull request.

andreabolognani:
@xypron based on the documentation, `$POST_PROCESS_PE_FLAGS` appears to be intended as an override mechanism for end users and distro packagers. No existing architecture is using it. So it seems to me that modifying gnu-efi is the proper way to go about it.

@vathpela can you please take a look at #806 (which contains the same CI changes mentioned earlier) and another look at this one? Given Heinrich's positive feedback, it should be possible to merge it now. And considering that all the changes made to common code are merely adding riscv64-specific branches, without touching the existing architectures at all, it should be entirely safe to do so.

Thanks!

xypron:
@andreabolognani

Looking at the gnu-efi code a bit deeper. HAVE_EFI_OBJCOPY is set for riscv64. This means the value of DllCharacteristics is not defined via gnuefi/crt0-efi-riscv64-local.S but provided by the toolchain. Only armv7 and mips use local crt0 files, assuming binutils >= 2.38. So setting the NX flag in shim make sense.

andreabolognani:
@xypron I'm all for that, but unclear on how to make it happen cleanly. `$POST_PROCESS_PE_FLAGS` doesn't seem like the way to go for all the reasons I mentioned earlier. How do other architectures handle this? I've poked around a bit and I haven't been able to figure it out so far.

xypron:
@andreabolognani 

Ubuntu 26.10 carries a patch
https://git.launchpad.net/ubuntu/+source/shim/tree/debian/patches/Build-an-additional-NX-shim-mark-MokManager-and-Fallback-.patch which builds both a version of shim with NX flag and one without. The build parameter `POST_PROCESS_PE_FLAGS=-n` is used for the NX version.

OpenSuSE Leap shim-16.1-2.2.aarch64.rpm and Fedora shim-x64-16.1-7.x86_64.rpm don't set the NX flag.

andreabolognani:
@xypron thanks for providing that pointer!

It looks like Debian also uses `POST_PROCESS_PE_FLAGS=-n` to enable NX, see [debian/rules#L48-49](https://salsa.debian.org/efi-team/shim/-/blob/b41cd6471060b12317159090b4a69a7115fc61e1/debian/rules#L48-49).

In both all cases seen so far the change happens at the distro package level, and it's applied regardless of architecture. So it appears to me that we don't really need to do anything in shim itself? Please let me know if I've missed anything.

andreabolognani:
Looks like both Debian and CentOS have been fixed. We're back to a fully green pipeline!

@vathpela can you please consider merging this? Before some other unrelated thing breaks it again ;)

vathpela:
Mostly this looks good but I have a few questions:
- what's up with `edk2asm.h` and `riscv64_cache.S`?  Do we really need a header file for two macros that each get used once in a source file of basically the same length?
- but also what's up going on with `__riscv_flush_icache()` in general?  I can't see where it's being used.  We never call it directly, and it doesn't look like it's fulfilling some implementation detail of `__builtin___clear_cache()`
- but if it were the latter, why are we doing that here instead of defining `CLEAR_INSN_CACHE()` in gcc's `config/riscv...` so it'll be right in libgcc?

At the least I think this needs some comments documenting what it's doing and how, because right now I can't figure either out.

xypron:
__riscv_flush_icache() seems to be a neglected left-over from my original merge request.

The instruction cache must be flushed before executing a binary (grub) loaded by shim. We link with libgcc.a, so we can use the builtin function in handle_image(). File lib/riscv64_cache.S can be removed together with lib/edk2asm.h.

andreabolognani:
@xypron @vathpela dropping the files results in the following error at build time:

```
riscv64-linux-gnu-ld -o shimriscv64.so --hash-style=sysv -nostdlib -znocombreloc -T /root/shim/elf_riscv64_efi.lds -shared -Bsymbolic -Lgnu-efi/gnuefi -Lgnu-efi/lib -LCryptlib -LCryptlib/OpenSSL gnu-efi/gnuefi/crt0-efi-riscv64.o --build-id=sha1  --no-undefined shim.o cert.o csv.o dp.o errlog.o hexdump.o httpboot.o globals.o load-options.o loader-proto.o memattrs.o mok.o netboot.o pe.o pe-relocate.o sbat.o sbat_data.o sbat_var.o time.o tpm.o utils.o verify.o version.o Cryptlib/libcryptlib.a Cryptlib/OpenSSL/libopenssl.a lib/lib.a gnu-efi/lib/libefi.a gnu-efi/gnuefi/libgnuefi.a -lefi -lgnuefi --start-group Cryptlib/libcryptlib.a Cryptlib/OpenSSL/libopenssl.a --end-group /usr/lib/gcc/riscv64-linux-gnu/15/lib64/lp64d/libgcc.a lib/lib.a
riscv64-linux-gnu-ld: pe.o: in function `handle_image':
/root/shim/pe.c:612:(.text+0x1498): undefined reference to `__riscv_flush_icache'
```

I've poked around the gcc source repo and that's what I can gather: initially gcc would emit the `fence.i` instruction itself, but later [this patch](https://gcc.gnu.org/legacy-ml/gcc-patches/2018-01/msg00072.html) (commit [`f1bdc63a898f`](https://github.com/gcc-mirror/gcc/commit/f1bdc63a898f2fa164e79528c03e96a21a779e82)) changed that hardcoded behavior into a more dynamic one where gcc defines the entry point (`__riscv_flush_icache`) and the program / libc / vDSO is responsible for providing the actual implementation.

While I can't claim that I fully grasp the details of the issue with multicore processors that is referenced in the gcc patch, my rough understanding is that Linux needs to be involved directly because its scheduler can move processes between harts and that could cause the incorrect cache to be invalidated under the "right" conditions.

In the case of shim, everything is running before the OS gets involved and (I assume?) on a single hart. So calling the fence instruction directly ourselves doesn't seem unreasonable. Indeed, that appears to be what edk2 is doing too ([`RiscVInvalidateInstCacheFenceAsm`](https://github.com/tianocore/edk2/blob/master/MdePkg/Library/BaseLib/RiscV64/RiscVCacheMgmt.S#L17-L19)). On the other hand, they don't seem to be linking against libgcc so they don't have to get involved into shenanigans with used-but-undefined symbols.

I'm not sure what the best way forward would be. One option could be implementing a mechanism in gcc which allows the application linking against libgcc to declare "I don't want the dynamic handling of icache invalidation to be involved, just run `fence.i`", perhaps via defining a preprocessor macro or something like that. We'd then need to wait for the change to be merged and percolate all the way down to the various distros. The alternative would be to keep things as they are in this PR, and rely on the fact that the interface between gcc and Linux / vDSO / glibc has been set in stone at this point and will not be changed in the future.

Personally I'm leaning towards the latter, but I agree that the header file in particular doesn't seem to offer much value, and that the reason why the function needs to be implemented is not explained at all. I will take a stab at improving on those two aspects, then we can look at the result and discuss further. I'm obviously open to hearing additional options that I haven't considered too :)

andreabolognani:
Made the changes mentioned above. Haven't tested the resulting binary yet (will do that within the next few hours) but the disassembly is identical, so I'm quite confident things will work just fine.

andreabolognani:
I can confirm that a shim binary built from the updated branch works as expected.

xypron:
> While I can't claim that I fully grasp the details of the issue with multicore processors that is referenced in the gcc patch, my rough understanding is that Linux needs to be involved directly because its scheduler can move processes between harts and that could cause the incorrect cache to be invalidated under the "right" conditions.

EFI runs on a single hart. Here it is allowable to use fence.i. The other harts are waiting in OpenSBI for Linux to wake them up.

andreabolognani:
Thanks @vathpela! And of course thanks to @xypron, @brianredbeard, @jmontleon and @gmbr3 too! This would never, ever have been possible without all of your contributions :)

PR #779, opened 2025-10-17
mok: Add a building option to avoid the vendor_dbx being appended to …

joeyli:
…MokListRT

The new building option DISABLE_APPEND_V_DBX_TO_MOKX avoids appending the vendor_dbx to MokListRT. Instead, the vendor_dbx will be exposed through the mok config table space.

As the vendor-dbx grows, it caused some problems when writing such a large variable. Some firmwares lie the avaiable space [1], and some even crash [2] for no good reason after the writing of MokListXRT. If your shim and kernel don't rely on MokListXRT to block anything, you can use this building option to expose vendor_dbx through the mok config table instead of appending it to MokListRT.

Link: https://bugzilla.suse.com/show_bug.cgi?id=1185261 [1]
Link: https://github.com/rhboot/shim/pull/369#issuecomment-855275115 [2]

### DIFF
--- a/BUILDING
+++ b/BUILDING
@@ -75,6 +75,17 @@ Variables you could set to customize the build:
 - ENABLE_CODESIGN_EKU
   This changes the certificate validation logic to require Extended Key
   Usage 1.3.6.1.5.5.7.3.3 ("Code Signing").
+- DISABLE_APPEND_V_DBX_TO_MOKX
+  This building option avoids the vendor_dbx being appended to MokListRT.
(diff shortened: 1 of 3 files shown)


vathpela:
I don't think I have a problem with any of this, but it does lead me to some questions about the direction we should be going:

1. I wonder if it's not worth always having VendorDBX
2. If we're doing that, why do we merge at all, why not separate everything?  In the past we've merged them so that userland tools could just look at one thing, but maybe that's not the best way forward.
3. Clearly we should be moving away from creating the EFI variables at all...

PR #781, opened 2025-10-28
Aarch64: make section alignment larger

vathpela:
In UEFI Mantis 2549, ARM proposes to recommend 64kB section alignment in
UEFI binaries.  The reason for this is that some machines may be
configured to have no pages smaller than 64kB for performance reasons,
and in such a scenario things like W^X will be impossible to correctly
implement on binaries where one section ends and another begins on the
same 64kB page.

This change makes aarch64 binaries use a 64kB rather than 4kB section
alignment.  Note that this only actually changes virtual addresses, not
file addresses, so the size is unchanged.

FWIW I ran this by ardb and he said it looks sane to him.


josuedhg (approved):
LGTM

PR #784, opened 2025-11-24
sbat: bump for November Grub CVEs

Fabian-Gruenbichler:
while this are not major vulnerabilities, we still need an SBAT bump to allow referencing and revoking them on their own.

since the main grub level is bumped, we can drop the .proxmox one.


Fabian-Gruenbichler:
since https://github.com/rhboot/shim-review/pull/507 was merged - could we maybe take another look at this as well? ;)

steve-mcintyre (approved):
Looks fine to me

PR #790, opened 2026-03-12
fallback: fix memory leak in find_boot_csv()

yueyuankun:
In find_boot_csv(), bootcsv and bootarchcsv are allocated by StrDuplicate() but were not being freed before the function returns, causing a memory leak.


vathpela (commented):
Looks right to me.  Please fix the DCO.

yueyuankun:
@vathpela  DCO fixed.

PR #792, opened 2026-03-19
CI: Update from f39/f40/f41 to f42/f43/f44

vathpela:
Fedora 41 is pretty old, so update to some supported (or nearly supported) OSes.


vathpela:
This makes CI work again for other posts, so I'm gonna push it.

PR #804, opened 2026-06-24
Update CentOS Stream kernel signing cert.

vathpela:
CentOS recently rotated kernel and grub's signing certs and keys, so we need to reflect the new one now.


vathpela:
Right now this fails because 1) as currently implemented the mkosi code doesn't support `VENDOR_DB_FILE=` instead of `VENDOR_CERT_FILE=` so we can only have one cert, and 2) the CentOS Stream 10 kernel is incorrectly still being signed with the old certificate.

I'm gonna push this anyway and manually wave CI on things that need those two targets, while others work on fixing the kernel.

PR #806, opened 2026-06-30
mkosi: Use Debian stable

andreabolognani:
This restores CI functionality for the most part, modulo a couple CentOS failures that we were already [aware of](https://github.com/rhboot/shim/pull/777#issuecomment-4790364304) and for which a fix is being prepared on the CentOS side.

I will file an issue against mkosi to try and figure out how to get the Debian testing jobs working again, but in the meantime this allows us to validate incoming changes.

### DIFF
--- a/mkosi/mkosi.conf
+++ b/mkosi/mkosi.conf
@@ -8,6 +8,8 @@ OutputDirectory=mkosi.output
 [Build]
 History=yes
 ToolsTree=default
+ToolsTreeDistribution=debian
+ToolsTreeRelease=stable
 BuildDirectory=mkosi.builddir
 CacheDirectory=mkosi.cache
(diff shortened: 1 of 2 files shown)


andreabolognani:
Based on additional testing, this seems to have been addressed in Debian unstable already. Which means that we should be able to revert this commit and switch back to Debian testing in a week or so.

andreabolognani:
> we should be able to revert this commit and switch back to Debian testing in a week or so

Actually it could take longer because the autopkgtest results are not looking good. https://tracker.debian.org/pkg/systemd

andreabolognani:
The new version of systemd managed to migrate to testing earlier than I thought it would. This is no longer needed.

PR #807, opened 2026-07-02
Periodically update the UEFI watchdog timer when we're printing to the console...

vathpela:
Two different people recently mentioned to me that they'd set SHIM_VERBOSE and soft-bricked their devices because the console was too slow and it eventually tripped the watchdog.

This patch series adds code to console_print() and related routines such that when we are in verbose mode and thus printing to the console, on each print call, if we have not updated the watchdog in the last minute, we will set the watchdog to go off in 5 minutes.

This way we're not updating it very often but we are giving ourselves *plenty* of time to do whatever we're doing, and if the watchdog does legitimately need to fire it still will, eventually.

### DIFF
--- a/Cryptlib/Include/CrtLibSupport.h
+++ b/Cryptlib/Include/CrtLibSupport.h
@@ -115,7 +115,6 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 // Basic types mapping
 //
 typedef UINTN   off_t;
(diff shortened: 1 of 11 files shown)


nfrayer (approved):
lgtm.
Also Tested-by: Nicolas Frayer <nfrayer@redhat.com>

PR #818, opened 2026-07-23
More minor fuzzing cleanups

vathpela:


include/fuzz.mk
corvus-callidus, line 105: Since you're already doing extra cleanup:

issue (non-blocking): write one profile per fuzz worker

The fuzzer starts 24 workers with `-jobs=24`, but they all write to the
same `LLVM_PROFILE_FILE`. Their exit-time profile writes can race or overwrite
one another, so the resulting coverage data is incomplete or unreliable.

LLVM supports `%p` for per-process filenames:
https://clang.llvm.org/docs/SourceBasedCodeCoverage.html#id4

Could we use `LLVM_PROFILE_FILE="$@.%p.profraw"`, merge the shards with
`llvm-profdata merge`, then remove them after a successful merge? My understanding is that cov-analysis doesn't use these files, but llvm-cov does, and it would be nice to have correct results.

Makefile
```diff
@@ -469,23 +469,31 @@ else
...
-test test-clean test-coverage test-lto : generated_sbat_var_defs.h
+test test-coverage test-lto : generated_sbat_var_defs.h
 	@make -f $(TOPDIR)/include/test.mk \
 		COMPILER="$(COMPILER)" \
 		CROSS_COMPILE="$(CROSS_COMPILE)" \
 		CLANG_WARNINGS="$(CLANG_WARNINGS)" \
 		ARCH_DEFINES="$(ARCH_DEFINES)" \
 		EFI_INCLUDES="$(EFI_INCLUDES)" \
-		test-clean $@
+		$@
```
corvus-callidus, line 490: question: is cumulative coverage data intentional?

Previously top-level `test-coverage` invoked `test-clean` before running the
coverage target, giving each coverage run a fresh `.gcda` baseline. This now
retains existing counters, so later runs accumulate prior execution data.

That can be useful for aggregate coverage across several runs, but it changes
the meaning of a single `make test-coverage` invocation. Is that intentional?

include/fuzz.mk
vathpela: That sounds good.  Can you submit a PR for that?

Makefile
vathpela: > question: is cumulative coverage data intentional?

For the fuzzers I was thinking yes but I don't really have a strong goal.  For the "test-" targets, not really, but we're not storing them anywhere, so in CI they won't be.  But in the current code we've got:
```
test test-coverage test-lto : | clean-test-results
test test-coverage test-lto : generated_sbat_var_defs.h
```
so if you're just running `make test` it'll clean it.

include/fuzz.mk
corvus-callidus: Will do.

Makefile
corvus-callidus: Sounds reasonable to me. I just wanted to make sure we weren't ending up with unintended consequences.

corvus-callidus (approved):

PR #827, opened 2026-08-24
Added ML-DSA verification

nfrayer:
- Enabled SHA3 in the configuration header for SHAKE-128/SHAKE-256
- Modified Pkcs7Verify() to add a code path for ML-DSA-87


Cryptlib/Pk/CryptPkcs7VerifyCommon.c
```diff
@@ -898,6 +979,35 @@ Pkcs7Verify (
...
+    PKCS7_SIGNER_INFO  *Si = NULL;
+    EVP_PKEY  *Pkey = NULL;
+
+    if (Sinfos != NULL && sk_PKCS7_SIGNER_INFO_num (Sinfos) > 0 &&
+        Signers != NULL && sk_X509_num (Signers) > 0) {
+      Si = sk_PKCS7_SIGNER_INFO_value (Sinfos, 0);
+      Pkey = X509_get0_pubkey (sk_X509_value (Signers, 0));
+    }
+
+    if (Pkey != NULL && EVP_PKEY_is_a (Pkey, "ML-DSA-87")) {
```
vathpela: Is there a reason we're specifically checking for 87 here?  It seems like all ML-DSA should go through this path, and any size should work so long as they're the same in `VerifyMlDsaPkcs7Signature()`

vathpela (changes requested):

PR #829, opened 2026-09-03
Harden SBAT handling against truncation and impossible to revoke levels

julian-klode:
Commit 1 removes the truncation to UINTN to UINT16, and uses a custom parser to avoid the discrepancy between atoi() in the test environment and atoi() in EFI.

Commit 2 hard rejects images with SBAT level UINT16_MAX or greater such that UINT16_MAX can be used to revoke the last binary for that component.


julian-klode:
Not sure the "invalid" sbat variable handling makes sense, should we detect that earlier and re-write the sbat variable if it is corrupted?

julian-klode:
Cleaning up the missing partial line aand some weird typing

julian-klode:
All lines are fully covered now. If we want we could also allow 32-bit integers, but this keeps the 16-bit limit.

